PR #250, opened 2019-04-09
bugfix for fetching query results

### LINKED ISSUE #249: Bugfix for fetching query results through multiple rpc call in gohive.
for the field `HasMoreRows` in TFetchResultsResponse always return false in hiveserver2, in order to fetch query results data through multiple requests, instead, we check if the value in resultSet is empty to judge if there is more data to be fetched.

yuyicg:
fix #249


tonyyang-svail:
@tonyyang-svail I will follow up this PR.

PR #265, opened 2019-04-11
Use if [[ instead of if [

wangkuiyi:
Travis CI complains about the use of `if [`. So I am trying to use `if [[`.


tonyyang-svail (approved):
LGTM.

PR #266, opened 2019-04-11
Travis deploy

wangkuiyi:
It seems that Travis CI doesn't use Bash to interpret the script in `.travis.yml`, and cannot handle the `if [` or `if [[` statement. I am merging the two deploy scripts into one so to avoid the `if` in `-script`.


tonyyang-svail (approved):
LGTM.

PR #283, opened 2019-04-17
Add column resolver in ALPS submitter

### LINKED ISSUE #106: Thinking about column type inference from SQL field types
The current ad-hoc solution is to use a map

https://github.com/wangkuiyi/sqlflow/blob/80e1fe77eb5c6e8ed1c0ffd3869671e8574ae7de/sql/codegen.go#L15

It might be too weak to support any case other than (1) each field is a feature, and (2) all fields are of float type.

For example, users might write the following statement which operates on a single string-typed field `f`:

```sql
SELECT f
FROM table
COLUMN f, hash(f, 100), cross(f, hash(f, 100))
```

In our Estimator code, the feature list needs three elements:

```python
[
   tf.feature_column.categorical_column_with_vocabulary_list("f", vocab_list), 
   tf.feature_column.categorical_column_with_hash_bucket(
       tf.feature_column.categorical_column_with_vocabulary_list("f", vocab_list), 100),
   tf.feature_column.cross_column(
       tf.feature_column.categorical_column_with_vocabulary_list("f", vocab_list), 
       tf.feature_column.categorical_column_with_hash_bucket(
           tf.feature_column.categorical_column_with_vocabulary_list("f", vocab_list), 100))
]
```

uuleon:
Fix https://github.com/sql-machine-learning/sqlflow/issues/106

### DIFF
--- a/sql/alps_submitter.go
+++ b/sql/alps_submitter.go
@@ -1,0 +1,163 @@
+package sql
+
+import (
+	"fmt"
+	"io"
+	"strconv"
+)
+
+type GraphConfig struct {
+	TableName string
+	Fields    []string
+	X         []FeatureSpec
+	Y         FeatureSpec
+}
+
+type FeatureSpec struct {
+	FeatureName string
+	IsSparse    bool
+	Shape       []int
+	DType       string
+}
+
+type NumericColumn struct {
+	Key   string
+	Shape int
+}
+
+type BucketColumn struct {
+	SourceColumn NumericColumn
+	Boundaries   []int
+}
+
+type CrossColumn struct {
+	Keys           []interface{}
+	HashBucketSize int
+}
+
+func resolveFeatureSpec(el *exprlist) (*FeatureSpec, error) {
+	headExpr := (*el)[0].val
+	if headExpr == "DENSE" {
+		shape, err := strconv.Atoi((*el)[2].val)
+		if err != nil {
+			return nil, err
+		}
+		return &FeatureSpec{
+			FeatureName: (*el)[1].val,
+			IsSparse:    false,
+			Shape:       []int{shape},
+			DType:       "double"}, nil
+	} else if headExpr == "SPARSE" {
+		shape, err := strconv.Atoi((*el)[2].val)
+		if err != nil {
+			return nil, err
+		}
+		return &FeatureSpec{
+			FeatureName: (*el)[1].val,
+			IsSparse:    false,
+			Shape:       []int{shape},
+			DType:       "double"}, nil
+	}
+	return nil, fmt.Errorf("not supported encoding type: %s", headExpr)
+}
+
+func resolveExpr(e *expr) (string, *exprlist) {
+	if e.val != "" {
+		return e.val, nil
+	} else {
+		return "", &e.sexp
+	}
+}
+
+func resolveExprList(el *exprlist) (interface{}, error) {
+	headName, headExprList := resolveExpr((*el)[0])
+	if headName == "" {
+		return resolveExprList(headExprList)
+	} else {
+		switch headName {
+		case "DENSE":
+			return resolveFeatureSpec(el)
+		case "SPARSE":
+			return resolveFeatureSpec(el)
+		case "NUMERIC":
+			shape, err := strconv.Atoi((*el)[2].val)
+			if err != nil {
+				return nil, err
+			}
+			return &NumericColumn{
+				Key:   (*el)[1].val,
+				Shape: shape}, nil
+		case "BUCKET":
+			sourceExprList := (*el)[1].sexp
+			boundariesExprList := (*el)[2].sexp
+			source, err := resolveExprList(&sourceExprList)
+			if err != nil {
+				return nil, err
+			}
+			boundaries, err := resolveExprList(&boundariesExprList)
+			if err != nil {
+				return nil, err
+			}
+			return &BucketColumn{
+				SourceColumn: *source.(*NumericColumn),
+				Boundaries:   transformToIntList(boundaries.([]interface{}))}, nil
+		case "CROSS":
+			keysExpr := (*el)[1].sexp
+			bucketSize, _ := strconv.Atoi((*el)[2].val)
+			keys, err := resolveExprList(&keysExpr)
+			if err != nil {
+				return nil, err
+			}
+			return &CrossColumn{
+				Keys:           keys.([]interface{}),
+				HashBucketSize: bucketSize}, nil
+		case "square":
+			var list []interface{}
+			for idx, expr := range *el {
+				if idx > 0 {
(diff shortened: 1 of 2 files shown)


sql/alps_submitter.go
```diff
@@ -0,0 +1,167 @@
+package sql
+
+import (
+	"fmt"
+	"io"
+	"strconv"
+)
+
+type GraphConfig struct {
```
weiguoz: Should we expose `GraphConfig` struct? If no, turn it to `graphConfig` with lowercase.

sql/alps_submitter.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+	return nil, nil
+}
+
+func transformToIntList(interfaceList []interface{}) []int {
+	var b []int
+	for i := range interfaceList {
+		b = append(b, interfaceList[i].(int))
+	}
+	return b
```
weiguoz: Since we know interfaceList's length. You might want to define a bounded dest array for more efficiency.

sql/alps_submitter.go
```diff
@@ -0,0 +1,167 @@
...
+
+func resolveExprList(el *exprlist) (interface{}, error) {
+	headName, headExprList := resolveExpr((*el)[0])
+	if headName == "" {
+		return resolveExprList(headExprList)
+	} else {
+		switch headName {
+		case "DENSE":
+			return resolveFeatureSpec(el)
+		case "SPARSE":
```
weiguoz: "SPARSE" appears more than once. Perhaps we should define a const. (Same to "DENSE")

wangkuiyi (dismissed):

sql/alps_submitter.go
```diff
@@ -0,0 +1,167 @@
...
+	if headExpr == "DENSE" {
+		shape, err := strconv.Atoi((*el)[2].val)
+		if err != nil {
+			return nil, err
+		}
+		return &FeatureSpec{
+			FeatureName: (*el)[1].val,
+			IsSparse:    false,
+			Shape:       []int{shape},
+			DType:       "double"}, nil
```
tonyyang-svail: Should we allow the users to config `"double"`?

sql/alps_submitter.go
```diff
@@ -0,0 +1,167 @@
...
+	Boundaries   []int
+}
+
+type CrossColumn struct {
+	Keys           []interface{}
+	HashBucketSize int
+}
+
+func resolveFeatureSpec(el *exprlist) (*FeatureSpec, error) {
+	headExpr := (*el)[0].val
```
tonyyang-svail, line 50: Golang coding style: avoid `else if`.
```go
if headExpr == "DENSE" {
    return ...
}

if headExpr == "SPARSE" {
    return ....
}

return ...
```

sql/alps_submitter.go
```diff
@@ -0,0 +1,167 @@
...
+	Fields    []string
+	X         []FeatureSpec
+	Y         FeatureSpec
+}
+
+type FeatureSpec struct {
+	FeatureName string
+	IsSparse    bool
+	Shape       []int
+	DType       string
```
tonyyang-svail, line 30: Do we need a field `Separator` for `DENSE`?

sql/alps_submitter.go
```diff
@@ -0,0 +1,167 @@
...
+func resolveExpr(e *expr) (string, *exprlist) {
+	if e.val != "" {
+		return e.val, nil
+	} else {
+		return "", &e.sexp
+	}
+}
+
+func resolveExprList(el *exprlist) (interface{}, error) {
+	headName, headExprList := resolveExpr((*el)[0])
```
tonyyang-svail, line 101: Code style: avoid else if.
```golang
if headName == "" {
    return ...
}

switch headName {
case "DENSE":
    return ...
default:
    return ...
}
```

sql/alps_submitter_test.go
```diff
@@ -0,0 +1,62 @@
...
+
+func TestColumnResolve(t *testing.T) {
+	a := assert.New(t)
+	r, e := newParser().Parse(testSQLStatement)
+	a.NoError(e)
+
+	result, err := resolveTrainColumns(&r.columns)
+
+	a.NoError(err)
+	a.Equal("FeatureSpec", getFeatureColumnType(result[0]))
```
tonyyang-svail: Could you also check the content of the result?

sql/alps_submitter.go
```diff
@@ -0,0 +1,167 @@
...
+
+type CrossColumn struct {
+	Keys           []interface{}
+	HashBucketSize int
+}
+
+func resolveFeatureSpec(el *exprlist) (*FeatureSpec, error) {
+	headExpr := (*el)[0].val
+	if headExpr == "DENSE" {
+		shape, err := strconv.Atoi((*el)[2].val)
```
tonyyang-svail, line 58: Check the length of `el`, `(*el)[2]`will panic if a user writes `DENSE(c2)`.

sql/alps_submitter.go
uuleon: Fixed

sql/alps_submitter.go
uuleon: Fixed

sql/alps_submitter.go
uuleon: Fixed

sql/alps_submitter.go
uuleon: Add a TODO here, let's make the POC working firstly.

sql/alps_submitter.go
uuleon: Fixed

sql/alps_submitter.go
uuleon: Fixed

sql/alps_submitter.go
uuleon: Fixed

sql/alps_submitter.go
uuleon: Yes, just added a commit for this kind of situation, please take a review. @tonyyang-svail

sql/alps_submitter_test.go
uuleon: Fixed.

sql/alps_submitter.go
```diff
@@ -6,18 +6,28 @@ import (
 	"strconv"
 )
 
-type GraphConfig struct {
+const (
+	SPARSE  = "SPARSE"
```
weiguoz: In Go, we encourage using snake_case for constant. ([mixed caps](https://github.com/golang/go/wiki/CodeReviewComments#mixed-caps)).

uuleon: Fixed

weiguoz:
LGTM

weiguoz (approved):

PR #286, opened 2019-04-18
Code generation for feature columns in alps submitter

uuleon:


sql/alps_submitter.go
```diff
@@ -46,6 +51,32 @@ type CrossColumn struct {
...
 
+func (nc *NumericColumn) GenerateCode() string {
+	return fmt.Sprintf("tf.feature_column.numeric_column(\"%s\", shape=(%d,))", nc.Key, nc.Shape)
+}
+
+func (bc *BucketColumn) GenerateCode() string {
+	return fmt.Sprintf(
+		"tf.feature_column.bucketized_column(%s, boundaries=%s)",
+		bc.SourceColumn.GenerateCode(),
+		strings.Join(strings.Split(fmt.Sprint(bc.Boundaries), " "), ","))
```
tonyyang-svail, line 62: Why not just `fmt.Sprint(bc.Boundaries)`?

sql/alps_submitter_test.go
```diff
@@ -64,13 +66,7 @@ func TestColumnResolve(t *testing.T) {
 	a.Equal(10, nl2.Shape)
 	a.Equal([]int{1, 10}, bl.Boundaries)
 
-	a.Equal("NumericColumn", getFeatureColumnType(cl.Keys[1]))
```
tonyyang-svail, line 67: why removing the following assertion?

sql/alps_submitter.go
uuleon: The `fmt.Sprint` results a string with space separator which I want it to be a comma .

sql/alps_submitter_test.go
uuleon: That is a previous mistake. All keys of `cross` must be either string or categorical column except HashedCategoricalColumn, given `NumericColumn` which is not supported.

sql/alps_submitter_test.go
```diff
@@ -12,7 +12,7 @@ const (
 		WITH n_classes = 3 
 		COLUMN 
 			DENSE(c2, 5, comma), 
-			cross([BUCKET(NUMERIC(c1, 10), [1, 10]), NUMERIC(c4, 9), c5], 20) 
+			cross([BUCKET(NUMERIC(c1, 10), [1, 10]), c5], 20)
```
tonyyang-svail, line 15: Should we change `cross` to `CROSS` to make the feature column naming consistent?

tonyyang-svail (approved):

sql/alps_submitter_test.go
uuleon: The purpose of this is to check if resolver can handle lowercase correctly.

PR #291, opened 2019-04-20
resolve attribute and estimator code gen

uuleon:


sql/alps_submitter.go
```diff
@@ -33,7 +34,7 @@ type featureSpec struct {
 }
 
 type FeatureColumn interface {
-	GenerateCode() string
+	GenerateCode() (string, error)
```
tonyyang-svail, line 37: I like this revision. :)

tonyyang-svail (approved):
LGTM.

PR #295, opened 2019-04-23
Update installation demo

hungry1526:


wangkuiyi (approved):
I like all the edits! Thank you, @hungry1526 for the insights of restructuring the documentation.

PR #303, opened 2019-04-24
Use upper case for keyword predict and some text edits. 

hungry1526:
1) Adding the last commit for Readme.md

2) It would be nice to add some API docs for SQLFlow keywords like predict and train in the future, if SQLFlow is developed for developers instead of business analysts. 

3) Going through a new check-in path from checking out remove develop branch, merging conflicts locally, git push to remote personal repo, and finally PR to main repo.


wangkuiyi (approved):
Thank you, @hungry1526, for being such detail-oriented, which would definitely lead our project to be an excellent one!

PR #304, opened 2019-04-25
Add hive db-api in codegen

weiguoz:
Here still has an issue,
Consider a query: `SELECT * FROM churn.train`
Hive returns a schema: `[('train.customerid', 'VARCHAR_TYPE', None, None, None, None, True), ...]`, notice:
1. the column name train.customerid has a prefix: *train*
Our parser `sql.y` and MySQL return `customerId`
2. the column name consists of lowercases always
3. the field type name with a suffix: *_TYPE

Especially the first 2 points are incompatible with our parser `sql.y`. These will cause a mismatch [in  `codegen`](https://github.com/sql-machine-learning/sqlflow/blob/develop/sql/codegen.go#L62).

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -5,10 +5,12 @@ RUN apt-get install -y git
 RUN apt-get install -y wget curl
 RUN apt-get install -y python3-pip
 
+ENV PIP_INDEX_URL="http://pypi.douban.com/simple/" PIP_TRUSTED_HOST="pypi.douban.com"
 RUN pip3 install --upgrade pip
 RUN pip3 install grpcio-tools
 RUN pip3 install tensorflow
 RUN pip3 install mysql-connector-python
+RUN pip3 install pyhive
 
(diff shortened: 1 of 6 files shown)


wangkuiyi (dismissed):
Thanks for the fix! @weiguoz

tonyyang-svail (dismissed):
LGTM.

Dockerfile.dev
```diff
@@ -5,10 +5,12 @@ RUN apt-get install -y git
 RUN apt-get install -y wget curl
 RUN apt-get install -y python3-pip
 
+ENV PIP_INDEX_URL="http://pypi.douban.com/simple/" PIP_TRUSTED_HOST="pypi.douban.com"
```
tonyyang-svail: We probably shouldn't modify pip index URL in the Dockerfile.

tonyyang-svail (dismissed):

Dockerfile.dev
weiguoz: agree

tonyyang-svail (approved):
LGTM.

PR #311, opened 2019-04-29
run through local train in alps

uuleon:


sql/alps_submitter.go
```diff
@@ -329,30 +381,185 @@ func generateEstimatorCreator(estimator string, attrs []*attribute) (string, err
...
+		Y:                    y.ToString(),
+		OdpsConf:             odpsMap,
+		TrainSpec:            trainMap,
+		DatasetConf:          datasetMap,
+		FeatureColumnCode:    fcCode,
+		EstimatorCreatorCode: estimatorCode}, nil
+}
+
+func genALPS(w io.Writer, pr *extendedSelect) error {
+	r, e := newAlpsFiller(pr)
```
wangkuiyi: In `newAlpsFiller`, the system is referred to as `Alps`; in `genALPS`, it is `ALPS`. Which one should we use?

To my understanding, if the system is named after the name of the mountain, it should be `Alps`; or, if it is a synonym, it should be `ALPS`.

sql/alps_submitter.go
```diff
@@ -329,30 +381,185 @@ func generateEstimatorCreator(estimator string, attrs []*attribute) (string, err
...
+
+    experiment = Experiment(
+        user="sqlflow",
+		engine=LocalEngine(),
+		{{if .TrainSpec.max_steps}}
+        train=TrainConf(input=ds, max_steps={{.TrainSpec.max_steps}}),
+		{{else}}
+		train=TrainConf(input=ds),
+		{{end}}
+        scratch_dir="./scratch",
```
wangkuiyi: Is the `scratch_dir` the current working directory (`cwd`)? 

If so, it seems that the function `genALPS` should take an addition parameter `cwd`  and uses the value to replace the value of `scratch_dir` here.

If not, please make sure that multiple locally-running Alps jobs spawned from the same SQLFlow server process would not write to the same `./scratch` directory.

sql/alps_submitter.go
```diff
@@ -329,30 +381,185 @@ func generateEstimatorCreator(estimator string, attrs []*attribute) (string, err
 	return fmt.Sprintf("tf.estimator.%s(%s)", estimator, strings.Join(cl, ",")), nil
 }
 
-func genALPS(w io.Writer, pr *extendedSelect, fts fieldTypes, db *DB) error {
-	// TODO(uuleon): codegen alps
+func newAlpsFiller(pr *extendedSelect) (*alpsFiller, error) {
+	fcList, fsMap, err := resolveTrainColumns(&pr.columns)
```
wangkuiyi, line 387: Will Alps support prediction/inference soon? 

If so, we might want to add a check here

```go
if pr.training {
```

If not, we might not need the function `submitALPS` in addition to `trainALPS`.

wangkuiyi (commented):
Thanks to @uuleon for this PR. LGTM!

Just a few comments about the API.

Please @tonyyang-svail complete the review.

sql/alps_submitter.go
uuleon: You are right, they should be unified into `ALPS`

sql/alps_submitter.go
uuleon: `scratch_dir` is the directory where `checkpoint` and `model` are stored. Fixed directories are really inappropriate. Thank you for pointing out.

sql/alps_submitter.go
uuleon: Yes, ALPS will support inference in the next few days. This check already exists in the `submitALPS` function. To avoid misunderstanding, what do you think about renaming `newAlpsFiller` to `newALPSTrainerFiller`?

sql/alps_submitter.go
```diff
@@ -329,30 +383,191 @@ func generateEstimatorCreator(estimator string, attrs []*attribute) (string, err
...
+	}
+
+	y := &featureSpec{
+		FeatureName: pr.label,
+		IsSparse:    false,
+		Shape:       []int{1},
+		DType:       "int",
+		Delimiter:   ","}
+
+	dir, err := ioutil.TempDir("/tmp", "alps_scratch_dir")
```
tonyyang-svail, line 448: It would great to remove the temporary directory after the training. Maybe move these lines to `trainALPS` so that we can remove it using defer.
```go
dir, err := ioutil.TempDir("/tmp", "alps_scratch_dir")
if e != nil {
	return e
}
defer os.RemoveAll(dir)
```

tonyyang-svail (approved):
LGTM.

sql/alps_submitter.go
tonyyang-svail: Also, is it possible to combine `dir` and `cwd`?

uuleon: The `scratch_dir` will be deleted after model saving in the next PR.

uuleon: > Also, is it possible to combine `dir` and `cwd`?

In my opinion, no matter which `submitter` is used, the "current working directory" which is stateful should be managed (created, saved, deleted) by `runExtendedSQL` in SQLFlow, but the `scratch_dir` which is an inner attribute in `ALPS submitter`.

PR #319, opened 2019-04-29
Support run extendedSQL for hive

### LINKED ISSUE #318: Hold "go-describeTable" to codegen
In SQLFlow, we use "describeTable" to verify fields, read field type and gen python code. Such "describeTable" is read from database engine by [database/sql](https://golang.org/pkg/database/sql/) driver. We mark it as "go-describeTable".

Meanwhile, we use [db api](https://www.python.org/dev/peps/pep-0249/) to read "describeTable" and access database. We mark it as "python-describeTable".     

Always, "python-describeTable" equals to "go-describeTable", but differ to sqlflow parser(sql.y). And each database has it's own "describeTable", as talked in ["Add hive db-api in codegen"](https://github.com/sql-machine-learning/sqlflow/pull/304#issue-273521061).

statement:     
```sql
SELECT *   
FROM iris.train 
TRAIN DNNClassifier
WITH n_classes = 3, hidden_units = [10, 20]
COLUMN train.sepal_length,train.sepal_width,train.petal_length,train.petal_width
LABEL class
```

- hive

|field|type name|
|-|-|
|train.sepal_length|FLOAT_TYPE|
|train.petal_length|FLOAT_TYPE|

Field name comes with a `table name` as prefix

- mysql

|field|type name|
|-|-|
|sepal_length|FLOAT|
|petal_length|FLOAT|

---
If "go-describeTable" used to verify merely, the database description may be lost in `codegen`. The python code complains field not found. To avoid this, we should use `sql.y`'s description to verify, and hold "go-describeTable" to `codegen`.

weiguoz:
fix #318
 
TODO:
- need more tests


Dockerfile
```diff
@@ -4,7 +4,7 @@ RUN apt-get update
 RUN apt-get install -y python3-pip
 
 RUN pip3 install --upgrade pip
-RUN pip3 install tensorflow mysql-connector-python pyhive jupyter sqlflow
+RUN pip3 install tensorflow mysql-connector-python thrift pyhive jupyter sqlflow
 # Fix jupyter server "connecting to kernel" problem
 # https://github.com/jupyter/notebook/issues/2664#issuecomment-468954423
 RUN pip3 install tornado==4.5.3
```
hungry1526: Is it also possible to incorporate this line?

sql/codegen.go
```diff
@@ -46,6 +41,21 @@ type filler struct {
...
 }
 
+// TODO(tonyyang): This is currently a quick hack to map from SQL
+// field types to feature types.  We will enhance it to support more
+// complex cases like cross features.
+func newColumnType(n, t string) *columnType {
+	return &columnType{n, t}
+}
+
+func translateColumnType(ct *columnType) columnType {
```
typhoonzero: how about use a constant map to define the type transform, like:
```
var column2feature = map[string]string{
  "FLOAT":            "numeric_column",
  "FLOAT_TYPE" : "numeric_column"   // for hive only
}
```

sql/codegen.go
```diff
@@ -95,15 +104,24 @@ func newFiller(pr *extendedSelect, fts fieldTypes, db *DB) (*filler, error) {
 		}
 		sa := strings.Split(cfg.Addr, ":")
 		r.Host, r.Port, r.Database = sa[0], sa[1], cfg.DBName
-		// FIXME(weiguo): make gohive support Auth
-		r.User, r.Password, r.Auth = cfg.User, cfg.Passwd, "NOSALS"
+		// FIXME(weiguo): 1 make gohive support Auth; 2 remove tail semicolon
```
typhoonzero: seems No.2 is done already?

weiguoz: nod

weiguoz:
Apologize this pr is a bit complicated. Most commits here support run `extendedSQL` for hive.
It contains as below in details:
1. Enhance database compatible, like type and field name are as different as SQL engines. [ issue 318：describeTable](https://github.com/sql-machine-learning/sqlflow/issues/318)
1. Change hive [dbapi](https://www.python.org/dev/peps/pep-0249/) from [dropbox/pyhive](https://github.com/dropbox/PyHive) to [cloudera/impyla](https://github.com/cloudera/impyla) in `codegen`, to avoid this [executemany bug](https://github.com/dropbox/PyHive/issues/50)

sqlfs/writer.go
```diff
@@ -56,31 +51,26 @@ func (w *Writer) Write(p []byte) (n int, e error) {
...
+
+	if len(w.buf) > 0 {
+		block := base64.StdEncoding.EncodeToString(w.buf)
+		query := fmt.Sprintf("INSERT INTO %s (id, block) VALUES(%d, '%s')",
+			w.table, w.flushID, block)
+		if _, e := w.db.Exec(query); e != nil {
+			return fmt.Errorf("flush to %s, error:%v", w.table, e)
+		}
+		w.buf = w.buf[:0]
+		w.flushID++
```
typhoonzero, line 73: Just curious that if we are writing multiple versions of model parameters to DB ( have many flushIDs) then how to determine which model to use when running prediction?

typhoonzero: Is there already checks that can enure allways writing parameters of the same model? Like if a user trained a `DNN` model first, then he change the model type to `LR` but didn't change the table name where we save the model parameters?

sqlfs/table.go
```diff
@@ -8,20 +8,27 @@ import (
...
 // database if necessary.
-func createTable(db *sql.DB, table string) error {
-	// FIXME(tony): HIVE and ODPS don't support AUTO_INCREMENT
-	// FIXME(tony): Hive and ODPS don't support BLOB, use BINARY instead
-	stmt := fmt.Sprintf("CREATE TABLE IF NOT EXISTS %s (id INT AUTO_INCREMENT, block BLOB, PRIMARY KEY (id))", table)
+func createTable(db *sql.DB, driver, table string) error {
+	// HIVE and ODPS don't support AUTO_INCREMENT
+	// Hive and ODPS don't support BLOB, use BINARY instead
+	var stmt string
+	if driver == "mysql" || driver == "sqlite3" {
```
typhoonzero, line 15: Maybe we need `DriverType` type to unify these checks in order to avoid things like coding mistakes like `"Hive" == "hive"`

Dockerfile.dev
```diff
@@ -21,7 +21,8 @@ RUN /bin/bash -c "source activate sqlflow-dev && python -m pip install \
     grpcio-tools \
     tensorflow==2.0.0-alpha0 \
     mysql-connector-python \
-    pyhive \
+    impyla \
+    thrift \
```
tonyyang-svail: What is thrift used for?

sql/codegen.go
```diff
@@ -46,6 +40,30 @@ type filler struct {
...
+	}
+	ctype, e := universalizeColumnType(driverName, ct.Type)
+	if e != nil {
+		return nil, e
+	}
+
+	if ctype == "FLOAT" {
+		return &columnType{ident, "numeric_column"}, nil
+	}
+	return &columnType{ident, ""}, nil
```
tonyyang-svail: We should probably return an error if we can't map a `ctype` to a feature column, so that we can make error message like https://github.com/sql-machine-learning/sqlflow/issues/350 more informative.

sqlfs/table.go
weiguoz: Good catch.
This will be fixed in [Unify DriverType name](https://github.com/sql-machine-learning/sqlflow/issues/374).

sql/codegen.go
```diff
@@ -95,15 +112,22 @@ func newFiller(pr *extendedSelect, fts fieldTypes, db *DB) (*filler, error) {
 		}
 		sa := strings.Split(cfg.Addr, ":")
 		r.Host, r.Port, r.Database = sa[0], sa[1], cfg.DBName
-		// FIXME(weiguo): make gohive support Auth
-		r.User, r.Password, r.Auth = cfg.User, cfg.Passwd, "NOSALS"
+		r.User, r.Password = cfg.User, cfg.Passwd
+		r.StandardSelect = removeLastSemicolon(r.StandardSelect)
```
tonyyang-svail, line 110: Why removing the last semicolon?

Dockerfile.dev
weiguoz: `pyhive` relies on `thrift`, or a `ModuleNotFoundError` will be reported, but `impyla` does not.
Since we won't use `pyhive` at this moment, `thrift` should be removed.

sql/verifier.go
```diff
@@ -7,7 +7,7 @@ import (
 
 // fieldTypes[field][table]type.  For more information, please check
 // verifier_test.go.
-type fieldTypes map[string]map[string]string
+type fieldTypes map[string]map[string](*columnType)
```
tonyyang-svail: why do we need this change? Looks like `columnType.Name` is redundant.

sqlfs/writer.go
```diff
@@ -2,51 +2,46 @@ package sqlfs
...
 		return nil, fmt.Errorf("create: %v", e)
 	}
-	return &Writer{db, table, make([]byte, 0, bufSize), nil}, nil
+	return &Writer{db, table, make([]byte, 0, bufSize), 0}, nil
 }
 
 // Write write bytes to sqlfs and returns (num_bytes, error)
 func (w *Writer) Write(p []byte) (n int, e error) {
 	n = 0
-	for len(p) > 0 {
```
tonyyang-svail: Could you explain the necessity of the following changes?

sql/codegen.go
weiguoz: In `hive`, a statement with the last semicolon will cause a `ParseException` error.

sql/verifier.go
weiguoz: You're right.
It is an outdated definition.

sqlfs/writer.go
weiguoz: Since `p` is unchangeable in a loop. Then we just reuse the result of `len()`.

sql/codegen.go
tonyyang-svail: return error like "unsupported field XXX with type XXX"

sql/codegen.go
```diff
@@ -46,6 +40,23 @@ type filler struct {
...
+	}
+	ctype, e := universalizeColumnType(driverName, ct)
+	if e != nil {
+		return nil, e
+	}
+
+	if ctype == "FLOAT" || ctype == "INT" || ctype == "DOUBLE" {
+		return &columnType{ident, "numeric_column"}, nil
+	}
+	return nil, fmt.Errorf("unsupported field type [%s]", ctype)
```
tonyyang-svail: Also include field name `ident` in the error message?

sqlfs/writer.go
```diff
@@ -2,51 +2,46 @@ package sqlfs
...
-			fill = len(p)
+		if fill > nBytes {
+			fill = nBytes
 		}
 		w.buf = append(w.buf, p[:fill]...)
 		p = p[fill:]
 		n += fill
 		if len(w.buf) >= bufSize {
 			if e = w.flush(); e != nil {
-				return n, fmt.Errorf("writer flush failed: %v", e)
```
tonyyang-svail: This might be a leftover for debugging. Could you change back to `return n, fmt.Errorf("writer flush failed: %v", e)`? A writer has many functions, the error message should include the function name. :)

tonyyang-svail (dismissed):

sqlfs/writer.go
tonyyang-svail: I see. Since `len()` is O(1), we might not need this level of optimization. :)

sqlfs/writer.go
weiguoz: We are considering using [independent storage media](https://github.com/sql-machine-learning/sqlflow/issues/377) to save models. Let's keep on discussing in that thread.

tonyyang-svail (approved):
LGTM. Excellent job!

Issue #326, opened 2019-04-30
"Quick Overview"


llxxxll:
A "Quick Overview" is needed on the [home](https://github.com/sql-machine-learning/sqlflow#how-to-use-sqlflow) to demonstrate how SQLFlow can quickly complete the training and prediction tasks.

Examples of training：
```sql
SELECT *
FROM iris.train
TRAIN DNNClassifier
LABEL class
INTO my_dnn_model;
```

Examples of prediction:
```sql
SELECT *
FROM iris.test
PREDICT iris.predict.class
USING my_dnn_model;
```

PR #329, opened 2019-05-04
Fix path issue in installation.md

hungry1526:
Small fix in the doc with no issue created, following #326


tonyyang-svail (approved):
LGTM.

PR #331, opened 2019-05-04
Fix 330: dealing with multiple python versions and pip

### LINKED ISSUE #330: Continuous integration fails
**Describe the bug**
CI build [failed](https://travis-ci.org/sql-machine-learning/sqlflow/builds/528104956) at `Step 6/34 : RUN pip3 install --upgrade pip`.

**To Reproduce**
`docker build -t sqlflow:dev -f Dockerfile.dev .`

Screen Shot 2019-05-04 at 7 36 10 PM

tonyyang-svail:
Fix #330

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,19 +1,33 @@
 FROM ubuntu:16.04
 
 RUN apt-get update
-RUN apt-get install -y wget curl python3-pip
+RUN apt-get install -y curl
 
-# Make python3 as default Python interpreter
-RUN ln -s /usr/bin/python3 /usr/bin/python
-RUN chmod +x /usr/bin/python
+# Miniconda - Python 3.6, 64-bit, x86, latest
+ARG CONDA_OS=Linux
+RUN curl -sL https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -o mconda-install.sh && \
+    bash -x mconda-install.sh -b -p miniconda && \
+    rm mconda-install.sh
+ENV PATH="/miniconda/bin:$PATH"
 
-# Upgrade pip using the bootstrapping method
-# https://github.com/pypa/pip/issues/5255#issuecomment-381702613
-RUN curl https://bootstrap.pypa.io/get-pip.py | python
-RUN python -m pip install tensorflow==2.0.0-alpha0 mysql-connector-python pyhive jupyter sqlflow
+ARG CONDA_ADD_PACKAGES=""
+RUN conda create -y -q -n sqlflow-dev python=3.6 ${CONDA_ADD_PACKAGES}
+
+RUN echo ". /miniconda/etc/profile.d/conda.sh" >> ~/.bashrc && \
+    echo "source activate sqlflow-dev" >> ~/.bashrc
+
+ARG PIP_ADD_PACKAGES=""
+RUN /bin/bash -c "source activate sqlflow-dev && python -m pip install \
+    tensorflow==2.0.0-alpha0 \
+    mysql-connector-python \
+    pyhive \
+    jupyter \
+    sqlflow \
+    ${PIP_ADD_PACKAGES} \
+    "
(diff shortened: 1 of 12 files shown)


wangkuiyi (dismissed):

Dockerfile
```diff
@@ -1,13 +1,19 @@
 FROM ubuntu:16.04
 
 RUN apt-get update
-RUN apt-get install -y python3-pip
+RUN apt-get install -y wget curl python3-pip
```
zou000: FYI: I found using a python version manager, such as conda, pyenv,  is more convenient.  For example, we are using this approach in tf-plus: https://github.com/tensorflow/io/blob/master/dev/Dockerfile#L19

tonyyang-svail: @zou000 Thanks for the suggestion. Indeed, adding `source activate tfio-dev` to `~/.bashrc` neatly isolate the Linux built-in Python2 and our Python3. This approach avoids the error-prone
```bash
# Make python3 as default Python interpreter
RUN ln -s /usr/bin/python3 /usr/bin/python
RUN chmod +x /usr/bin/python
```

Dockerfile
```diff
@@ -1,13 +1,19 @@
...
 
-RUN pip3 install --upgrade pip
-RUN pip3 install tensorflow==2.0.0-alpha0 mysql-connector-python pyhive jupyter sqlflow
+# Make python3 as default Python interpreter
```
hungry1526: Should we also updated the doc telling user to use python3 and the latest pip version?

tonyyang-svail:
CI [failed](https://travis-ci.com/sql-machine-learning/sqlflow/builds/110638465) at 
```
=== RUN   TestExecutorTrainAndPredictDNN
FAIL	github.com/sql-machine-learning/sqlflow/sql	0.265s
```

Update: `TestExecutorTrainAndPredictDNN` failed because `go test` was not able to find `tensorflow` Python package. Because we enable the Python virtual environment in `.bashrc`, but `.bashrc` is skipped on non-interactive bashes such as `docker run image_names bash test.sh`.

Dockerfile
tonyyang-svail: @zou000  enabling python env in `.bashrc` has certain drawbacks. For example, if you do `docker run image_name bash test.sh` i.e. non-interactive bash, `.bashrc` is skipped, so the python env is not activated. One workaround is to define the `source activate` in `entrypoint`.

Dockerfile
tonyyang-svail: Absolutely.

Dockerfile
zou000: I can think of 2 other ways:

1. Invoke bash interactively e.g. ` docker run image_name bash -i test.sh`
2. The idea of a pyenv is very simple, it just needs to find the python binary under the env root dir, e.g. in your case, you just need to modify PATH:
```bash
ENV PATH=/miniconda/envs/sqlflow-dev/bin:$PATH
```  
I like the PATH approach better. You can verify the result by:
```
docker run image_name python -c 'import sys; print(sys.path)'

```

Dockerfile.dev
```diff
@@ -1,15 +1,29 @@
...
-RUN apt-get install -y wget curl
-RUN apt-get install -y python3-pip
+RUN apt-get install -y git wget curl bzip2 build-essential
 
-RUN pip3 install --upgrade pip
-RUN pip3 install grpcio-tools
-RUN pip3 install tensorflow==2.0.0-alpha0
-RUN pip3 install mysql-connector-python
-RUN pip3 install pyhive
+# Miniconda - Python 3.6, 64-bit, x86, latest
```
hungry1526, line 6: If we are duplicating some code in .Dev, can we group the duplicated ones and add some notes in case when they need to be synchronized?

weiguoz (approved):
LGTM

weiguoz (approved):

PR #342, opened 2019-05-06
Wrap the deferred timer in a function

### LINKED ISSUE #341: Wrap the deferred timer in a function
The statement after the `defer` is evaluated at the time of the first encounter. For example,

https://github.com/sql-machine-learning/sqlflow/blob/ff96c186817ef5aefb1f1cbe0e0dce9176f379ad/sql/executor.go#L177-L179

This code snippet will always log `elapsed: 0 second`. The correct usage is to wrap the deferred timer in a function, such as
```go
defer func() {
        fmt.Printf("elapsed:%v", time.Since(startAt))
}()
```

tonyyang-svail:
Fix #341

### DIFF
--- a/sql/executor.go
+++ b/sql/executor.go
@@ -57,6 +57,9 @@ func runQuery(slct string, db *DB) *PipeReader {
 		err := func() error {
 			startAt := time.Now()
 			log.Infof("Starting runStanrardSQL:%s", slct)
+			defer func() {
+				log.Infof("runQuery finished, elapsed: %v", time.Since(startAt))
+			}()
 
 			rows, err := db.Query(slct)
 			if err != nil {
@@ -111,7 +114,6 @@ func runQuery(slct string, db *DB) *PipeReader {
 					return e
 				}
 			}
-			log.Infof("runQuery finished, elapsed: %v", time.Since(startAt))
 			return nil
 		}()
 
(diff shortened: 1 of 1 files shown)


sql/executor.go
```diff
@@ -134,7 +136,10 @@ func runExec(slct string, db *DB) *PipeReader {
 
 		err := func() error {
 			startAt := time.Now()
-			log.Infof("Starting runStanrardSQL1:%s", slct)
+			log.Infof("Starting runExec:%s", slct)
```
typhoonzero: Maybe we only need time elapsed for the `slct` statement in the log, so this line can be `Debugf`? And below elapsed log can log together with `slct`, or it maybe harder to figure out which statement this time was.

tonyyang-svail: Sounds good. :)

sql/executor.go
```diff
@@ -55,8 +55,9 @@ func runQuery(slct string, db *DB) *PipeReader {
 		defer wr.Close()
 
 		err := func() error {
-			startAt := time.Now()
-			log.Infof("Starting runStanrardSQL:%s", slct)
+			defer func(startAt time.Time) {
+				log.Debugf("runQuery finished, elapsed:%v", time.Since(startAt))
```
typhoonzero: Maybe also output the `slct` string?

typhoonzero (approved):

sql/executor.go
tonyyang-svail: My bad..

PR #346, opened 2019-05-07
Fix a dead link in the document pipe.md, and update to support latest precommit

typhoonzero:
- Fix a deadlink in the document: https://github.com/sql-machine-learning/sqlflow/blob/develop/doc/pipe.md
- update `.pre-commit-config.yaml` to be compatible to latest pre-commit versions (> 1.7.0), use `rev` instead of sha


tonyyang-svail (approved):
Excellent!

PR #349, opened 2019-05-07
fixed small errors

0xflotus:


llxxxll (approved):
Thank you, @0xflotus, for being such detail-oriented, which would lead our project to be an excellent one!

Issue #347, opened 2019-05-07
Improvement ideas for the installation guide.


helinwang:
1. https://sql-machine-learning.github.io/sqlflow/example/datasets/README.md:
Maybe put the trouble shooting tips inside "Popularize Datasets to MySQL" section. I encountered the same error, retried and it worked. The trouble shooting section perhaps is too far away to be found.

2. Manually building another image for loading data seems a barrier for trying out SQLFlow. Maybe create playground docker image that contains the database, MySQL server, Jupiter notebook file, and everything else? To reduce the steps for trying out the demo.

3. I was going through the second step of "Running your first SQLFlow query" in
https://sql-machine-learning.github.io/sqlflow/doc/installation. Opened http://127.0.0.1:8888, but could not find any Jupiter notebook file. I had to manually copy https://sql-machine-learning.github.io/sqlflow/example/jupyter/example.ipynb into the container.

4. After getting into the notebook, when running `%%sqlflow
show databases;` inside the notebook, I got
`_Rendezvous: <_Rendezvous of RPC that terminated with:
status = StatusCode.UNAVAILABLE
details = "Connect Failed"`
Seems the py notebook container can not connect to the data container. I tried some solutions but ran out of time in the end. I think putting everything into a same docker image would help in this case.

Keep up the great work!

hungry1526:
1-4 will be addressed in a recent PR , thanks for the feedback.

typhoonzero:
We are moving to the [cloud playground](https://playground.sqlflow.tech/) for quick start.

PR #353, opened 2019-05-07
Add all in one quickstart docker image

typhoonzero:
Inspired by https://github.com/sql-machine-learning/sqlflow/issues/347 No.2, add all-in-one docker image to speed up user tryouts.

***NOTE:*** should upload image `sqlflow/quickstart:latest` to dockerhub before merging.


doc/installation.md
```diff
@@ -8,31 +8,21 @@ issue is welcome. :)
...
 
 1. Install [Docker Community Edition](https://docs.docker.com/install/) on your Macbook.
-1. Build and run a dockerized MySQL server (In this example, called sqlflowdata) following [example/datasets](/example/datasets). Note that there is no need to install a local MySQL server, in which case you will have a port conflict in 3306. 
-1. Pull the latest SQLFlow Docker image: 
+1. Pull the latest SQLFlow "all-in-one" Docker image, which contains pre-built SQLFlow
+   binary, sample datasets (under `example/datasets`), and jupyter notebook.
 
    ```
-   docker pull sqlflow/sqlflow:latest
+   docker pull sqlflow/quickstart:latest
```
tonyyang-svail, line 15: Shall we release `sqlflow/quickstart:latest` to DockerHub on every successful build? This can be done at `scripts/deploy.sh`.

tonyyang-svail (dismissed):
Excellent!

doc/installation.md
typhoonzero: Sure, will add!

doc/installation.md
```diff
@@ -8,31 +8,21 @@ issue is welcome. :)
...
-
-1. Start a Docker container that runs SQLFlow server and Jupyter Notebook. If you are
-   using Docker for Linux, please change `host.docker.internal:3306` to `localhost:3306`.
+1. Simply type the below command to start the service:
 
    ```
-   docker run --rm -it -p 8888:8888 sqlflow/sqlflow:latest \
-   bash -c "sqlflowserver --datasource='mysql://root:root@tcp(host.docker.internal:3306)/?maxAllowedPacket=0' &
-   SQLFLOW_SERVER=localhost:50051 jupyter notebook --ip=0.0.0.0 --port=8888 --allow-root"
+   docker run --rm -p 8888:8888 sqlflow/quickstart:latest
```
hungry1526: Should we leave option -RM up to user? In some cases, developers want to re-run docker

typhoonzero: good point, or shall we run the container in the background, adding `-d` so that the service will survive long?

hungry1526: I personally agree, but removing -rm should also sync up other folks. I personally want to play around for a while but forgot we using -rm when it starts. 

We should probably simplified all the possible options for each commands and leave most of decisions to users with sufficient notes.

typhoonzero: Seems `docker start` to restart the container will run the script again when first-time `docker run` ran: https://stackoverflow.com/questions/26032929/does-docker-start-execute-the-cmd-command. So removing `--rm` should be fine.

Adding `-d` will cause user to have to run `docker logs` to get the notebook token, so will remove `--rm` in this PR only.

example/startall.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+# Start mysqld
+docker-entrypoint.sh mysqld &
+sleep 1
```
hungry1526: Question, is 1s enough? What are the factors for length of sleep time?

doc/installation.md
hungry1526: Sounds good.

example/startall.sh
typhoonzero: below lines will wait until the connection is ready. So sleep 1 here is OK~

hungry1526: Can we then remove line 5 as we have the until loop to wait for the DB connection.

typhoonzero: Some times mysqld starts very fast, so line 5 here can reduce the annoying error logs.

hungry1526: I see, we could use 1s for sleep time for now, but it would be nice to explicitly catch/handle the MySQL connection error instead of waiting for 1s. Guess it is hard to decide if 1s is enough for the wait before entering into until execution. 

:-)

scripts/deploy.sh
```diff
@@ -14,3 +14,9 @@ docker push sqlflow/sqlflow:$DOCKER_TAG
 
 docker tag sqlflow:dev sqlflow/sqlflow:dev
 docker push sqlflow/sqlflow:dev
+
+# NOTE: this script should be called by .travis.yml
+pushd ./example/
+docker build -t sqlflow/quickstart:$DOCKER_TAG
```
tonyyang-svail: How about we name it `sqlflow/sqlflow:quickstart`? Because we are naming gohive image as `sqlflow/gohive`, `sqlflow/quickstart` is not clear which repo it refers to.

example/Dockerfile
```diff
@@ -0,0 +1,32 @@
+# NOTE: This Dockerfile combines all quick start components in a
```
tonyyang-svail: I believe `example/` directory will be moved to `doc/` and all the website content will be built from `doc/`.

So shall we put this Dockerfile to PROJECT_ROOT, right next to other Dockerfiles, and name it `Dockerfile.quickstart`?

scripts/deploy.sh
typhoonzero: Sure.

example/Dockerfile
typhoonzero: Sure.

doc/installation.md
tonyyang-svail: => `docker pull sqlflow/sqlflow:quickstart`

tonyyang-svail (approved):
LGTM! Please update the doc to `docker pull sqlflow/sqlflow:quickstart` in the next PR.

PR #364, opened 2019-05-09
Write default log to stdout

### LINKED ISSUE #362: Log to stdout during continuous integration
To make CI messages more informative, we should log to stdout.

typhoonzero:
Fix https://github.com/sql-machine-learning/sqlflow/issues/362


scripts/test.sh
```diff
@@ -6,4 +6,4 @@ service mysql start
 go generate ./...
 go get -v -t ./...
 go install ./...
-go test -v ./...
+SQLFLOW_log_dir= go test -v ./...
```
weiguoz: I'm not sure what the effect is.
Maybe we just run `go test -v ./...`, because the default `SQLFLOW_log_dir` is ""

tonyyang-svail: Shall we also add `SQLFLOW_log_level=debug`?

typhoonzero: Thanks @weiguoz @tonyyang-svail , Done.

weiguoz (approved):
LGTM

PR #371, opened 2019-05-12
Add go module in SQLFlow

WingGao:


weiguoz:
Would you please add some comments about the purpose of this PR :)

WingGao:
Using go module will make the sqlflow build to be more reliable and reproducible.
And for those who use go module as the dependency management will need this `go.mod` file while use this project from `fork`.

wangkuiyi:
HI @WingGao , thank you very much for this PR! I can see how carefully you listed the dependencies and verified their versions. And I believe we will need the Go module someday in the future.

Unfortunately, SQLFlow is currently under very active development, the dependencies are subject to frequent and significant changes -- no to mention their versions.

How about we add Go modules back when we reach version 1.0 in the next few months when the software is production-ready?

Thank you again for this contribution. We look forward to more from you!

tonyyang-svail:
Kindly close this PR for now.

PR #376, opened 2019-05-13
Add evaluation to alps_submitter

uuleon:
The major work of this PR contains the following:
1. refactoring `alpsFiller` to make it adaptive to train and predict
2. add `Get` method which supports default value to make code generation easier
3. support `eval_spec` setting and train_with_evaluation

### DIFF
--- a/sql/alps_submitter.go
+++ b/sql/alps_submitter.go
@@ -367,7 +367,7 @@ func generateFeatureColumnCode(fcs []interface{}) (string, error) {
 	return fmt.Sprintf("[%s]", strings.Join(codes, ",")), nil
 }
 
-func generateEstimatorCreator(estimator string, attrs []*attribute) (string, error) {
+func generateEstimatorCreator(estimator string, attrs []*attribute, args []string) (string, error) {
 	cl := make([]string, len(attrs))
 	for idx, a := range attrs {
 		code, err := a.GenerateCode()
@@ -376,10 +376,20 @@ func generateEstimatorCreator(estimator string, attrs []*attribute) (string, err
 		}
 		cl[idx] = code
 	}
+	if args != nil {
+		for _, arg := range args {
+			cl = append(cl, arg)
+		}
+	}
 	return fmt.Sprintf("tf.estimator.%s(%s)", estimator, strings.Join(cl, ",")), nil
 }
 
 func newALPSTrainFiller(pr *extendedSelect) (*alpsFiller, error) {
+	scratchDir, err := ioutil.TempDir("/tmp", "alps_scratch_dir_")
+	if err != nil {
+		return nil, err
+	}
+
 	fcList, fsMap, err := resolveTrainColumns(&pr.columns)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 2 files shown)


hungry1526:
@uuleon It would be a good practice to add some detail information in description sector for a particular PR.

uuleon:
@hungry1526 thanks for the reminder,  the description has been updated.

sql/alps_submitter.go
```diff
@@ -22,16 +22,35 @@ const (
...
+	FeatureColumnCode    string
+	TrainSpec            collection
+	EvalSpec             collection
+
+	// Config
+	OdpsConf    collection
+	DatasetConf collection
+}
+
+type collection struct {
```
tonyyang-svail: Not sure we need to add this level of abstraction: `OdpsConf.Dict["Key"]` vs `OdpsConf["Key"]`. I think the latter is more concise. If we need to define a method, such as `Get(key, defaultValue string)`, on `map[string]string`, we can simply give `map[string]string` a name. For example,

```golang
package main

import "fmt"

type collection map[string]string

type collections struct {
	name string
	c2 collection
}

func (c collection) Get(key, fallback string) string {
	if v, ok := c[key]; ok {
		return v
	}
	return fallback
}

func main() {
	config := make(map[string]string)
	config["k1"] = "v1"
	var c collection = config
	fmt.Println(c.Get("k1", "fallback1")) // v1
	fmt.Println(c.Get("k2", "fallback2")) // fallback2
	cs := collections{name: "hi", c2: config}
	fmt.Println(cs) // {hi map[k1:v1]}
}
```
By doing so we can avoid the redundant `Dict` field.

sql/alps_submitter.go
```diff
@@ -441,59 +487,62 @@ func newALPSTrainFiller(pr *extendedSelect) (*alpsFiller, error) {
...
-	dir, err := ioutil.TempDir("/tmp", "alps_scratch_dir")
-	if err != nil {
-		return nil, err
-	}
-
 	return &alpsFiller{
-		TableName:            tableName,
+		IsTraining:           true,
+		TrainInputTable:      tableName,
+		EvalInputTable:       tableName,
```
tonyyang-svail: Add `//FIXME(uuleon): Train and Eval should use different dataset.`

sql/alps_submitter.go
```diff
@@ -376,10 +402,21 @@ func generateEstimatorCreator(estimator string, attrs []*attribute) (string, err
...
+	if args != nil {
+		for _, arg := range args {
+			cl = append(cl, arg)
+		}
+	}
 	return fmt.Sprintf("tf.estimator.%s(%s)", estimator, strings.Join(cl, ",")), nil
 }
 
 func newALPSTrainFiller(pr *extendedSelect) (*alpsFiller, error) {
+	scratchDir, err := ioutil.TempDir("/tmp", "alps_scratch_dir_")
```
tonyyang-svail, line 413: Should we remove `scratchDir` after the training?

sql/alps_submitter.go
uuleon: That is a great idea! Thanks for pointing it out.

sql/alps_submitter.go
uuleon: The `scratchDir` will be deleted until the output model is saved using `sqlfs` which is blocking by `ODPS DB`.

tonyyang-svail: Would you please at least add a TODO?

uuleon: Ok, thanks

sql/alps_submitter.go
```diff
@@ -367,7 +391,7 @@ func generateFeatureColumnCode(fcs []interface{}) (string, error) {
 	return fmt.Sprintf("[%s]", strings.Join(codes, ",")), nil
 }
 
-func generateEstimatorCreator(estimator string, attrs []*attribute) (string, error) {
+func generateEstimatorCreator(estimator string, attrs []*attribute, args []string) (string, error) {
```
tonyyang-svail, line 394: Could you please provide an example on `args []string`?

tonyyang-svail (approved):
LGTM. Great job!

PR #379, opened 2019-05-13
Better CI error messages

### LINKED ISSUE #378: Better CI error messages
In travis.yml, we define the following testing stages 
```
- ls -lah
- docker build -t sqlflow:dev -f Dockerfile.dev .
- docker run --rm -v $GOPATH/src:/go/src -w /go/src/github.com/sql-machine-learning/sqlflow sqlflow:dev pre-commit run -a
# NOTE: mount all $GOPATH into container, then after build and test,
# $GOPATH/bin will have expected binaries.
- docker run --rm -v $GOPATH:/go -w /go/src/github.com/sql-machine-learning/sqlflow sqlflow:dev bash scripts/test.sh
```
Travis CI considers them as independent stages. For example, in [the CI log](https://travis-ci.com/sql-machine-learning/sqlflow/builds/111532246), pre-commit fails so the CI fails. However, you need to scroll up serveral hundred lines to find the reason.

I would suggest putting all the stages in a single `test.sh` file and make each line dependent via `set -e`.

tonyyang-svail:
Fix #378

### DIFF
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -3,10 +3,6 @@ set -e
 
 service mysql start
 
-cat example/datasets/popularize_churn.sql | mysql -uroot -proot
-cat example/datasets/popularize_iris.sql | mysql -uroot -proot
-cat example/datasets/create_model_db.sql | mysql -uroot -proot
-
 go generate ./...
 go get -v -t ./...
 go install ./...
--- a/scripts/test_runner.sh
+++ b/scripts/test_runner.sh
@@ -1,0 +1,10 @@
+#!/bin/bash
+set -e
+
+ls -lah
+docker build -t sqlflow:dev -f Dockerfile.dev .
+docker run --rm -v $GOPATH/src:/go/src -w /go/src/github.com/sql-machine-learning/sqlflow sqlflow:dev pre-commit run -a
+
+# NOTE: mount all $GOPATH into container, then after build and test,
+# $GOPATH/bin will have expected binaries.
+docker run --rm -v $GOPATH:/go -w /go/src/github.com/sql-machine-learning/sqlflow sqlflow:dev bash scripts/test.sh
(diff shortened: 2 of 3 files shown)


scripts/test.sh
```diff
@@ -3,10 +3,6 @@ set -e
 
 service mysql start
 
-cat example/datasets/popularize_churn.sql | mysql -uroot -proot
```
tonyyang-svail, line 6: `sql/db_test.go` is responsible for data population.

scripts/test_runner.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+set -e
```
typhoonzero, line 2: Have you tested adding `set -e` in travis.yml? if that works we may reduce the level of script calls.

scripts/test.sh
typhoonzero: I'm not sure `db_test.go` can always run before `main_test` since `main_test` depend on these sample data.

tonyyang-svail:
> I'm not sure db_test.go can always run before main_test since main_test depend on these sample data.

@typhoonzero You are right. I forgot the test in `cmd/` also needs populated data. My bad.

scripts/test_runner.sh
tonyyang-svail: Good point. Let me try it out.

scripts/test.sh
tonyyang-svail: You are right. I forgot the tests in `cmd/` also need the populated data. My bad.

tonyyang-svail:
Follow @typhoonzero 's suggestion, using `set -e` directly in `travis.yml`. PR submitted in https://github.com/sql-machine-learning/sqlflow/pull/381 .

PR #385, opened 2019-05-15
Reduce unnecessary readme.md

hungry1526:
Resolve issue 384:

1) delete unnecessary readme.md while going through the repo
2) too many readme.md is hard to maintain, therefore removing
3) remove empty space and newline 
4) simplify bug/issue template


README.md
```diff
@@ -53,10 +53,10 @@ Done predicting. Predict table : iris.predict
 - [Installation](doc/installation.md)
 - [Running a Demo](doc/demo.md)
 - [Extended SQL Syntax](doc/syntax.md)
+- [Build from source code](doc/build.md)
```
wangkuiyi: Users don't need to build from source, only contributors need to.

sql/python/README.md
```diff
@@ -1,13 +0,0 @@
-# Python Code Template
-
```
wangkuiyi: I agree that we should merge the content of this file into build from source doc. but I am not sure the docker command in this file runs now. I'd suggest we confirm it works before moving it into the build from source file

sql/python/README.md
```diff
@@ -1,13 +0,0 @@
-# Python Code Template
```
wangkuiyi: I am afraid this file is not ready to merge into the build from source file

wangkuiyi (changes requested):

README.md
hungry1526: Users and developers seem be inter-changable,  I agree we can keep it under contribution.

sql/python/README.md
hungry1526: Will create an issue to track changes. For now, I will undo delete.

hungry1526:
All comments are resolved.

wangkuiyi (approved):

PR #397, opened 2019-05-17
Rename alps_submitter to codegen_alps

uuleon:
The name of alps_submitter doesn't match its content, rename it to make it reasonable.


uuleon:
Currently, not only code generation but also template execution exists in the "ALPS Submitter", delaying the renaming until the template execution has been stripped out.

PR #417, opened 2019-05-21
SQL predict statement supported in ALPS

uuleon:
The major work of this PR contains the following
1. implementation of `newALPSPredictFiller ` function
2. add code generation of `predict` to  the ALPS template

### DIFF
--- a/sql/alps_submitter.go
+++ b/sql/alps_submitter.go
@@ -22,7 +22,7 @@ import (
 	The `train` job clear the `model_dir` at first before training, then the `predict` job restores the output model of last `train` job.
 
 */
-const model_dir = "/tmp/alps_model_dir/"
+const modelDir = "/tmp/alps_model_dir/"
 
 const (
 	sparse    = "SPARSE"
@@ -422,7 +422,7 @@ func generateEstimatorCreator(estimator string, attrs []*attribute, args []strin
 }
 
 func newALPSTrainFiller(pr *extendedSelect) (*alpsFiller, error) {
-	modelDir := model_dir
+	modelDir := modelDir
 
 	fcList, fsMap, err := resolveTrainColumns(&pr.columns)
 	if err != nil {
@@ -511,7 +511,7 @@ func newALPSTrainFiller(pr *extendedSelect) (*alpsFiller, error) {
 }
 
 func newALPSPredictFiller(pr *extendedSelect) (*alpsFiller, error) {
-	modelDir := model_dir
+	modelDir := modelDir
 
 	inputTableName := pr.tables[0]
 	outputTableName := pr.predictClause.into
--- a/sql/alps_submitter.go
+++ b/sql/alps_submitter.go
@@ -504,7 +504,34 @@ func newALPSTrainFiller(pr *extendedSelect) (*alpsFiller, error) {
 }
 
 func newALPSPredictFiller(pr *extendedSelect) (*alpsFiller, error) {
-	return nil, fmt.Errorf("alps predict not supported")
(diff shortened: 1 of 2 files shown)


sql/alps_submitter.go
```diff
@@ -504,7 +511,22 @@ func newALPSTrainFiller(pr *extendedSelect) (*alpsFiller, error) {
 }
 
 func newALPSPredictFiller(pr *extendedSelect) (*alpsFiller, error) {
-	return nil, fmt.Errorf("alps predict not supported")
+	modelDir := modelDir
```
wangkuiyi, line 514: Unnecessary alias?

sql/alps_submitter.go
```diff
@@ -4,12 +4,26 @@ import (
...
+
+	In general, the `model_dir` should be a temporary directory for each `train` job who saves the output model to DB, then
+	the `predict` job restores model from DB.
+
+	But, the `maxCompute` type of DB which `ALPS` used is not supported yet, so a fixed `model_dir` is used here.
+
+	The `train` job clear the `model_dir` at first before training, then the `predict` job restores the output model of last `train` job.
+
+*/
+const modelDir = "/tmp/alps_model_dir/"
```
wangkuiyi, line 25: Would multiple simultaneous Alps jobs share the same directory? If so, would they conflict with each other?

sql/alps_submitter.go
```diff
@@ -661,12 +685,71 @@ if __name__ == "__main__":
...
+			model_dir=model_dir,
+		),
+		model_builder=SQLFlowEstimatorBuilder()
+	)
+
+	return experiment
+
+{{else}}
+
+def inference(model_dir):
```
wangkuiyi, line 699: It seems that this Go template is becoming longer.

It seems to me we need the following stuff in this Python template:

1. the Estimator builder class
1. the train function
1. the predict function
1. the main function

To make it easy to maintain and test, I would move 1., 2., and 3. out from this .go file into a Python file `python/alps_codegen.py`. So we could have `python/alps_codegen_test.py` for unit testing these facilities. To make this happen, we need to remove all template syntax from 1., 2., and 3. to make them legal Python code. I think this should be pretty easy -- we can change template placeholders into function parameters or class data members.

And, I would leave all template syntax in 4., and keep 4. in this .go file.

Following the above steps, this alps_codegen.go file would generate a short .py file which contains only a main function, which, calls functions/classes defined in python/alps_codegen.py.

To make this solution complete, we should remember that we need to pack python/alps_codegen.py into our release Docker image, so it could be called by the generated Alps submitter programs.

I has this idea few months ago and talked to @tonyyang-svail at that time. Now I believe this is a must-go way, because following our feature derivation design, we have to make feature derivation a Python package instead of a Go package, as it must scan the data in order to derive features. As we'd have to have Python package called by submitters, we'd put alps_codegen.py part of this Python package.

@uuleon please feel free to talk to @tonyyang-svail and me for more about this idea.

wangkuiyi (changes requested):

sql/alps_submitter.go
uuleon: Yes，it's unnecessary.

sql/alps_submitter.go
uuleon: You are right. That's why a `FIXME` is added here, and the confliction will be resolved after the DB of `MaxCompute` is available.

sql/alps_submitter.go
uuleon: Thanks for this comment which is reasonable and valuable, I will close this PR and do the refactoring of the code template at first.

uuleon:
Close this PR and do the refactoring of the code template at first.

PR #424, opened 2019-05-22
support keras custom model

typhoonzero:
Depends on: https://github.com/sql-machine-learning/models/pull/7, rerun the test when it merged.


cmd/sqlflowserver/main_test.go
```diff
@@ -273,6 +274,66 @@ FROM iris.predict LIMIT 5;`
 	}
 }
 
+// CaseTrainCustomModel tests using customized models
+func CaseTrainCustomModel(t *testing.T) {
+	a := assert.New(t)
+	trainSQL := `SELECT *
+FROM iris.train
```
hungry1526, line 281: Fix the indentation?

tonyyang-svail: Hi @hungry1526, string constant is supposed to be indented like this. :) Otherwise, the statement will become
```
SELECT *
    FROM iris.train
    ...
```

typhoonzero: @hungry1526 Thanks for review, sorry for the late reply due to I was debugging some errors in this PR. As @tonyyang-svail commented, this indentation is expected, so I'll keep it.

hungry1526: thanks guys.

scripts/test.sh
```diff
@@ -10,6 +10,8 @@ cat example/datasets/create_model_db.sql | mysql -uroot -proot
 
 export SQLFLOW_TEST_DB=mysql
 
+python -c "import sqlflow_models"
```
tonyyang-svail, line 13: Is this a leftover for debugging?

sql/codegen.go
```diff
@@ -238,39 +252,70 @@ for name in feature_column_names:
...
-classifier = tf.estimator.{{.Estimator}}(
-    feature_columns=feature_columns,{{range $key, $value := .Attrs}}
-    {{$key}} = {{$value}},{{end}}
-    model_dir = "{{.Save}}")
+classifier = {{.Estimator}}(
+	feature_columns=feature_columns,{{if .SelfDefined}}{{range $key, $value := .Attrs}}
+	{{$key}} = {{$value}},{{end}})
+	{{else}}{{range $key, $value := .Attrs}}
+	{{$key}} = {{$value}},{{end}}
+	model_dir = "{{.Save}}")
```
tonyyang-svail, line 261: Suggested feature, maybe adding in the future PR:
Is it possible to add `model_dir` parameter to our self-defined model? So that `my_model.h5` will be stored at `model_dir` as `modir_dir/my_model.h5`. Later on, if we need to save training logs and visualize ten, we can also save them at `model_dir`.

sql/codegen.go
```diff
@@ -238,39 +252,70 @@ for name in feature_column_names:
 Y = columns[field_names.index("{{.Y.Name}}")]
 {{- end}}
 
-classifier = tf.estimator.{{.Estimator}}(
-    feature_columns=feature_columns,{{range $key, $value := .Attrs}}
-    {{$key}} = {{$value}},{{end}}
-    model_dir = "{{.Save}}")
+classifier = {{.Estimator}}(
+	feature_columns=feature_columns,{{if .SelfDefined}}{{range $key, $value := .Attrs}}
+	{{$key}} = {{$value}},{{end}})
```
tonyyang-svail: Correct me if I am wrong, both self-defined and premade models need `{{range $key, $value := .Attrs}}{{$key}} = {{$value}},{{end}}`, is it possible to combine them?

sql/python/sqlflow/models/example.py
```diff
@@ -78,16 +78,16 @@ def prepare_prediction_column(self, prediction):
         return prediction.argmax(axis=-1)
 
 model = DNNClassifier(feature_columns=feature_columns, hidden_units=[10, 10], n_classes=3)
-model.compile(optimizer=model.default_optimizer(), loss=model.default_loss())
-
 is_training = False
 if is_training:
+    model.compile(optimizer=model.default_optimizer(), loss=model.default_loss())
     model.fit(train_ds, validation_data=val_ds, epochs=model.default_training_epochs(), verbose=0)
+    #model.save_weights('my_model', save_format="h5")
```
tonyyang-svail: Please clean up the commented code. :)

sql/python/sqlflow/models/example.py
```diff
@@ -78,16 +78,16 @@ def prepare_prediction_column(self, prediction):
...
     model.save_weights('my_model.h5')
     print("Done training.")
 else:
     model.predict(test_ds)
-    model.load_weights('my_model.h5')
+    model.load_weights('my_model.h5', by_name=True)
```
tonyyang-svail: What is the purpose of `by_name=True`? In the [official document](https://keras.io/models/about-keras-models/):

>  model.load_weights(filepath, by_name=False) loads the weights of the model from a HDF5 file (created by save_weights). By default, the architecture is expected to be unchanged. To load weights into a different architecture (with some layers in common), use by_name=True to load only those layers with the same name.

I am not sure if we are expecting to change the architecture.

tonyyang-svail (dismissed):
Excellent!

Generally LGTM. I have left some comments for your consideration.

scripts/test.sh
typhoonzero: It's useful to leave it here to ensure `sqlflow_models` is installed, or the following `go test` is not going to output a meaningful error message.

Adding pretty debug information when training script fails seems another work to do, I'll create an issue if needed.

Dockerfile
```diff
@@ -64,6 +64,12 @@ RUN mkdir -p /go/src/github.com/sql-machine-learning && \
     go install -v ./... && \
     cd /
 
+# Install latest sqlflow_models for testing custom models, see main_test.go:CaseTrainCustomModel
+RUN git clone https://github.com/sql-machine-learning/models.git && \
```
wangkuiyi, line 68: Should we make a git submodule in the sqlflow repo pointing to the models repo? -- just a reminder, no need to reply in this PR.

wangkuiyi (dismissed):

Dockerfile
typhoonzero: I could be wrong but submodule I thought should be a **required** dependency, but in this case, SQLFlow can still work if `sqlflow_models` is not installed.

wangkuiyi: Good point. I take back my suggestion.

sql/codegen.go
typhoonzero: Created an issue: https://github.com/sql-machine-learning/sqlflow/issues/432 to track this.

sql/codegen.go

sql/python/sqlflow/models/example.py
typhoonzero: Done

sql/python/sqlflow/models/example.py
typhoonzero: Removed

tonyyang-svail (approved):
LGTM!

PR #426, opened 2019-05-22
Adjust python directory structure of sqlflow and refactoring db.py

uuleon:
1. Adjust the python code directory structure
2. refactoring db.py


sql/python/sqlflow/template/db.py
```diff
@@ -0,0 +1,177 @@
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+import abc
+import six
+
+
+@six.add_metaclass(abc.ABCMeta)
```
tonyyang-svail: Why do we need meta programming?

sql/python/sqlflow/template/db.py
```diff
@@ -0,0 +1,177 @@
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+import abc
+import six
```
tonyyang-svail: SQLFlow only supports Python3. I don't think `__future__` is necessary. Python2 is retiring in 7 months anyway. :(

sql/python/sqlflow/template/db.py
```diff
@@ -0,0 +1,177 @@
...
+            values(list(tuple)): the values
+
+        Returns:
+            cursor(cursor)
+        """
+        cursor = self._db.cursor()
+        cursor.executemany(query, values)
+        return cursor
+
+    def create_table(self, table_name, table_schema):
```
tonyyang-svail: Which part of the codegen needs this method?

sql/python/sqlflow/template/db.py
```diff
@@ -0,0 +1,177 @@
...
+        self._password = password
+        self._host = host
+        self._port = port
+
+        self._db = self._connect()
+
+    def close(self):
+        self._db.close()
+
+    def execute(self, query):
```
tonyyang-svail: Python has DBAPI specification, which already standardize functions such as `execute`. Why adding another layer?

sql/python/sqlflow/template/db.py
```diff
@@ -0,0 +1,177 @@
...
+                       port=self._port)
+
+    def _retrieve_fields_and_value_from_cursor(self, cursor):
+        field_names = None if cursor.description is None \
+            else [i[0][i[0].find('.')+1:] for i in cursor.description]
+        field_columns = list(map(list, zip(*cursor.fetchall())))
+
+        return field_names, field_columns
+
+    def _gen_insert_table_statement(self, table_name, table_schema):
```
tonyyang-svail: prediction job needs this function.

sql/python/sqlflow/template/db.py
```diff
@@ -0,0 +1,177 @@
...
+
+    def _connect(self):
+        from mysql.connector import connect
+        return connect(user=self._user,
+                       passwd=self._password,
+                       database=self._database,
+                       host=self._host,
+                       port=self._port)
+
+    def _retrieve_fields_and_value_from_cursor(self, cursor):
```
tonyyang-svail: `_retrieve_fields_and_value_from_cursor` I found this function is exactly the same across `MySQLDataBase`, `SQLite3DataBase` and `HiveDataBase`...

sql/python/sqlflow/template/db.py
```diff
@@ -0,0 +1,177 @@
...
+            field_names(list(str)): the names of fields
+            field_data(list): return data of column-style
+        """
+        cursor = self.execute(statement)
+        return self._retrieve_fields_and_value_from_cursor(cursor)
+
+    def _connect(self):
+        raise NotImplementedError("DataBase.connect")
+
+    def _gen_create_table_statement(self, table_name, table_schema):
```
tonyyang-svail: Which part of the submitter needs to create a table?

tonyyang-svail (dismissed):
~~@uuleon I have no idea why things need to be implemented in such an overwhelming way, where~~
1. ~~Metaprogramming: `@six.add_metaclass(abs.ABSMeta)`~~
1. ~~OOP : `classSQLDataBase(DataBase)`~~
1. ~~ORM: `create_table`~~

~~are involved. Not to mention many methods are not implemented...~~

~~The goal of extracting python snippet from code template is to make the code more readable and testable. Please keep it simple... Readability matters...~~

~~I would suggest starting from simply taking out the code from the template like what @wangkuiyi did.~~

https://github.com/sql-machine-learning/sqlflow/blob/652aaaa4096f2bc2cb580ed7aa39ad81148c959d/sql/python/db.py#L1-L16

sql/python/sqlflow/template/db.py
uuleon: Should we create a table first before saving the prediction result?

tonyyang-svail: The current implementation does this in Go.

https://github.com/sql-machine-learning/sqlflow/blob/652aaaa4096f2bc2cb580ed7aa39ad81148c959d/sql/executor.go#L304-L306

sql/python/sqlflow/template/db.py
uuleon: No, `HiveDataBase ` is different

sql/python/sqlflow/template/db.py
uuleon: Ok

sql/python/sqlflow/template/db.py
uuleon: > @uuleon I have no idea why things need to be implemented in such an overwhelming way, where
> 
> 1. Metaprogramming: `@six.add_metaclass(abs.ABSMeta)`
> 2. OOP : `classSQLDataBase(DataBase)`
> 3. ORM: `create_table`
> 
> are involved. Not to mention many methods are not implemented...
> 
> The goal of extracting python snippet from code template is to make the code more readable and testable. Please keep it simple... Readability matters...
> 
> I would suggest starting from simply taking out the code from the template like what @wangkuiyi did.
> 
> https://github.com/sql-machine-learning/sqlflow/blob/652aaaa4096f2bc2cb580ed7aa39ad81148c959d/sql/python/db.py#L1-L16

With all due respect, I don' think it's an overwhelming design. The goal of designing `DataBase` is to hide all the specific details between databases, to make the code template of the `submitter` clean and simple.

Here are two differences between databases:
1. get `field_names` from `cursor`
Hive:  [i[0][i[0].find('.')+1:] for i in cursor.description]
Others: [i[0] for i in cursor.description]

2. template of the insert statement
MySQL: insert into table_name (features, labels) values (%s, %s)
SQLite3: insert into table_name (features, labels) values(?, ?)
Hive: insert into table table_name (featuers, labels) values (%s, %s)

That's why class `DataBase` is created and OOP is involved. If OOP is removed, some methods still have to exist for resolve the conflictions mentioned above and the code is not simple as you wish, or maybe leave it in the code template of the submitter? I don't think it's a good idea.

Should we have to use `create_table` method before saving prediction result? Correct me if I am wrong, please.

At last,  only one method which is `_gen_insert_table_statement` in `HiveDataBase` is not implemented, not many as you mentioned.

tonyyang-svail: > At last, only one method which is _gen_insert_table_statement in HiveDataBase is not implemented, not many as you mentioned.

@uuleon I apologize for the statement "Not to mention many methods are not implemented". It is inaccurate. I didn't realize most `NotImplementedError` lives in the base class.

sql/python/sqlflow/template/db.py
tonyyang-svail: My bad.

sql/python/sqlflow/template/db.py
tonyyang-svail: > Should we have to use `create_table` method before saving prediction result? Correct me if I am wrong, please.

In the current implementation, we don't. Table creation is done at Go, so the Python program doesn't need to create_table. A few month ago @wangkuiyi and I discussed keeping the database complication in Go and leaving the Python program small.

> Here are two differences between databases: ...

You are right. I feel (2) is relatively easy to solve in Go by using different statements for different SQL engines. However, we haven't figured out how to solve (1) in Go. @weiguoz has a detailed description of (1) at https://github.com/sql-machine-learning/sqlflow/issues/318.

sql/python/sqlflow/template/test_sqlite3_db.py
```diff
@@ -0,0 +1,32 @@
+from __future__ import absolute_import
```
tonyyang-svail: Would you mind enabling this test in our CI? This can be done at `scripts/test.sh`.

If you wanna test against Hive, please refer to modify `scripts/test_hive.sh`.

sql/python/sqlflow/template/db.py
```diff
@@ -0,0 +1,65 @@
+
```
wangkuiyi: Given that db.py is not a "template", but a "real" Python program that is runnable and testable, the directory name should be `sql/python/sqlflow` instead of `sql/python/sqlflow/template`.

sql/python/sqlflow/template/instance_db_test_.py
```diff
@@ -0,0 +1,48 @@
+import os
```
wangkuiyi: I am a little confused about why we need two test files for db.py.

Then I realized that is because .travis.yml called two shell scripts: scripts/test.sh and scripts/test_hive.sh.

However, I still feel that we can have only one test file, say, test_db.py, which contains

```python
class TestDB(TestCase):
   def test_mysql(self):
      ...
   def test_sqlite(self):
      ...
   def test_hivel(self):
      ...
```

where in each method, say, `test_mysql`, it judges the value of environment variable `SQLFLOW_TEST_DB`:

```python
def test_mysql(self):
    if os.environ['SQLFLOW_TEST_DB'] == "" OR os.environ['SQLFLOW_TEST_DB'] == "mysql:
        go_on_test
```

sql/python/sqlflow/template/db.py
uuleon: ok

wangkuiyi (dismissed):

sql/python/sqlflow/db.py
```diff
@@ -0,0 +1,65 @@
...
+        statement = '''insert into table {} ({}) values()'''.format(
+            table_name,
+            ", ".join([n for n, t in table_schema]),
+            ", ".join(["%s"] * len(table_schema))
+        )
+    else:
+        raise ValueError("unrecognized database driver: %s" % driver)
+
+    cursor = conn.cursor()
+    cursor.executemany(statement, values)
```
tonyyang-svail, line 68:
```
test_hive (sqlflow.db_test.TestDB) ... ok
test_mysql (sqlflow.db_test.TestDB) ... ERROR
test_sqlite3 (sqlflow.db_test.TestDB) ... ok
======================================================================
ERROR: test_mysql (sqlflow.db_test.TestDB)
----------------------------------------------------------------------
  File "/go/src/github.com/sql-machine-learning/sqlflow/sql/python/sqlflow/db_test.py", line 29, in test_mysql
    self._do_test(driver, conn)
  File "/go/src/github.com/sql-machine-learning/sqlflow/sql/python/sqlflow/db_test.py", line 44, in _do_test
    execute(driver, conn, self.drop_statement)
  File "/go/src/github.com/sql-machine-learning/sqlflow/sql/python/sqlflow/db.py", line 36, in execute
    field_columns = list(map(list, zip(*cursor.fetchall())))
  File "/miniconda/envs/sqlflow-dev/lib/python3.6/site-packages/mysql/connector/cursor_cext.py", line 489, in fetchall
    raise errors.InterfaceError("No result set to fetch from.")
mysql.connector.errors.InterfaceError: No result set to fetch from.
```

@uuleon 
CI failed on MySQL with `No result set to fetch from.`. Maybe add `coursor.commit()` after `executemany()`?
Reference: https://stackoverflow.com/questions/384228/database-does-not-update-automatically-with-mysql-and-python

uuleon: I don't think so. From the error stack information, the result of `fetchall` is indeed None because of the query is `drop table ***`, I have added code to handle this situation.

tonyyang-svail: Great!

uuleon: The actual reason for this failure is the syntax of `fetchall` is different for `MySQL` and `Sqlite3` when there are no results from `execute`, `None` is returned from `Sqlite3` and `InterfaceError` is raised by `MySQL`.

tonyyang-svail (approved):
Excellent job! With these Python functions, the codegen template will be much simpler! :)

wangkuiyi (approved):
LGTM++

PR #429, opened 2019-05-23
Fix drivername parse

weiguoz:
`maxcompute` use `url.Parse(dsn)` to parse a data source, so the `dsn` must begin with a scheme, which means sqlflow should accept the input like: "maxcompute://http://u:p@host/api?curr_project=xx"

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@
 /logs
 sql/my_dnn_model
 .DS_Store
+coverage.out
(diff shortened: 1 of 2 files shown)


tonyyang-svail (approved):
I feel double `://`s in `maxcompute://http://u:p@host/api?curr_project=xx` seems weird. I am approving this PR since it is a major blocker on supporting MaxCompute. I will investigate in defining the source name in another way such as `maxcompute://u:p@host/api?curr_project=xx&scheme=http`

PR #446, opened 2019-05-24
unify test data preparation

### LINKED ISSUE #436: Unify data population strategy for testing
Currently, there exist two ways to populate the testing data.

1. Script, which uses `example/datasets/*.sql`

https://github.com/sql-machine-learning/sqlflow/blob/c287208147baff4b6f40db5e1e052613c90d01c7/scripts/test.sh#L6-L9

1. `TestMain` in `some_test.go`, which uses `sql/testdata/*.sql`.

https://github.com/sql-machine-learning/sqlflow/blob/c287208147baff4b6f40db5e1e052613c90d01c7/sql/db_test.go#L17-L61

I prefer the latter because `TestMain` is guaranteed to run on every `go test`, including `go test -v` and `go test -run TestFunc`. So that a developer can rerun the same tests several times without worrying about the effect of the previous run. Because  `sql/testdata/*.sql` includes `drop table` statements that do the cleaning up.

I would suggest combine `example/datasets/*.sql` and `sql/testdata/*.sql` and put them at `/sqlflow-test-initdb.d/*.sql`. The advantage has two folds:
1. `/sqlflow-test-initdb.d/` can be accessed by both `sql/*_test.go` and `cmd/*_test.go`.
1. `/sqlflow-test-initdb.d/` only exists in the Docker image, this enforces developers to do test inside the Docker, as suggested in [build.md](https://github.com/sql-machine-learning/sqlflow/blob/develop/doc/build.md#build-and-test).

Note: during quickstart, a first-time user also needs some pre-existing data to play with. Data population is down at `scripts/startall.sh`

typhoonzero:
Fix https://github.com/sql-machine-learning/sqlflow/issues/436


sql/testdata/toutiao.sql
```diff
@@ -0,0 +1,110 @@
+DROP TABLE IF EXISTS toutiao.train_processed;
```
wangkuiyi: I see this file is named toutiao. Is this from ByteDance Inc.?

sql/popularize.go
```diff
@@ -0,0 +1,49 @@
...
+		for i := 0; i < len(data); i++ {
+			if data[i] == ';' {
+				return i + 1, data[:i], nil
+			}
+		}
+		return 0, nil, nil
+	}
+
+	scanner := bufio.NewScanner(f)
+	scanner.Split(onSemicolon)
```
wangkuiyi: How if there is a statement in the file `sqlfile` which contains `;` in a string constant? For example,

```sql
INSERT INTO toutiao.train_processed VALUES ('news;music;video')
```

there are `;` in the inserted value.

sql/testdata/toutiao.sql
typhoonzero: It's from an open source dataset: https://github.com/fate233/toutiao-text-classfication-dataset

sql/popularize.go
typhoonzero: This function does not consider this case, I just move it out from the original file. I'll add some comments here because we only need to process some sample data files, if we want to put it into general use, we should probably update this function.

wangkuiyi: If the only use of the function `Popularize` is in `cmd/sqlflowserver/`, it seems that we should move it there. The directory `sql` doesn't seem the right place for a hacky test-only function.

sql/testdata/toutiao.sql
wangkuiyi: How about we rename `toutiao` into `text`? It would be even better if we use English text dataset.

typhoonzero: Sure. Will add English dataset in next PR.

sql/popularize.go
```diff
@@ -0,0 +1,50 @@
+package sql
```
wangkuiyi: I see how comes the content of this file. I would recommend to move it from `sql/popularize.go` into `sql/testdata/popularize.go`.

cmd/sqlflowserver/main_test.go
```diff
@@ -114,6 +115,40 @@ func ParseRow(stream pb.SQLFlow_RunClient) ([]string, [][]*any.Any) {
 	return columns, rows
 }
 
+func prepareTestData(dbStr string) error {
+	// popularize test data
+	testDB, err := sql.Open(dbStr)
+	if err != nil {
+		return err
+	}
+	_, err = testDB.Exec("CREATE DATABASE IF NOT EXISTS iris;")
```
wangkuiyi: Could we move `CREATE DATABASE IF NOT EXISTS iris` in `sql/testdata/iris.sql`, other than in this .go file?

wangkuiyi (commented):
Also, I would suggest rename `sql/testdata/toutiao.sql` int `sql/testdata/text.sql`.

cmd/sqlflowserver/main_test.go
```diff
@@ -123,8 +158,14 @@ func TestEnd2EndMySQL(t *testing.T) {
 	if testDBDriver != "mysql" {
 		t.Skip("Skipping mysql tests")
 	}
-	go start("mysql://root:root@tcp/?maxAllowedPacket=0")
+	dbStr := "mysql://root:root@tcp/?maxAllowedPacket=0"
+	go start(dbStr)
 	WaitPortReady("localhost"+port, 0)
```
hungry1526, line 152: Not related to this PR...where is this variable defined in this go file?

typhoonzero: `port` is defined here: https://github.com/sql-machine-learning/sqlflow/blob/develop/cmd/sqlflowserver/main.go#L33

sql/testdata/churn_sql.go
```diff
@@ -140,4 +157,4 @@ INSERT INTO churn.test VALUES("1862-QRWPE","Female","0","Yes","Yes","48","Yes","
 INSERT INTO churn.test VALUES("2796-NNUFI","Female","0","Yes","Yes","46","Yes","No","No","No internet service","No internet service","No internet service","No internet service","No internet service","No internet service","Two year","Yes","Mailed check","19.95","927.1","No");
 INSERT INTO churn.test VALUES("3016-KSVCP","Male","0","Yes","No","29","No","No phone service","DSL","No","No","No","No","Yes","No","Month-to-month","No","Mailed check","33.75","1009.25","No");
 INSERT INTO churn.test VALUES("4767-HZZHQ","Male","0","Yes","Yes","30","Yes","No","Fiber optic","No","Yes","Yes","No","No","No","Month-to-month","No","Bank transfer (automatic)","82.05","2570.2","No");
-INSERT INTO churn.test VALUES("2424-WVHPL","Male","1","No","No","1","Yes","No","Fiber optic","No","No","No","Yes","No","No","Month-to-month","No","Electronic check","74.7","74.7","No");
+INSERT INTO churn.test VALUES("2424-WVHPL","Male","1","No","No","1","Yes","No","Fiber optic","No","No","No","Yes","No","No","Month-to-month","No","Electronic check","74.7","74.7","No");`
```
tonyyang-svail, line 160: ~~typo?~~

sql/popularize.go
```diff
@@ -0,0 +1,36 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import (
+	"strings"
+)
+
+// Popularize reads SQL statements from the file named *.sql
+// and runs each SQL statement with db.
+func Popularize(db *DB, sqlcontent string) error {
```
typhoonzero: Put `Popularize` in package `sql` but not in `testdata` so that would not generate cycle import

sql/testdata/churn_sql.go
```diff
@@ -1,3 +1,20 @@
+package testdata
+
+// Copyright 2019 The SQLFlow Authors. All rights reserved.
```
tonyyang-svail, line 1: Copyright should come before `package testdata`. Is this a bug in `scripts/copyright.py`?

sql/testdata/text_cn_sql.go
```diff
@@ -0,0 +1,127 @@
+package testdata
+
+// Copyright 2019 The SQLFlow Authors. All rights reserved.
```
tonyyang-svail, line 1: Copyright should come before `package testdata`. Is this a bug in `scripts/copyright.py`?

Dockerfile
```diff
@@ -90,7 +90,6 @@ ENV PATH=/miniconda/envs/sqlflow-dev/bin:$PATH
 # Prepare sample datasets
 COPY example/datasets/popularize_churn.sql /docker-entrypoint-initdb.d/popularize_churn.sql
 COPY example/datasets/popularize_iris.sql /docker-entrypoint-initdb.d/popularize_iris.sql
-COPY example/datasets/create_model_db.sql /docker-entrypoint-initdb.d/create_model_db.sql
```
tonyyang-svail: This `.sql` is needed for SQLFlow quickstart. We need to `CREATE DATABASE IF NOT EXISTS sqlflow_models;` in advance so that `sqlfs` can save a model into `sqlflow_models.my_dnn_model`.

example/datasets/create_model_db.sql
```diff
@@ -1 +0,0 @@
-CREATE DATABASE IF NOT EXISTS sqlflow_models;
```
tonyyang-svail: This `.sql` is needed for SQLFlow quickstart.

sql/popularize.go
tonyyang-svail: You can use `database/sql` to avoid circular import. In `package testdata`
```go
package testdata

import (
	"strings"
	"database/sql"    // not github.com/sql-machine-learning/sqlflow/sql
)

func Popularize(db *sql.DB, sqlcontent string) error { ... }
```

Then in other packages,

```go
package main

import (
	"github.com/sql-machine-learning/sqlflow/sql"
	"github.com/sql-machine-learning/sqlflow/sql/testdata"
)

func main() {
	sqlflowDB := sql.Open(...)
	testdata.Popularize(sqlflowDB.DB, testdata.IrisSQL) // notice sqlflowDB.DB
}
```

and

```go
package sql

import (
	"github.com/sql-machine-learning/sqlflow/sql/testdata"
)

func main() {
	sqlflowDB := Open(...)
	testdata.Popularize(sqlflowDB.DB, testdata.IrisSQL) // notice sqlflowDB.DB
}
```

sql/popularize.go
typhoonzero: Comments done, thanks for advices.

sql/popularize.go
typhoonzero: Done.

sql/testdata/churn_sql.go
typhoonzero: Seems so. Will do more tests and try to fix this this week.

Dockerfile
typhoonzero: Added back, maybe unify these later in another PR.

example/datasets/create_model_db.sql
typhoonzero: Done.

tonyyang-svail (approved):
LGTM! Excellent!

PR #448, opened 2019-05-24
Codegen refactor

uuleon:
1. use the python module of `sqlflow` in the `codegen` template.
2. add the error message for executor_test.go if a test failed.


wangkuiyi (approved):
LGTM!

sql/codegen.go
```diff
@@ -213,32 +207,15 @@ STEP = 1000
...
-db = connect({{.Database}})
-{{else if eq .Driver "hive"}}
-db = connect(user="{{.User}}",
-            password="{{.Password}}",
-            {{if ne .Database ""}}database="{{.Database}}",{{end}}
-            host="{{.Host}}",
-            port={{.Port}})
+driver="{{.Driver}}"
+{{if ne .Database ""}}
+database="{{.Database}}"
```
wangkuiyi, line 212: Can we move this if-else logic into db.py? Something like

```python
def connect(..., database, ...):
    if database == "":
        database = None
    ...
```

sql/codegen.go
```diff
@@ -213,32 +207,15 @@ STEP = 1000
...
-cursor = db.cursor()
-cursor.execute("""{{.StandardSelect}}""")
-{{if eq .Driver "hive"}}
-field_names = [i[0][i[0].find('.')+1:] for i in cursor.description]
-{{else}}
-field_names = [i[0] for i in cursor.description]
-{{end}}
-columns = list(map(list, zip(*cursor.fetchall())))
+conn = connect(driver, database, user="{{.User}}", password="{{.Password}}", host="{{.Host}}", port={{.Port}})
+field_names, columns = execute(driver, conn, """{{.StandardSelect}}""")
```
wangkuiyi, line 219: How about in the future PRs, let us do something more:

- Move the following code snippet that constructs `feature_columns` and `feature_column_name` into a Python file, say, `naive_feature_dreivation.py`, and make it a function `derive_features`?

- Move the code snippet that constructs the classifier into `create_estimator.py` and make it a function `create_estimator`?

- Move the function `train_input_fn` and `eval_input_fn` and the logic that calls them to train and to evaluate the model into `train.py`?

wangkuiyi (commented):
Added some comments in hope to extremely simplify the Python template in Go source files. Not for doing in this PR, but should be done in future PRs.

sql/codegen.go
uuleon: Ok

sql/codegen.go
uuleon: Yes, I will do this in the future PRs after both considering the code in `codegen.go` and `codegen_alps.go`.

PR #449, opened 2019-05-26
Add calcite parser Go client and update the Java server

wangkuiyi:
This PR implements a building block of the [`external_parser`](https://github.com/sql-machine-learning/sqlflow/blob/ab1307d93ef3973ba9853924e0f094c9035f59fb/parser/README.md#the-external-parser-abstraction) in the design of integrated parser.


parser/calcite/README.md
```diff
@@ -1,38 +1,33 @@
...
- docker run --rm -it -v $PWD:/work -w /work calcite:dev protoc --grpc-java_out=. CalciteParser.proto
- ```
+## Build and Run
 
-Build and generate `.class` files:
+The following command builds `CalciteParserServer.class` and runs it:
 
 ```bash
-docker run --rm -it -v $PWD:/work -w /work calcite:dev javac *.java
+docker run --rm -d -p 50051:50051 -v $PWD:/work -w /work calcite:dev bash ./build_and_run.bash
```
weiguoz: SQLFlow server exposes [port 50051](https://github.com/sql-machine-learning/sqlflow/blob/develop/doc/installation.md#use-your-own-database) already. I would suggest taking another port.

wangkuiyi: Done

weiguoz (approved):
Excellent PR!

parser/calcite/CalciteParserServer.java
```diff
@@ -81,24 +94,42 @@ public void parse(
...
+          .build());
       responseObserver.onCompleted();
     }
 
-    private static int PosToIndex(String query, SqlParserPos pos) {
+    // calciteParse returns <len(query),null> if Calcite parser
+    // accepts query, or returns <pos,null> if a second parsing
+    // accepts the content to the left of the error position from the
+    // first parsing, or <-1,e> if both parsing failed.
+    private static Pair<Integer, SqlParseException> calciteParse(String query) {
```
uuleon, line 112: Normally, in `Java` an `exception` is thrown out instead of returning like `go`. In that way, the `pair` in `JavaFX` is also unnecessary.  What do you think?

parser/calcite/CalciteParserServer.java
```diff
@@ -81,24 +94,42 @@ public void parse(
...
+      try {
+        SqlParser parser = SqlParser.create(query);
+        SqlNode sqlNode = parser.parseQuery();
```
uuleon, line 115: Is the `sqlNode` necessary? Why not remove it?

`SqlParser.create(query).parseQuery();`

parser/calcite/CalciteParserServer.java
```diff
@@ -81,24 +94,42 @@ public void parse(
...
+      } catch (SqlParseException e) {
+        int epos = posToIndex(query, e.getPos());
+        try {
+          SqlParser parser = SqlParser.create(query.substring(0, epos));
+          SqlNode sqlNode = parser.parseQuery();
+        } catch (SqlParseException ee) {
+          return new Pair<Integer, SqlParseException>(epos, ee);
+        }
+        return new Pair<Integer, SqlParseException>(epos, null);
+      }
```
uuleon: If the `FROM` block is written at the end of the query, is the result of `calciteParse` correct?

`
SELECT *
TRAIN DNNClassifier
...
FROM table
`

parser/calcite/CalciteParserServer.java
wangkuiyi: Will fix it in the next PR.

parser/calcite/CalciteParserServer.java
wangkuiyi: Do you mean returning multiple values by throwing an exception? I am afraid that the exception is not designed for returning values?

PR #450, opened 2019-05-26
Make tidb.Parser calls TiDB parser twice as Calcite paser does

wangkuiyi:
This PR implements a building block of the [`external_parser`](https://github.com/sql-machine-learning/sqlflow/blob/ab1307d93ef3973ba9853924e0f094c9035f59fb/parser/README.md#the-external-parser-abstraction) in the design of integrated parser.

### DIFF
--- a/parser/tidb/tidb_parser.go
+++ b/parser/tidb/tidb_parser.go
@@ -15,29 +15,54 @@ package tidb
 
 import (
 	"fmt"
+	"log"
 	"regexp"
 	"strings"
+	"sync"
 
 	"github.com/pingcap/parser"
 	_ "github.com/pingcap/tidb/types/parser_driver" // As required by https://github.com/pingcap/parser/blob/master/parser_example_test.go#L19
 )
+
+var (
+	// Use the global variable to save the time of creating parser.
+	psr *parser.Parser
+	re  *regexp.Regexp
+	mu  sync.Mutex
+)
+
+// Init creates the TiDB parser instance and other resources.
+func Init() {
+	mu.Lock()
+	defer mu.Unlock()
+
+	psr = parser.New()
+	re = regexp.MustCompile(`.* near "([^"]+)".*`)
+}
 
 // Parse calls TiDB's parser to parse a statement sql.  It returns
 // <-1,nil> if TiDB parser accepts the statement, or <pos,nil> if TiDB
 // doesn't accept but returns a `near "..."` in the error message, or
 // <-1,err> if the error messages doens't contain near.
 func Parse(sql string) (idx int, err error) {
-	p := parser.New()
-	_, _, err = p.Parse(sql, "", "")
+	if psr == nil || re == nil {
+		log.Fatalf("Parser must be called after Init")
+	}
 
-	if err != nil {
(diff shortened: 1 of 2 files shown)


parser/tidb/tidb_parser.go
```diff
@@ -15,29 +15,54 @@ package tidb
...
+	mu.Lock()
+	defer mu.Unlock()
 
+	if _, _, err = psr.Parse(sql, "", ""); err != nil {
+		matched := re.FindAllStringSubmatch(err.Error(), -1)
 		if len(matched) != 1 || len(matched[0]) != 2 {
 			return -1, fmt.Errorf("Cannot match near in %q", err)
 		}
-		return strings.Index(sql, matched[0][1]), nil
+		idx = strings.Index(sql, matched[0][1])
```
typhoonzero, line 60: Maybe can use `FindAllStringIndex` to get the matched index directly?

wangkuiyi: Good point!

weiguoz (approved):
LGTM

parser/tidb/tidb_parser.go
wangkuiyi: Actually, we cannot use FindAllStringIndex here. The goal is to parse the substring after the keyword "near" in the error string, so could we find the index of the substring in the SQL statement.

typhoonzero: Indeed. Yet I found another problem, and maybe you have already noticed before: the snippet below shows that, SQL like `SELECT * FROM iris.mytable2 TRAIN..` can not get a proper index of substatement before`TRAIN`.

```go
package main

import (
	"fmt"
	"regexp"
	"strings"

	"github.com/pingcap/parser"
	_ "github.com/pingcap/tidb/types/parser_driver" // As required by https://github.com/pingcap/parser/blob/master/parser_example_test.go#L19
)

func testParse(psr *parser.Parser, sql string) {
	_, _, e := psr.Parse(sql, "", "")
	fmt.Printf("error msg: %s\n", e.Error())
	re := regexp.MustCompile(`.* near "([^"]+)".*`)
	matched := re.FindAllStringSubmatch(e.Error(), -1)
	idx := strings.Index(sql, matched[0][1])
	fmt.Printf("Index: %d, sql before: %s\n", idx, sql[:idx])
	fmt.Println("---------------------------")
}

func main() {
	psr := parser.New()
	testParse(psr, "SELECT * FROM iris.mytable2 TRAIN DNNClassifier COLUMN c1,c2,c3 LABLE class INTO mymodels.dnn1")
	testParse(psr, "SELECT * FROM iris.mytable2 LIMIT 100 TRAIN DNNClassifier COLUMN c1,c2,c3 LABLE class INTO mymodels.dnn1")
}

```

wangkuiyi: Yes, I noticed this, as stated in 

https://github.com/sql-machine-learning/sqlflow/blob/1424350537aaf42ebf86fba10d1a61469937be5a/parser/tidb/tidb_parser_test.go#L59-L65

Still, I don't have a solution in my mind. But anyway, thanks for reminding me about this issue. Let us find a solution to it.

PR #451, opened 2019-05-26
Add parser/README.md

wangkuiyi:
This is the design doc of the integrated parsing.

### DIFF
--- a/parser/README.md
+++ b/parser/README.md
@@ -52,7 +52,7 @@ func Parse(sql string) (acceptable bool, err error) {
     // The left part is a SELECT and the right part is TRAIN or PREDICT.
     return false, nil 
 }
-`"
+```
 
 This procedure has two assumptions: (1) a SQL engine has a parser, denoted by sql_engine.Parse, and (2) the parser reports the position of syntax errors if there is any.  Both assumptions are reasonable. However, a key question here is how to call parsers of various SQL engines. For open source SQL engines, this doesn't seem an issue. Here we list parsers used by some well-known engines:
 
--- a/parser/README.md
+++ b/parser/README.md
@@ -127,7 +127,7 @@ The parser verifies the syntax of the input.  SQLFlow also needs to check the lo
 
 ```sql
 SELECT a, b FROM t1 TRAIN DNNClassifier COLUMN c LABEL b;
-`"
+```
 
 There is no syntax error in the above example; however, it has a logical mistake -- the column `c` is undefined.
 
--- a/parser/README.md
+++ b/parser/README.md
@@ -1,0 +1,134 @@
+# Integrating with Parsers of SQL Engines
+
+## The Challenge
+
+Typical SQLFlow statements are like the following.
+
+```sql
+SELECT * FROM training_data TRAIN DNNClassifier LABEL kind INTO my_model;
+
+SELECT * FROM testing_data PREDICT testing_data.predicted_kind USING my_model;
+```
+
+The point is, assuming someone already has a SELECT statement for data cleaning and augmentation, (s)he could add a TRAIN or PREDICT clause to enable AI, no matter how complex the statement is or if it is hundreds of lines of code including nested SELECT.
+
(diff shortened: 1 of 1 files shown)


parser/README.md
```diff
@@ -0,0 +1,134 @@
...
+
+Some external parsers are in Go and don't need RPC.  For example, TiDB parser is in Go, and HiveQL parser is a set of AntLR grammar rules, and AntLR can generate Go code.  However, we would unify the interface to external parsers:
+
+```go
+func external_parser(kind, sql string) (idx int, err error) {
+   switch(kind) {
+   case 'calcite':
+       r := grpc.Call("CalciteParser.Parse", CalciteParserRequest{sql})
+       return r.GetIndex(), r.GetError()
+   case 'mysql':
```
zou000, line 98: should be 'tidb'?

wangkuiyi: It is MySQL. We used TiDB parser for MySQL.

parser/README.md
```diff
@@ -0,0 +1,134 @@
...
+    rightPart = sql[pos:]
+
+    errLeft := sql_engine.Parse(leftPart)
+    // In this case, the SQL is not acceptable due to the syntax error
+    if err != nil {
+       return false, err 
+    }
+
+    // If leftPart is acceptable, it is a legitimate  SELECT statement. We then try right part with SQLFlow parser. 
+    errRight := SQLFlow.Parse(rightPart)
```
weiguoz, line 47: ~~I would suggest putting the right part before the left part to reduce several network latency. 
Because of the `SQLFlow.Parse` run **locally** but `sql_engine.Parse` not.~~

parser/README.md
```diff
@@ -0,0 +1,134 @@
...
+| Storm      | Calcite |
+| Hive       | Hplsql.g4 |
+| MySQL      | sql_parse.{h,cc} |
+
+It is notable that MySQL doesn't provide sufficient documentation on how to call its parser, so we call a Go implementation of MySQL's SQL parser by PingCap https://github.com/pingcap/parser instead. For proprietary engines, we have a plan to ask the licensing of their parsers.
+
+
+## The External Parser Abstraction
+
+The above example program calls `sql_engine.Parse` twice before calling `SQLFlow.Parse`.  In practice, because Calcite parser is in Java, SQLFlow is a Go program, to enable SQLFlow calling Calcite parser, we have to wrap Calcite parser up into a gRPC server.  It is time-consuming to make an RPC call, so we pack the two calls to `sql_engine.Parse` into one:
```
weiguoz, line 72: Agree with `packing the two calls into one` to reduce time-consuming.
But I would suggest keeping the parses as simple as possible:
1. for a correct extended SQL: latency is mainly caused by the execution phase
2. I'm not sure if `SQLFow` would expand keywords like `yet_another_train`. If yes: we should update `SQLFlow` & `external parsers`, which means more complicate for deploy.

parser/README.md
```diff
@@ -0,0 +1,134 @@
...
+
+The parser verifies the syntax of the input.  SQLFlow also needs to check the logic.  For example, have a look at the following SQLFlow statement
+
+```sql
+SELECT a, b FROM t1 TRAIN DNNClassifier COLUMN c LABEL b;
+```
+
+There is no syntax error in the above example; however, it has a logical mistake -- the column `c` is undefined.
+
+In the design of SQLFlow, we do verification after parsing.  The verifier would like to know the field and table named mentioned in the statement, which is supposed to be returned by `external_parser` in addition to `idx`. Both TiDB parser and Calcite parser return an abstract syntax tree; it seems that we can traverse the tree and find out the field and table names.
```
typhoonzero, line 134: Just curious, in order to retrieve the `abstract syntax tree`, the RPC reply like `CalciteParserReply` should return a tree structure that can be serialized by protobuf, maybe can be a [protobuf map](https://developers.google.com/protocol-buffers/docs/proto3#maps)

typhoonzero (approved):
LGTM++

parser/README.md
wangkuiyi: We might make the RPC server traverse the AST and return two lists -- fields and tables.

typhoonzero: I see, Thanks!

PR #458, opened 2019-05-27
support configure some hyper params

typhoonzero:
Support some hyper parameters, so can train set below params when training like:

```sql
SELECT *
FROM iris.train
TRAIN DNNClassifier
WITH n_classes = 3, hidden_units = [10, 20], BATCH_SIZE = 128, STEPS = 100, EPOCHS=30
LABEL class
```

### DIFF
--- a/cmd/sqlflowserver/main_test.go
+++ b/cmd/sqlflowserver/main_test.go
@@ -357,7 +357,6 @@ FROM iris.predict LIMIT 5;`
 // CaseTrainTextClassification is a simple End-to-End testing for case training
 // text classification models.
 func CaseTrainTextClassification(t *testing.T) {
-	t.Skip()
 	a := assert.New(t)
 	trainSQL := `SELECT *
 FROM toutiao.train_processed
@@ -388,7 +387,7 @@ func CaseTrainSQLWithHyperParams(t *testing.T) {
 	trainSQL := `SELECT *
 FROM iris.train
 TRAIN DNNClassifier
-WITH n_classes = 3, hidden_units = [10, 20], BATCH_SIZE = 10, EPOCHS = 2
+WITH n_classes = 3, hidden_units = [10, 20], BATCHSIZE = 10, EPOCHS = 2
 COLUMN sepal_length, sepal_width, petal_length, petal_width
 LABEL class
 INTO sqlflow_models.my_dnn_model;`
--- a/cmd/sqlflowserver/main_test.go
+++ b/cmd/sqlflowserver/main_test.go
@@ -130,6 +130,8 @@ func TestEnd2EndMySQL(t *testing.T) {
 	t.Run("TestTrainSQL", CaseTrainSQL)
 	t.Run("TestTextClassification", CaseTrainTextClassification)
 	t.Run("CaseTrainCustomModel", CaseTrainCustomModel)
+	t.Run("CaseTrainSQLWithHyperParams", CaseTrainSQLWithHyperParams)
+	t.Run("CaseTrainCustomModelWithHyperParams", CaseTrainCustomModelWithHyperParams)
 }
 
 func TestEnd2EndHive(t *testing.T) {
@@ -355,6 +357,7 @@ FROM iris.predict LIMIT 5;`
 // CaseTrainTextClassification is a simple End-to-End testing for case training
 // text classification models.
 func CaseTrainTextClassification(t *testing.T) {
+	t.Skip()
 	a := assert.New(t)
 	trainSQL := `SELECT *
 FROM toutiao.train_processed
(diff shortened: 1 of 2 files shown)


cmd/sqlflowserver/main_test.go
```diff
@@ -355,6 +357,7 @@ FROM iris.predict LIMIT 5;`
 // CaseTrainTextClassification is a simple End-to-End testing for case training
 // text classification models.
 func CaseTrainTextClassification(t *testing.T) {
+	t.Skip()
```
tonyyang-svail: What is the purpose of this skip? Is `CaseTrainTextClassification` time consuming?

sql/codegen.go
```diff
@@ -200,13 +200,32 @@ import logging
...
+STEPS = 1000
+EPOCHS = 1
 NUM_BUCKETS=160000
 EMBEDDING_WIDTH=128
 
+train_args = dict()
+{{range $key, $value := .Attrs}}
+{{if eq $key "BATCHSIZE"}}
+BATCHSIZE = {{$value}}
+{{else if eq $key "BATCH_SIZE"}}
```
tonyyang-svail: Not sure we need to support both `BATCHSIZE` and `BATCH_SIZE`..

sql/codegen.go
```diff
@@ -290,7 +309,7 @@ eval_result = classifier.evaluate(eval_input_fn(X, Y, BATCHSIZE), verbose=0)
 print("Training set accuracy: {accuracy:0.5f}".format(**{"accuracy": eval_result[1]}))
 {{else}}
 eval_result = classifier.evaluate(
-	input_fn=lambda:eval_input_fn(X, Y, BATCHSIZE), steps=STEP)
+	input_fn=lambda:eval_input_fn(X, Y, BATCHSIZE), steps=STEPS)
```
tonyyang-svail, line 308: What if `STEPS` is larger than the eval dataset batch numbers?

I don't think `steps` should be configurable. A `PREDICT` should always go through the whole selected data. If a user wants a subset of the select, (s)he can put the restriction on the StandardSelect
```sql
SELECT ...
WHERE ...
LIMIT ...
PREDICT...
```

sql/codegen.go
```diff
@@ -200,13 +200,32 @@ import logging
...
 EMBEDDING_WIDTH=128
 
+train_args = dict()
+{{range $key, $value := .Attrs}}
+{{if eq $key "BATCHSIZE"}}
+BATCHSIZE = {{$value}}
+{{else if eq $key "BATCH_SIZE"}}
+BATCHSIZE = {{$value}}
+{{else if eq $key "EPOCHS"}}
+EPOCHES = {{$value}}
```
tonyyang-svail: `EPOCHES` or `EPOCHS`?

sql/codegen.go
```diff
@@ -271,13 +290,13 @@ classifier.compile(optimizer=classifier.default_optimizer(),
 	loss=classifier.default_loss(),
 	metrics=["accuracy"])
 classifier.fit(train_input_fn(X, Y, BATCHSIZE),
-	epochs=classifier.default_training_epochs(),
-	steps_per_epoch=STEP, verbose=0)
+	epochs=EPOCHS if EPOCHS != classifier.default_training_epochs() else classifier.default_training_epochs(),
```
tonyyang-svail: In the current configuration, I think `classifier.default_training_epochs()` will never be used...  Given `EPOCHS` has been initialized to a number, this line is equivalent to
```python
epochs = EPOCHS
```

How about we rewrite line 208 to `EPOCHS = None` then rewrite here as
```python
epochs=EPOCHS if EPOCHS else classifier.default_training_epochs()
```
So the logic becomes
```
if user_defined:
    epochs = user_defined_epochs
else:
    epochs = model_default_epochs
```

sql/codegen.go
typhoonzero: Determine the actual dataset size ( total number of elements ) sometimes is time-consuming especially on very large datasets ( like dataset stored in Hive ), if user didn't provide `LIMIT` clause then we must execute `SELECT count(*)  FROM table WHERE ...` then calculate `STEPS`.

I'm not sure whether we should do this now, so put `STEPS` configurable for now and move on to an automatic solution.

cmd/sqlflowserver/main_test.go
typhoonzero: Just for test, removed after clean.

sql/codegen.go

sql/codegen.go

sql/codegen.go
typhoonzero: Thanks, that's clever. Done!

sql/codegen.go
tonyyang-svail: We don't need to determine the actual dataset size. If we don't specify `steps`, `classifier.evaluate` will consume all the data points returned by`eval_input_fn`.

> steps: Integer or None. Total number of steps (batches of samples) before declaring the evaluation round finished. Ignored with the default value of None.

tonyyang-svail (approved):

sql/codegen.go
typhoonzero: Checked keras's document: https://keras.io/models/model/#fit `model.fit` can accept `steps_per_epoch=None` also. Removing this in the next PR while I'll refine the data reading part

PR #461, opened 2019-05-28
Use dataset from generator to deal with large datasets

typhoonzero:
Update reader part in `codegen.go`, use `tf.data.Dataset.from_generator` so that we don't use `fetchall()` to get all data into memory when dataset is large.


sql/python/sqlflow_submitter/db.py
```diff
@@ -55,6 +56,64 @@ def execute(driver, conn, statement):
...
+        label_idx = field_names.index(label_column_name)
+
+        rows = cursor.fetchmany(fetch_size)
+        while len(rows) > 0:
+            for row in rows:
+                label = row[label_idx]
+                features = dict()
+                for name in feature_column_names:
+                    if column_name_to_type[name] == "categorical_column_with_identity":
+                        cell = np.array([int(v) for v in row[field_names.index(name)].split(",")])
```
uuleon, line 80: Maybe can we use `np.fromstring` instead to make it efficiency?

sql/python/sqlflow_submitter/db.py
```diff
@@ -55,6 +56,64 @@ def execute(driver, conn, statement):
     return field_names, field_columns
 
 
+def db_generator(driver, conn, statement,
+                 feature_column_names, label_column_name,
+                 column_name_to_type, fetch_size=128):
+    def reader():
```
uuleon, line 62: It seems like this `reader` function and `execute` function looks similar, maybe we can combine them into a single function which can be controlled by argument to return all data or iterator.

uuleon: If the situation when `label_column_name` is null can be handled correctly, can we remove `db_generator_predict `?

sql/codegen.go
```diff
@@ -193,18 +193,16 @@ try:
 except:
 	pass
 
-from sqlflow_submitter.db import connect, execute, insert_values
+from sqlflow_submitter.db import connect, execute, insert_values, db_generator, db_generator_predict
```
uuleon, line 196: The `test.sh` script runs in the container of the latest `sqlflow` image, we need to commit the update of `db.py` firstly in another PR before using the newly added API, otherwise `CI` will fail.

typhoonzero: I think `db.py` could be covered by CI because the PR's code branch is mounted under `$GOPATH` and we always import from `$GOPATH` in the dockerfile env settings.

sql/python/sqlflow_submitter/db.py
typhoonzero: Good point! Wil do that!

sql/codegen.go
uuleon: You are right, this is not a problem.

sql/python/sqlflow_submitter/db.py
typhoonzero: I've been thinking about this when I'm writing this PR, adding this type if conditional check may let this generator function a bit hard to understand, and inside the generator, adding runtime checking string compare also slow down the execution.

sql/python/sqlflow_submitter/db.py
typhoonzero: I'll do this in next PR so that it's cleaner

wangkuiyi (dismissed):

sql/python/sqlflow_submitter/db.py
```diff
@@ -55,6 +56,64 @@ def execute(driver, conn, statement):
     return field_names, field_columns
 
 
+def db_generator(driver, conn, statement,
```
tonyyang-svail, line 59: Would you mind adding tests to these two functions? :)

sql/codegen.go
```diff
@@ -276,59 +259,98 @@ classifier = {{.Estimator}}(
...
+			feature_types[name] = tf.int64
+			feature_shapes[name] = tf.TensorShape([None])
+		else:
+			feature_types[name] = tf.float32
+			feature_shapes[name] = tf.TensorShape([])
+
+	gen = db_generator(driver, conn, """{{.StandardSelect}}""",
+		feature_column_names, "{{.Y.Name}}", column_name_to_type)
+	dataset = tf.data.Dataset.from_generator(gen, (feature_types, tf.int64), (feature_shapes, tf.TensorShape([1])))
+	dataset = dataset.shuffle(1000).repeat().batch(batch_size)
```
tonyyang-svail: How are `.shuffle` and `.repeat` implemented if the whole dataset is not loaded in the memory?

tonyyang-svail: Answer: to achieve `.repeat`, `dataset` will call `gen` multiple times, which execute the SQL statements multiple times.

sql/codegen.go
```diff
@@ -276,59 +259,98 @@ classifier = {{.Estimator}}(
 	{{end}}
 
 {{if .Train}}
-def train_input_fn(features, labels, batch_size):
-    dataset = tf.data.Dataset.from_tensor_slices((dict(features), labels))
-    dataset = dataset.shuffle(1000).repeat().batch(batch_size)
-    return dataset
+def train_input_fn(batch_size):
```
tonyyang-svail: `train_input_fn` and `eval_input_fn` are becoming more and more complicated. Shall we move it to `sql/python/sqlflow_submitter/input_fn.py` and create corresponding tests?

sql/codegen.go
```diff
@@ -276,59 +259,98 @@ classifier = {{.Estimator}}(
...
-    steps=STEPS * (EPOCHS if EPOCHS else 1))
+	input_fn=lambda:train_input_fn(BATCHSIZE),
+	steps=STEPS * (EPOCHS if EPOCHS else 1))
 {{end}}
 
-def eval_input_fn(features, labels, batch_size):
-    dataset = tf.data.Dataset.from_tensor_slices((dict(features), labels))
-    dataset = dataset.batch(batch_size)
-    return dataset
+def eval_input_fn(batch_size):
```
weiguoz: It looks like `eval_input_fn` and `train_input_fn` are similar.

sql/codegen.go
typhoonzero: I would like to keep it in here because these functions are using a lot of global variables defined above, moving it to another file means we must pass those variables as parameters. These parameters including:

- `feature_column_names`
- `column_name_to_type`
- `"""{{.StandardSelect}}"""`
- `"{{.Y.Name}}"`
- `driver`
- `conn`

sql/codegen.go
typhoonzero: Done!

sql/python/sqlflow_submitter/db.py
typhoonzero: Done.

wangkuiyi (approved):

PR #472, opened 2019-05-30
Remove unused Python files

### LINKED ISSUE #465: Remove unused Python files
@hungry1526 mentioned that the following Python files are no longer used and should be removed.
1. `sql/python/*.py`
1. `sql/python/sqlflow_submitter/models`

tonyyang-svail:
Fix #465


sql/python/sqlflow_submitter/db_test.py
```diff
@@ -39,20 +40,27 @@ def test_mysql(self):
             password = os.environ.get('SQLFLOW_TEST_DB_MYSQL_PASSWD') or "root"
             host = "127.0.0.1"
             port = "3306"
-            database = "iris"
-            conn = connect(driver, database, user=user, password=password, host=host, port=port)
```
tonyyang-svail, line 43: Python tests shouldn't assume the existence of `iris` database which is created by go test.

wangkuiyi (approved):

sql/python/sqlflow_submitter/db_test.py
typhoonzero: Seems `test_generator` I added before need this update too. Would you mind updating that part please? Thanks

tonyyang-svail: @typhoonzero thanks for reminding. Will do!

sql/python/README.md
```diff
@@ -1,13 +1,16 @@
 # Python Code Template
```
hungry1526, line 1: Since we are removing sql_data and the test, should we move the readme.md under submitter folder? Or we will rename python folder to submitter and move python code on level up

tonyyang-svail:

PR #474, opened 2019-05-30
Verify that TO TRAIN works

### LINKED ISSUE #473: Change SQLFlow extended SQL syntax from TRAIN into TO TRAIN
As mentioned in https://github.com/sql-machine-learning/sqlflow/pull/450#discussion_r287923105, SQL syntax allows aliases of table names. (Thank to Jian Zhang from PingCap helped me realized this.)

The following SQL statements are all legal:

```sql
SELECT * FROM t1
SELECT * FROM t1 t2 // t2 is an alias of t1
SELECT * FROM t1, t2 // t1 and t2 are table names
SELECT * FROM t1 t2, t3 // t1 and t3 are table names, t2 is an alias of t1
SELECT * FROM t1, t2, t3 t4 // t2 is an alias of t1, t4 is an alias of t3
```

So, if SQLFlow extends SQL SELECT statements by allowing users to append a TRAIN clause, we would see that the external parser (Calcite parser, TiDB parser, or Hive SQL parser) accepts the following statement until they meet the token "DNNClassifier", because they would think TRAIN is an alias of table t1.

```sql
SELECT * FROM t1 TRAIN DNNClassifier
                       ^
```

The above behavior is not what we want; we want the external parsers err on TRAIN instead of DNNClassifier.

A possible solution is to change the way SQLFlow extends SELECT statements from appending TRAIN to appending TO TRAIN.

```sql
SELECT * FROM t1 TO TRAIN DNNClassifier
```

wangkuiyi:
Fix https://github.com/sql-machine-learning/sqlflow/issues/473

### DIFF
--- a/parser/calcite/calcite_parser_test.go
+++ b/parser/calcite/calcite_parser_test.go
@@ -23,6 +23,7 @@ import (
 func TestCalciteParser(t *testing.T) {
 	if addr := os.Getenv("SQLFLOW_CALCITE_PARSER"); len(addr) > 0 {
 		Init(addr)
+		defer Cleanup()
 	} else {
 		t.Logf("Cannot connect to CalciteParserServer; skip TestCalciteParser")
 		return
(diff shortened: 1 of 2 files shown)


wangkuiyi:
Our CI would not run `calcite_parser_test.go` because of the if-condition at https://github.com/sql-machine-learning/sqlflow/pull/474/files#diff-16febf1d1acf2fcf0059d1661db3c6d1R24.  But anyway, I ran this test locally and it passed.

typhoonzero (approved):
LGTM++

PR #477, opened 2019-05-31
WIP Add HiveSQL parser as an external parser

wangkuiyi:


wangkuiyi:
It seems that I don't have to build the Hive QL parser from source code, but I can call parser from the pre-built Hive jar.

PR #480, opened 2019-05-31
Parser support multiple columns for DNNLinearCombinedEstimator

### LINKED ISSUE #467: SQL syntax on multiple feature columns as model inputs
SQLFlow syntax considers all fields in `COLUMN` as one feature column list. For example
```sql
SELECT ...
TRAIN DNNClassifier
COLUMN col1, col2, col3, col4
INTO ...
```
all columns in `COLUMN` will be translated to `feature_list` and passed as `feature_columns` to `DNNClassifier`.
```python
model = DNNClassifier(feature_columns=feature_list, ...)
```

However, some models may have multiple feature columns as inputs. Consider [the wide and deep model written in TensorFlow estimator](https://www.tensorflow.org/api_docs/python/tf/estimator/DNNLinearCombinedClassifier), whose constructor looks like the following
```python
model = DNNLinearCombinedClassifier(
    linear_feature_columns=linear_feature_list,
    dnn_feature_columns=deep_feature_list, ...)
```
Notice the `linear_feature_columns` and `dnn_feature_columns` are separate arguments. 

So we may need to propose a new SQL notation. Here I listed some proposals that @uuleon and I came up with. In all proposals, column `col1` and`col2` will passed as `linear_feature_columns`, and column `col3` and`col4` will be passed as `dnn_feature_columns`.

**Proposal 1**

Use `AS` keyword to denote different feature columns.
```sql
SELECT ...
TRAIN DNNLinearCombinedClassifier
COLUMN col1, col2 AS linear_feature_columns, col3, col4 AS dnn_feature_columns
INTO ...
```

**Proposal 2**

Use multiple `COLUMN` s to denote there are multiple feature columns and use `AS` to name the corresponding parameter names.
```sql
SELECT ...
TRAIN DNNLinearCombinedClassifier
COLUMN col1, col2 AS linear_feature_columns
COLUMN col3, col4 AS dnn_feature_columns
INTO ...
```

**Proposal 3**

Use special function `feature_column("parameter_name", ...)` to denote multiple feature columns.
```sql
SELECT ...
TRAIN DNNLinearCombinedClassifier
COLUMN
    feature_column("linear_feature_columns", col1, col2),
    feature_column("dnn_feature_columns", col1, col2)
INTO ...
```

The above proposals are still at a very early stage. Any comments are welcomed. :)

uuleon:
Fix #467


sql/sql.y
```diff
@@ -166,12 +173,16 @@ select
...
-	$$.columns = $6
-	$$.label = $8
-	$$.save = $10
+  }
+| train_clause COLUMN columns_clause {
+	$$.ccs = append($$.ccs, $3)
+  }
+| train_clause LABEL IDENT INTO IDENT {
+	$$.label = $3
+	$$.save = $5
```
wangkuiyi: There is a problem with this change, as it makes our parser accepts the following writings


```
TRAIN DNNClassifiere LABEL a INTO model1
TRAIN DNNClassifiere LABEL a INTO model1 LABEL b INTO model2
TRAIN DNNClassifiere LABEL a INTO model1 LABEL b INTO model2 LABEL c INTO model3
```

However, we want only one LABEL clause and one INTO clause.

My understanding is that we may not change too much of the rule train; instead, we might want just a little change from `COLUMN columns` into `column_clause`:

```
train : TRAIN IDENT column_clause LABEL IDENT INTO IDENT
```

where 

```
column_clause
: COLUMN exprlist
: column_groups

column_groups
: COLUMN exprlist FOR IDENT
| column_groups COLUMN exprlist FOR IDENT
````

Please be aware that the above rules don't accept

```
COLUMN a, b COLUMN c, d
```

which is what we want.

File: sql/sql.y [hunk 2]
```diff
@@ -74,11 +74,16 @@
...
 		estimator string
 		attrs     attrs
-		columns   exprlist
+		ccs   	  []columnsClause
                 label     string
 		save      string
 	}
 
+	type columnsClause struct {
+		name 	string
```
wangkuiyi: If `name` is the identifier after `FOR`, I think it would be better named `target` or `purpose` than `name`.

sql/sql.y
```diff
@@ -351,7 +367,21 @@ func (s trainClause) JSON() string {
...
+	ccs := []string{}
+	for _, cc := range s.ccs {
+		ccs = append(ccs, cc.JSON())
+	}
+	columns := "[\n" + strings.Join(ccs, ",\n") + "\n]"
+
+	return fmt.Sprintf(fmter, s.estimator, s.attrs.JSON(), columns, s.save)
+}
+
+func (s columnsClause) JSON() string {
```
wangkuiyi: I vaguely remember that the `JSON` functions were for some old purpose are no longer needed. Could @tonyyang-svail please confirm.

tonyyang-svail: Right. It's safe to remove it. :)

sql/sql.y
uuleon: Thanks for pointing that out, it's indeed a problem, will fix it.

sql/sql.y
uuleon: Ok

sql/sql.y
uuleon: As your suggestion, I have recovered the rule of `train_clause` and do the change only in `column_clause`.

Because I don't want to do the restriction that the location of `COLUMN ... FOR` must be after the anonymous `COLUMN` or there must be only one anonymous `COLUMN`, the rule of `column_clause` looks like this:
```
column_clause
: COLUMN columns
| COLUMN columns FOR IDENT
| column_clause COLUMN columns
| column_clause COLUMN columns FOR IDENT
```

If there is more than one anonymous `COLUMN` or `COLUMN`s which have the same `FOR IDENT`, the latter one makes effects.

@wangkuiyi What do you think?

sql/codegen.go
```diff
@@ -104,7 +104,7 @@ func newFiller(pr *extendedSelect, fts fieldTypes, db *DB) (*filler, error) {
 		r.Attrs[k] = v.String()
 	}
 
-	for _, c := range pr.columns {
+	for _, c := range pr.cc[""] {
```
wangkuiyi: cc => columns, or some other more meaningful name.

sql/codegen_alps.go
```diff
@@ -429,7 +429,8 @@ func newALPSTrainFiller(pr *extendedSelect) (*alpsFiller, error) {
 	}
 	modelDir := fmt.Sprintf("%s/model/", scratchDir)
 
-	fcList, fsMap, err := resolveTrainColumns(&pr.columns)
+	cols := pr.cc[""]
+	fcList, fsMap, err := resolveTrainColumns(&cols)
```
wangkuiyi, line 433: Why do we need the assignment to `cols`? Can we have just one line as the following?

```go
	fcList, fsMap, err := resolveTrainColumns(&(pr.cc[""]))
```

There are a few more cases below.

sql/parser_test.go
```diff
@@ -39,6 +39,19 @@ COLUMN
...
 INTO sqlflow_models.my_dnn_model;
+`
+	testMultiColumnTrainSelect = testStandardSelectStmt + `TRAIN DNNClassifier
+WITH
+  n_classes = 3,
+  hidden_units = [10, 20]
+COLUMN
+  employee.name,
+  bucketize(last_name, 1000),
+  cross(embedding(emplyoee.name), bucketize(last_name, 1000))
```
wangkuiyi, line 50: This COLUMN clause doesn't have the FOR keyword. Is that alright?

sql/sql.y
```diff
@@ -74,11 +73,13 @@
 	type trainClause struct {
 		estimator string
 		attrs     attrs
-		columns   exprlist
+		cc   	  columnClause
```
wangkuiyi: cc => columns or some other more meaningful name.

sql/sql.y
```diff
@@ -74,11 +73,13 @@
...
                 label     string
 		save      string
 	}
 
+	type columnClause map[string]exprlist
```
wangkuiyi, line 82: If there is only one column in the column clause, and there is no FOR in that column, how do we handle the case? Are we going to have only one entry in the map[string]exprlist and the key is ""?  If so, please add a comment here to explain.

sql/sql.y
```diff
@@ -182,6 +185,19 @@ predict_clause
 }
 ;
 
+column_clause
+: COLUMN columns {
+	$$ = make(map[string]exprlist, 0)
+	$$[""] = $2
```
wangkuiyi: We can merge the above two lines into 

```go
$$ = map[string]exprlist{"" : $2}
```

sql/sql.y
```diff
@@ -182,6 +185,19 @@ predict_clause
...
+  }
+| COLUMN columns FOR IDENT {
+	$$ = make(map[string]exprlist, 0)
+ 	$$[$4] = $2
```
wangkuiyi: Again,

```go
$$ = map[string]exprlist{$4 : $2}
```

wangkuiyi (commented):
Great improvement. Just some comments about Go conventions.

sql/sql.y
uuleon: You are right, will add a comment here.

sql/sql.y
uuleon: Ok

sql/parser_test.go
uuleon: That's alright, the `FOR` keyword is optional.

sql/codegen_alps.go
uuleon: GO forbid to take the address of a `map` member because of the address may be invalid if the map grows or shrinks.

sql/codegen.go
uuleon: Ok

sql/sql.y
uuleon: Ok

sql/sql.y
uuleon: Ok

sql/sql.y
```diff
@@ -187,9 +187,8 @@ predict_clause
 ;
 
 column_clause
-: COLUMN columns 				{ $$ = map[string]exprlist{"" : $2} }
-| COLUMN columns FOR IDENT 			{ $$ = map[string]exprlist{"" : $2} }
-| column_clause COLUMN columns 			{ $$[""] = $3 }
+: COLUMN columns 				{ $$ = map[string]exprlist{"feature_columns" : $2} }
+| COLUMN columns FOR IDENT 			{ $$ = map[string]exprlist{$4 : $2} }
 | column_clause COLUMN columns FOR IDENT 	{ $$[$5] = $3 }
```
wangkuiyi, line 192: This rule accepts the case that the first COLUMN doesn't have FOR but the subsequent COLUMNs have FORs, like the following example:

```
COLUMN a, b COLUMN b, c FOR in1 COLUMN d, f FOR in2
```

However, I think it is alright at this time. If we realize we do need more strict grammar rules, we can change later.

wangkuiyi (approved):

PR #482, opened 2019-06-03
Support batch operators for MaxCompute

weiguoz:


sql/python/sqlflow_submitter/maxcompute.py
```diff
@@ -19,19 +20,67 @@ class MaxCompute:
...
+    
+            r = inst.open_reader(tunnel=True, compress_option=compress)
+            field_names = None if r._schema.columns is None \
+                    else [col.name for col in r._schema.columns]
+            label_idx = field_names.index(label_column_name)
+    
+            i = 0
+            while i < r.count:
+                expected = r.count-i if r.count-i < fetch_size else fetch_size
+                for row in [[v[1] for v in rec] for rec in r[i: i+expected]]:
```
typhoonzero, line 53: Just curious about is indexing reader like `r[i: i+expected]` read only a part from the table?

sql/python/sqlflow_submitter/maxcompute.py
```diff
@@ -19,19 +20,67 @@ class MaxCompute:
...
+                    features = dict()
+                    for name in feature_column_names:
+                        if column_name_to_type[name] == "categorical_column_with_identity":
+                            cell = np.fromstring(row[field_names.index(name)], dtype=int, sep=",")
+                        else:
+                            cell = row[field_names.index(name)]
+                        features[name] = cell
+                    yield features
+                i += expected
+            print("done db_generator_predict")
```
typhoonzero: Can remove this line

typhoonzero (dismissed):
LGTM with some minor comments

tonyyang-svail (dismissed):
`sql/codegen.go` part LGTM.

I am approving this PR because it contains the hotfix on #487

sql/codegen.go
```diff
@@ -309,37 +309,36 @@ print("Training set accuracy: {accuracy:0.5f}".format(**eval_result))
...
-		feature_column_names, column_name_to_type)
-	dataset = tf.data.Dataset.from_generator(gen, feature_types, feature_shapes)
-	dataset = dataset.batch(batch_size)
-	return dataset
+def eval_input_fn(features, batch_size):
+	dataset = tf.data.Dataset.from_tensor_slices(dict(features))
+	return dataset.batch(batch_size)
 
+field_names, columns = execute(driver, conn, """{{.StandardSelect}}""")
 X = {}
```
typhoonzero, line 332: I think this is just a hotfix right, maybe should refine the `insert` function to insert predict result with original feature data.

typhoonzero (approved):

sql/codegen.go
weiguoz: Sure. I just recovered it by the former code here.

sql/python/sqlflow_submitter/maxcompute.py
weiguoz: The iterator in the loop calls [function: read](https://github.com/aliyun/aliyun-odps-python-sdk/blob/master/odps/models/instance.py#L808) which returns a generator here to read records on demand.

PR #483, opened 2019-06-04
More detailed error messages

tonyyang-svail:


sql/model.go
```diff
@@ -50,7 +50,13 @@ func (m *model) save(db *DB, table string) (e error) {
 
 	cmd := exec.Command("tar", "czf", "-", "-C", m.workDir, ".")
 	cmd.Stdout = sqlf
-	return cmd.Run()
+	var errBuf bytes.Buffer
+	cmd.Stderr = &errBuf
```
wangkuiyi, line 54: If you really want to capture all outputs (stderr + stdout), you can replace `os/exec.Run` by `os/exec.Output`, without the need to create `errBuf`.

However, the problem is that either method (setting `cmd.Stderr` to a buffer or calling `os/exec.Output`) might cause memory overflow when the output is lengthy.

What I have in my mind is to write a io.Writer implementation which keeps only the most recent N bytes. And call it like the following:

```go
cmd.Stderr := NewLastNBytesBuffer(N)
```

wangkuiyi: Maybe something like

```go
type LastNBytesBuffer {
    *bytes.Buffer // inherits all methods from bytes.Buffer.
    n int // the fixed-size.
}

func NewLastNBytesBuffer(n int) *LastNBytesBuffer {
    return &LastNBytesBuffer{new(bytes.Buffer), n}
}

func (w *LastNBytesBuffer) Write(s []byte) (int, error) {
    if w.Len() + len(s) > w.n {
        if len(s) > n {
            w.Buffer = bytes.NewBuffer(s[len(s)-w.n:])
            return len(s), nil
        } else {
            w.Buffer = bytes.NewBuffer(w.Bytes()[w.Len() + len(s) - w.n:])
            return w.Write(s)
        }
    }
    return w.Write(s)
}
```

wangkuiyi (approved):

sql/model.go
wangkuiyi: I thought too much more than needed -- this is just for capturing errors from `tar`, which couldn't be too much.

PR #484, opened 2019-06-04
Add logChanWriter close

tonyyang-svail:
`logChanWriter` should be flushed in the Close.


wangkuiyi (approved):
I think we can merge this PR, but we need a subsequent PR to fix https://github.com/sql-machine-learning/sqlflow/issues/486.

PR #489, opened 2019-06-05
Fix imdb load dataset

typhoonzero:
`keras.datasets.imdb` seems only work with `numpy==1.16.1`


wangkuiyi (approved):

Dockerfile
```diff
@@ -8,13 +8,15 @@ ARG CONDA_OS=Linux
...
 
+# keras.datasets.imdb only works with numpy==1.16.1
 RUN curl -sL https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -o mconda-install.sh && \
     bash -x mconda-install.sh -b -p miniconda && \
     rm mconda-install.sh && \
     /miniconda/bin/conda create -y -q -n sqlflow-dev python=3.6 ${CONDA_ADD_PACKAGES} && \
     echo ". /miniconda/etc/profile.d/conda.sh" >> ~/.bashrc && \
     echo "source activate sqlflow-dev" >> ~/.bashrc && \
     /bin/bash -c "source /miniconda/bin/activate sqlflow-dev && python -m pip install \
+    numpy==1.16.1 \
```
wangkuiyi, line 19: It would be great if we can add a comment here explaining the requirement of the certain version.

typhoonzero: The comment is on top of `RUN` line since we cannot add a comment inside one `RUN` command.

PR #492, opened 2019-06-05
Remove unused CONDA_OS ARG in dev Dockerfile

terrytangyuan:
Please correct me if I am wrong but this does not seem used currently.


typhoonzero (approved):
LGTM!

PR #503, opened 2019-06-10
Fix broken links in code walkthrough

terrytangyuan:
These links now direct to 404 ([example](https://sql-machine-learning.github.io/sqlflow/cmd/sqlflowserver/main.go)) since the documents are in `/doc`. I am not familiar with the setup of the website so here switching to use their absolute urls temporarily.

### DIFF
--- a/doc/walkthrough.md
+++ b/doc/walkthrough.md
@@ -25,23 +25,23 @@ If a SQL statement is of the standard syntax, SQLFlow throws it to the SQL engin
 
 ### SQLFlow as a gRPC Server
 
-SQLFlow is a gRPC server, which can connect with multiple clients.  A typical client is [pysqlflow](https://github.com/sql-machine-learning/pysqlflow), the SQLFlow plugin for Jupyter Notebook server.  Another once is a text-based client [/cmd/sqlflowserver/main.go](/cmd/sqlflowserver/main.go).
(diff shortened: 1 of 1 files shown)


doc/walkthrough.md
```diff
@@ -25,23 +25,23 @@ If a SQL statement is of the standard syntax, SQLFlow throws it to the SQL engin
 
 ### SQLFlow as a gRPC Server
 
-SQLFlow is a gRPC server, which can connect with multiple clients.  A typical client is [pysqlflow](https://github.com/sql-machine-learning/pysqlflow), the SQLFlow plugin for Jupyter Notebook server.  Another once is a text-based client [/cmd/sqlflowserver/main.go](/cmd/sqlflowserver/main.go).
+SQLFlow is a gRPC server, which can connect with multiple clients.  A typical client is [pysqlflow](https://github.com/sql-machine-learning/pysqlflow), the SQLFlow plugin for Jupyter Notebook server.  Another once is a text-based client [/cmd/sqlflowserver/main.go](https://github.com/sql-machine-learning/sqlflow/tree/develop/cmd/sqlflowserver/main.go).
```
typhoonzero, line 28: Better to use relative paths? Former code redirect from `/` so that it was not working I guess.

terrytangyuan: Relative paths work on the markdown file but won’t work on the actual website since the source code is not available under https://sqlflow.org. It depends on how the website is set up. As I mentioned in the description, this should be a temporary solution so at least no 404 is returned.

typhoonzero: I see.

typhoonzero (approved):

terrytangyuan:
@typhoonzero You can just go ahead and merge this. I don’t have write access.

PR #508, opened 2019-06-11
add model location for demo

weiguoz:
Follow [PR:Save model on the local filesystem](https://github.com/sql-machine-learning/sqlflow/pull/488)

### DIFF
--- a/cmd/demo/demo.go
+++ b/cmd/demo/demo.go
@@ -83,6 +83,7 @@ func render(rsp interface{}, table *tablewriter.Table) bool {
 
 func main() {
 	ds := flag.String("datasource", "", "database connect string")
+	modelDir := flag.String("model_dir", "", "model would be saved on the local dir, otherwise upload to the table.")
 	flag.Parse()
 	db, err := sql.Open(*ds)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


Yancey0623 (approved):
LGTM

cmd/demo/demo.go
```diff
@@ -83,6 +83,7 @@ func render(rsp interface{}, table *tablewriter.Table) bool {
 
 func main() {
 	ds := flag.String("datasource", "", "database connect string")
+	modelDir := flag.String("model_dir", "", "model would be saved on the local dir, otherwise upload to the table.")
```
wangkuiyi, line 86: "save the model to SQL engine as a human-unreadable table if this option is empty"

cmd/demo/demo.go
```diff
@@ -101,7 +108,7 @@ func main() {
 		isTable, tableRendered := false, false
 		table := tablewriter.NewWriter(os.Stdout)
 
-		stream := sql.Run(slct, db, "")
+		stream := sql.Run(slct, db, *modelDir)
```
wangkuiyi, line 111: Is it a better idea to create the modelDir in `sql.Run` than here in `main`? I vaguely feel that it is easier to maintain the code if we allocate the resource right before we use it.

wangkuiyi (approved):

cmd/demo/demo.go
weiguoz: Good idea.
SQLFlow needs a model directory only a SQL is extended.

PR #509, opened 2019-06-11
Polish TestStandardDropTable

### LINKED ISSUE #505: [Failed] TestStandardDropTable
image

tonyyang-svail:
Fix https://github.com/sql-machine-learning/sqlflow/issues/505

We expect our parser to raise an error on parsing non-extendedSQL.

### DIFF
--- a/sql/parser_test.go
+++ b/sql/parser_test.go
@@ -139,7 +139,7 @@ func TestSelectStarAndPrint(t *testing.T) {
 }
 
 func TestStandardDropTable(t *testing.T) {
-	if _, e := newParser().Parse(`DROP TABLE PREDICT`); e != nil {
-		t.Skipf("[FIXME]`drop table` expected no error, but got:%v", e)
-	}
+	a := assert.New(t)
+	_, e := newParser().Parse(`DROP TABLE PREDICT`)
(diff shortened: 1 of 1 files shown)


sql/parser_test.go
```diff
@@ -139,7 +139,7 @@ func TestSelectStarAndPrint(t *testing.T) {
 }
 
 func TestStandardDropTable(t *testing.T) {
-	if _, e := newParser().Parse(`DROP TABLE PREDICT`); e != nil {
-		t.Skipf("[FIXME]`drop table` expected no error, but got:%v", e)
-	}
+	a := assert.New(t)
+	_, e := newParser().Parse(`DROP TABLE PREDICT`)
+	a.Error(e)
```
wangkuiyi, line 144: The fact is that currently, our parser doesn't accept anything statements other than SELECT. It will support parsing any SQL statements and even dialects in the future.

I think we can either copy-n-paste the above explanation after the call to `a.Error(e)` as a comment, or simply remove this test.

tonyyang-svail: Sure. Copy-n-pasted the explanation.

wangkuiyi (approved):

PR #510, opened 2019-06-12
predict use generator to support predict on large dataset

typhoonzero:


sql/codegen.go
```diff
@@ -324,52 +324,70 @@ print("Training set accuracy: {accuracy:0.5f}".format(**eval_result))
 print("Done training")
 {{- else}}
 
-def eval_input_fn(features, batch_size):
-	dataset = tf.data.Dataset.from_tensor_slices(dict(features))
-	return dataset.batch(batch_size)
-
-field_names, columns = execute(driver, conn, """{{.StandardSelect}}""")
```
weiguoz, line 331: Since the `execute` is removed, we may remove it from the [import line](https://github.com/sql-machine-learning/sqlflow/pull/510/files#diff-a9c90abc2107d6cd0f30e093777900bfR200) either. 
What is more, I would suggest move the `execute` from `db.py` and `maxcompute.py` to `db_test.py`.

sql/codegen.go
```diff
@@ -324,52 +324,70 @@ print("Training set accuracy: {accuracy:0.5f}".format(**eval_result))
...
-    insert_values(driver, conn, "{{.TableName}}", field_names, val)
-
-insert("{{.TableName}}", X)
+def insert(table_name, eval_input_dataset, feature_column_names, predictions, insert_batch_size=64):
+	column_names = feature_column_names[:]
+	column_names.append("{{.Y.Name}}")
+	pred_rows = []
+	while True:
+		try:
+			in_val = eval_input_dataset.__next__()
```
weiguoz, line 369: I like this good practice!

sql/codegen.go
```diff
@@ -324,52 +324,70 @@ print("Training set accuracy: {accuracy:0.5f}".format(**eval_result))
...
-    field_names = [key for key in X]
-
-    val = []
-    for i in range(length[0]):
-        val.append(tuple([str(X[f][i]) for f in field_names]))
-
-    insert_values(driver, conn, "{{.TableName}}", field_names, val)
-
-insert("{{.TableName}}", X)
+def insert(table_name, eval_input_dataset, feature_column_names, predictions, insert_batch_size=64):
```
terrytangyuan, line 363: This hard-coded `insert_batch_size=64` probably needs to be exposed to users?

typhoonzero: Thanks, this should be chosen automatically, added TODO comment.

sql/codegen.go
typhoonzero: Done.

weiguoz (approved):
LGTM

PR #511, opened 2019-06-13
Hiveql parser - stage 0

wangkuiyi:
This PR is the early stage of HiveQL's parser. 

It verifies that we don't have to build the parser from the source code of Hive; instead, we can call the parser from the pre-built Hive. 

It verifies that we don't have to work with ANTLR, the converter that generates Java source code from the grammar rules; instead, we call a wrapper of the generated parser.

It verifies that the parser reports the location of the syntax error if there is any.


typhoonzero (commented):
LGTM, please @weiguoz take a look?

parser/hiveql/pom.xml
```diff
@@ -0,0 +1,96 @@
...
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <version>4.10</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.hive</groupId>
+            <artifactId>hive-exec</artifactId>
+            <version>2.1.0</version>
```
weiguoz, line 27: Is it related to the hive server?
Should we change the version number here as a hive server's version is 1.2.1?

weiguoz (approved):
LGTM ++

parser/hiveql/pom.xml
wangkuiyi: Yes, I tried. It works with Hive 1.2.1 too. I will update the version in the next PR. Thanks for the reminder.

parser/hiveql/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+```bash
+$ javar -jar /path/to/hiveql-parser.jar /path/to/your-code.sql
+```
+
+## Examples
+
+```bash
+$ java -jar hiveql-parser.jar <(echo "select count(*) as count, myfield from &0rz") 2>/dev/null
+[1,39]: line 1:39 cannot recognize input near '&' '0rz' '<EOF>' in join source
```
weiguoz, line 45: After ran
```bash
java -jar target/hiveql-parser-1.0-SNAPSHOT-jar-with-dependencies.jar < (echo "select count(*) as count, myfield from &0rz") 2>/dev/null
```
I got result: `bash: syntax error near unexpected token `('`

weiguoz: > After ran
> 
> ```shell
> java -jar target/hiveql-parser-1.0-SNAPSHOT-jar-with-dependencies.jar < (echo "select count(*) as count, myfield from &0rz") 2>/dev/null
> ```
> 
> I got result: `bash: syntax error near unexpected token `('`

My bad, `<(` cannot be separated by a space.

weiguoz: I see, `<(` cannot be separated by a space.

wangkuiyi: Will fix this by changing the command line into `-jar target/hiveql-parser-1.0-SNAPSHOT.jar` in the next PR. Thanks for the reminder!

PR #502, opened 2019-06-10
support categorical_column_identity and embedding column

uuleon:
1. support categorical_column_identity and embedding column in `codegen_alps`
2. refactoring the unit test of `codegen_alps`
3. bugfix

PR #512, opened 2019-06-14
seperate expression resolver so all submitter can use

typhoonzero:


sql/codegen_alps.go
```diff
@@ -65,346 +51,13 @@ type alpsFiller struct {
...
-
-	headName := (*el)[0].val
-	if headName == "" {
-		return resolveExpression(&(*el)[0].sexp)
-	}
-
-	headName = strings.ToUpper(headName)
-
-	switch headName {
-	case dense:
```
tonyyang-svail, line 253: I think `DENSE` and `SPARSE` only work for ALPS. You can refer to the generated code at https://github.com/sql-machine-learning/sqlflow/pull/502#issuecomment-501077347 where `DENSE` and `SPARSE` are translated to `DenseColumn` and `SparseColumn` in `DataX`, instead of `tf.feature_column.*` in the model constructor.

@uuleon I am not sure if ALPS has a plan for releasing these `DENSE` and `SPARSE` as part of the extension for feature column.

tonyyang-svail (approved):
LGTM.

sql/codegen_alps.go
typhoonzero: Thanks for reminding, I am moving on to make this resolver work for both submitters in later PRs.

PR #528, opened 2019-06-21
Use resolvedTrainClause in codegen_alps

### LINKED ISSUE #526: Codegen_alps should get ODPS credential from db object
As we have added gomaxcompute in sqlflow, the following line should be changed to `gomaxcompute.FormatDSN(db.dataSourceName)`

https://github.com/sql-machine-learning/sqlflow/blob/95cc60823f5f5ee70aaaf5ba500775e7d592247f/sql/codegen_alps.go#L136-L141

uuleon:
1. Renaming `CAT_ID` to `CATAGORY_ID`
2. Use `maxcompute.config`. Fix #526 
3. Create `resolvedTrainClause` in `expression_resolver.go` and use it in `codegen_alps.go`

### DIFF
--- a/sql/codegen_alps.go
+++ b/sql/codegen_alps.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"sqlflow.org/gomaxcompute"
 	"strings"
 	"text/template"
 )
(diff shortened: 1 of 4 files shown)


sql/codegen_alps.go
```diff
@@ -39,137 +40,63 @@ type alpsFiller struct {
...
+	for _, css := range resolved.ColumnSpecs {
+		for _, cs := range css {
+			csCode = append(csCode, cs.ToString())
+		}
+	}
 
-	for target, columns := range pr.columns {
-		fcs, fss, err := resolveTrainColumns(&columns)
+	var odpsConfig = &gomaxcompute.Config{}
+	if db != nil {
```
tonyyang-svail, line 83: What happens if `db == nil`? Or, should we change to
```go
if db == nil {
        return nil, fmt.Errorf(...)
}
odpsConfig, err := gomaxcompute.ParseDSN(db.dataSourceName);
if err != nil {
        return nil, err
}
```

tonyyang-svail (approved):
LGTM!

sql/codegen_alps.go
uuleon: `db == nil` happens in UnitTest

tonyyang-svail: @uuleon why do we need to test `db == nil`?

PR #533, opened 2019-06-26
refine and fix codegen_alps

typhoonzero:


wangkuiyi (dismissed):

sql/codegen_alps.go
```diff
@@ -224,82 +223,79 @@ from alps.io.reader.odps_reader import OdpsReader
...
+        )
+    )
+
+    export_path = "{{.ModelDir}}"
+
+    experiment = Experiment(
+        user="sqlflow",
+        engine=LocalEngine(),
+        train=TrainConf(input=trainDs,
+{{if (ne .TrainClause.MaxSteps -1)}}
```
tonyyang-svail, line 282: Just curious, is the parenthesis necessary of the if template?

tonyyang-svail (approved):

sql/codegen_alps.go
typhoonzero: Yes, tested on my environment. Or the number `-1` will be recognized as an argument of the function call `.TrainClause.MaxSteps`

PR #541, opened 2019-07-01
alps support multi-group

joyyoj:
Multi-group means user stores multi sparse FeatureColumns  into  one db's column. 
sql 
EMBEDDING(CATEGORY_ID(deep, 2000), 8, mean)
If deep is a multi-group column with group = 3, means really have 3 embedding column
deep_0, deep_1, deep_2
A feature map is a table that mapping input's id of type string/int64 to a int id range (0, shape).
Currently, a feature map table conforms to {train_table}_feature_map

### DIFF
--- a/sql/codegen_alps.go
+++ b/sql/codegen_alps.go
@@ -464,6 +464,7 @@ func (meta *metadata) getColumnInfo(featureColumns map[string][]featureColumn) (
 	return columns, nil
 }
 
+// get all referenced field names.
 func getAllKeys(fcs []featureColumn) []string {
 	output := make([]string, 0)
 	for _, fc := range fcs {
--- a/sql/codegen_alps.go
+++ b/sql/codegen_alps.go
@@ -15,13 +15,14 @@ package sql
 
 import (
 	"bytes"
+	"database/sql"
 	"fmt"
 	"io"
 	"io/ioutil"
+	"sqlflow.org/gomaxcompute"
+	"strconv"
 	"strings"
 	"text/template"
-
-	"sqlflow.org/gomaxcompute"
 )
 
 type alpsFiller struct {
@@ -42,8 +43,9 @@ type alpsFiller struct {
 	Y      string
 
 	// Train
-	ModelCreatorCode string
-	TrainClause      *resolvedTrainClause
+	ModelCreatorCode  string
+	FeatureColumnCode string
+	TrainClause       *resolvedTrainClause
 
 	// Feature map
 	FeatureMapTable     string
@@ -51,6 +53,11 @@ type alpsFiller struct {
 
 	// ODPS
 	OdpsConf *gomaxcompute.Config
+}
+
+type alpsFeatureColumn interface {
+	featureColumn
+	GenerateAlpsCode(metadata *metadata) ([]string, error)
 }
 
 func modelCreatorCode(resolved *resolvedTrainClause, args []string) (string, error) {
@@ -80,22 +87,6 @@ func newALPSTrainFiller(pr *extendedSelect, db *DB) (*alpsFiller, error) {
 		return nil, err
 	}
 
-	featureMapTable := ""
-	featureMapPartition := ""
-
-	csCode := make([]string, 0)
-	for _, css := range resolved.ColumnSpecs {
-		for _, cs := range css {
-			csCode = append(csCode, cs.ToString())
-			if cs.FeatureMap.Table != "" {
-				featureMapTable = cs.FeatureMap.Table
-			}
-			if cs.FeatureMap.Partition != "" {
-				featureMapPartition = cs.FeatureMap.Partition
-			}
-		}
-	}
-
 	var odpsConfig = &gomaxcompute.Config{}
 	if db != nil {
 		odpsConfig, err = gomaxcompute.ParseDSN(db.dataSourceName)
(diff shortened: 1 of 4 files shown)


sql/codegen_alps.go
```diff
@@ -18,10 +18,9 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"sqlflow.org/gomaxcompute"
```
typhoonzero, line 22: Use [pre-commit](https://pre-commit.com) to automatically fix style checks.

sql/codegen_alps.go
```diff
@@ -80,65 +85,58 @@ func newALPSTrainFiller(pr *extendedSelect, db *DB) (*alpsFiller, error) {
...
 	var odpsConfig = &gomaxcompute.Config{}
 	if db != nil {
 		odpsConfig, err = gomaxcompute.ParseDSN(db.dataSourceName)
 		if err != nil {
 			return nil, err
 		}
 	}
+	fields := make([]string, 0) // TODO use complete fields
+	csCode := make([]string, 0)
+	// TODO use table attributes ?
```
typhoonzero: TODO comment should look like `// TODO(user): ...`

sql/codegen_alps.go
```diff
@@ -80,65 +85,58 @@ func newALPSTrainFiller(pr *extendedSelect, db *DB) (*alpsFiller, error) {
...
+
+	fmap := featureMap{"sqlflow_softmax_estimator_feature_map", ""}
```
typhoonzero: Need to get `feature_map` table name from SQL

sql/metadata.go
```diff
@@ -0,0 +1,146 @@
...
+)
+
+type Metadata struct {
+	odpsConfig *gomaxcompute.Config
+	table      string
+	featureMap *featureMap
+	columnInfo *map[string]*columnSpec
+}
+
+func (meta *Metadata) GetColumnInfo(featureColumns map[string][]featureColumn) (map[string]*columnSpec, error) {
```
typhoonzero: Is `Metadata` for getting `columnSpec` automatically, not by using SQL lines in `COLUMN` clause?

Can we use `COLUMN` clause first and move to this auto way in another PR?

sql/metadata.go
```diff
@@ -0,0 +1,146 @@
+package sql
```
tonyyang-svail: @joyyoj could you please add `metadata_test.go`?

sql/codegen_alps.go
```diff
@@ -195,6 +194,109 @@ func submitALPS(w *PipeWriter, pr *extendedSelect, db *DB, cwd string) error {
 	return nil
 }
 
+func (nc *numericColumn) GenerateAlpsCode(metadata *Metadata) ([]string, error) {
```
tonyyang-svail: How is this function different from 
https://github.com/sql-machine-learning/sqlflow/blob/3867f64c17e852671acadc5d4ab6d68438aba5cc/sql/expression_resolver.go#L641-L644

sql/metadata.go
```diff
@@ -0,0 +1,146 @@
...
+
+func (meta *Metadata) GetColumnInfo(featureColumns map[string][]featureColumn) (map[string]*columnSpec, error) {
+	allKeys := make([]string, 0)
+	for _, fcs := range featureColumns {
+		fields := GetAllKeys(fcs)
+		allKeys = append(allKeys, fields...)
+	}
+	columns, err := meta.GetSparseColumnInfo()
+	if err != nil {
+		fmt.Println(err)
```
weiguoz: We use [logrus](github.com/sirupsen/logrus) to log errors
https://github.com/sql-machine-learning/sqlflow/blob/3867f64c17e852671acadc5d4ab6d68438aba5cc/sql/executor.go#L149

sql/metadata.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+	query := fmt.Sprintf("SELECT feature_type, max(cast(id as bigint)) as feature_num, group "+
+		"FROM %s WHERE %s GROUP BY group, feature_type", meta.featureMap.Table, filter)
+
+	sqlDB, _ := sql.Open("maxcompute", meta.odpsConfig.FormatDSN())
+	rows, err := sqlDB.Query(query)
+	if err != nil {
+		return output, err
+	}
+	defer sqlDB.Close()
```
weiguoz: `defer sqlDB.Close()` should follow `sql.Open(...)`

sql/codegen_alps.go
joyyoj: fixed

sql/metadata.go
typhoonzero: Talked with @joyyoj , that this feature seems needed for the model we are going to support.

sql/codegen_alps.go
```diff
@@ -178,7 +195,7 @@ func submitALPS(w *PipeWriter, pr *extendedSelect, db *DB, cwd string) error {
 		return fmt.Errorf("submitALPS: failed executing template: %v", err)
 	}
 	code := program.String()
-
+	fmt.Println(code)
```
typhoonzero: Can remove this line.

sql/codegen_alps.go
```diff
@@ -195,6 +212,118 @@ func submitALPS(w *PipeWriter, pr *extendedSelect, db *DB, cwd string) error {
 	return nil
 }
 
+func (nc *numericColumn) GenerateAlpsCode(metadata *metadata) ([]string, error) {
+	output := make([]string, 0)
+	output = append(output,
```
typhoonzero, line 216: Maybe can call one function here to wrap generated code to a list like:
```go
return ToOneElementList(nc.GenerateCode)
```

sql/codegen_alps_test.go
```diff
@@ -47,12 +47,12 @@ func TestTrainALPSFiller(t *testing.T) {
 
 	a.True(filler.IsTraining)
 	a.Equal("kaggle_credit_fraud_training_data", filler.TrainInputTable)
-	a.Equal("[\"dense\",\"deep\",\"wide\"]", filler.Fields)
-	a.True(strings.Contains(filler.X, "SparseColumn(name=\"deep\", shape=[2000], dtype=\"int\", separator='\x01', group_separator='\x02')"), filler.X)
-	a.True(strings.Contains(filler.X, "SparseColumn(name=\"wide\", shape=[1000], dtype=\"int\", separator='\x01', group_separator='\x02')"), filler.X)
+	//a.Equal("[\"dense\",\"deep\",\"wide\"]", filler.Fields)
```
typhoonzero: can remove commented code.

typhoonzero (dismissed):
LGTM, can fix comments in next PR.

sql/codegen_alps.go
joyyoj: GenerateCode return (string, error)
GenerateAlpsCode return ([]string, error)

joyyoj: for multi group,
category_column(deep) needs return category_column(deep0), category_column(deep_1) ...
so i introduce generateAlpsCode.

I thinks GenerateCode be a function of feature column is a bad idea. GenerateCode should be a part of physical-plan.

sql/metadata.go
joyyoj: i think `defer sqlDB.Close()` should not be called if error occurs ?
Encouraged by:
https://github.com/go-sql-driver/mysql/wiki/Examples#a-word-on-sqlopen

sql/codegen_alps.go
joyyoj: forgot it

typhoonzero (approved):

PR #550, opened 2019-07-04
The architecture diagram does not display properly

StephenRi:
The relative path is wrong, and the architecture diagram does not display properly.


typhoonzero (approved):
LGTM

PR #556, opened 2019-07-05
Fix alps custom estimator

joyyoj:
Alps support custom estimator.

### DIFF
--- a/sql/codegen_alps.go
+++ b/sql/codegen_alps.go
@@ -105,19 +105,17 @@ func newALPSTrainFiller(pr *extendedSelect, db *DB) (*alpsFiller, error) {
 	// TODO(joyyoj) pr may contains partition.
 	fmap := featureMap{pr.tables[0] + "_feature_map", ""}
 	var meta metadata
-	selectFields := pr.standardSelect.fields
 	fields := make([]string, 0)
-	for _, field := range selectFields {
-		if field != pr.label {
-			fields = append(fields, field)
-		}
-	}
 	if db != nil {
 		odpsConfig, err = gomaxcompute.ParseDSN(db.dataSourceName)
 		if err != nil {
 			return nil, err
 		}
 		meta = metadata{odpsConfig, pr.tables[0], &fmap, nil}
+		fields, err = getFields(&meta, pr)
+		if err != nil {
+			return nil, err
+		}
 		columnInfo, err = meta.getColumnInfo(resolved, fields)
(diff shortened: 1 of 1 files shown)


sql/codegen_alps.go
```diff
@@ -43,6 +43,7 @@ type alpsFiller struct {
 	Y      string
 
 	// Train
+	ImportCode        string
```
typhoonzero, line 46: maybe the name `CustomEstimatorImportCode` make more sence?

sql/codegen_alps.go
```diff
@@ -89,20 +100,28 @@ func newALPSTrainFiller(pr *extendedSelect, db *DB) (*alpsFiller, error) {
 
 	var odpsConfig = &gomaxcompute.Config{}
 	var columnInfo map[string]*columnSpec
-	var meta metadata
 
 	// TODO(joyyoj) read feature mapping table's name from table attributes.
 	// TODO(joyyoj) pr may contains partition.
 	fmap := featureMap{pr.tables[0] + "_feature_map", ""}
+	var meta metadata
+	selectFields := pr.standardSelect.fields
```
typhoonzero: What if `SELECT *`?

sql/codegen_alps.go
```diff
@@ -111,7 +130,7 @@ func newALPSTrainFiller(pr *extendedSelect, db *DB) (*alpsFiller, error) {
 		}
 		meta.columnInfo = &columnInfo
 	}
-	fields := make([]string, 0) // TODO use complete fields
+	//fields := make([]string, 0) // TODO use complete fields
```
typhoonzero: please clean commented code

sql/codegen_alps.go
joyyoj: I think ImportCode is better for it is more general. It's probable not limited to custom estimator imports In future.

sql/codegen_alps.go
joyyoj: I know this problem. I try use 'desc table' clause to get fields but maxcompute report "not support instance type". I add a TODO here.

joyyoj: Use `select * from table limit 1 `clause to get fields. if table is not empty, it works

typhoonzero (approved):
LGTM++

sql/codegen_alps.go
typhoonzero: Cool~

PR #560, opened 2019-07-06
Fix parsing shape

### LINKED ISSUE #559: Shape is not parsed correctly
**Description**
When parsing `NUMERIC(name, [5,10])` in
```SQL
SELECT ...
TRAIN ...
COLUMN NUMERIC(name, [5,10])
...
```
`[5,10]` won't be parsed as `[`, `5`, `,`, `10`, and `]`, but `[`, `5,10`, `]` instead.

**Reproduction Steps**
```go
func TestNumericColumn(t *testing.T) {
	a := assert.New(t)
	parser := newParser()

	// Note it is [5,10] without the space
	normal := statementWithColumn("NUMERIC(c2, [5,10])")
	moreArgs := statementWithColumn("NUMERIC(c1, 100, args)")
	badShape := statementWithColumn("NUMERIC(c1, bad)")

	r, e := parser.Parse(normal)
	a.NoError(e)
	c := r.columns["feature_columns"]
	fcs, _, e := resolveTrainColumns(&c)
	a.NoError(e)
	nc, ok := fcs[0].(*numericColumn)
	a.True(ok)
	code, e := nc.GenerateCode()
	a.NoError(e)
	a.Equal("c2", nc.Key)
	a.Equal([]int{5, 10}, nc.Shape)
	a.Equal("tf.feature_column.numeric_column(\"c2\", shape=[5,10])", code)
}
```

tonyyang-svail:
Fix #559.

### DIFF
--- a/sql/lexer.go
+++ b/sql/lexer.go
@@ -138,8 +138,7 @@ func (l *lexer) emitIdentOrKeyword(lval *sqlSymType) int {
 }
 
 var (
-	// Stolen https://www.regular-expressions.info/floatingpoint.html.
-	reNumber = regexp.MustCompile("[-+]?[0-9]*.?[0-9]+([eE][-+]?[0-9]+)?")
+	reNumber = regexp.MustCompile("[-+]?[0-9]*[.]?[0-9]+([eE][-+]?[0-9]+)?")
 )
 
 func (l *lexer) lexNumber(lval *sqlSymType) int {
--- a/sql/lexer_test.go
+++ b/sql/lexer_test.go
@@ -56,6 +56,16 @@ func TestLexNumber(t *testing.T) {
 	var n sqlSymType
 	a.Equal(NUMBER, l.Lex(&n))
 	a.Equal("123.4", n.val)
+
+	l = newLexer("[5,10]")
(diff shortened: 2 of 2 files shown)


sql/lexer.go
```diff
@@ -138,8 +138,7 @@ func (l *lexer) emitIdentOrKeyword(lval *sqlSymType) int {
 }
 
 var (
-	// Stolen https://www.regular-expressions.info/floatingpoint.html.
-	reNumber = regexp.MustCompile("[-+]?[0-9]*.?[0-9]+([eE][-+]?[0-9]+)?")
+	reNumber = regexp.MustCompile("[-+]?[0-9]*[.]?[0-9]+([eE][-+]?[0-9]+)?")
```
tonyyang-svail, line 141: `.` will match any character. What we want is the exact match of character `.`.

typhoonzero: does that mean should `\.` also works?

typhoonzero (approved):
LGTM

tonyyang-svail:
> does that mean should `\.` also works?

@typhoonzero Right. Both `"\\."` and `"[.]"` will work.

PR #562, opened 2019-07-06
Alps support distributed running

joyyoj:
user can set worker num and ps num
Currently, support follows attrs:
* engine.worker_num
* engine.worker_memory
* engine.ps_num
* engine.ps_memory

```sql
...
train dnnclassifier 
with model.n_classes = 2, engine.worker_num = 2, engine.ps_num = 1
```
If worker_num > 1, it will run distributed training.

### DIFF
--- a/sql/expression_resolver.go
+++ b/sql/expression_resolver.go
@@ -33,6 +33,18 @@ const (
 	comma         = "COMMA"
 )
 
+type resourceSpec struct {
+	Num    int
+	Memory int
+	Core   int
+}
+
+type engineSpec struct {
+	etype  string
+	ps     resourceSpec
+	worker resourceSpec
+}
+
 type resolvedTrainClause struct {
 	IsPreMadeModel         bool
 	ModelName              string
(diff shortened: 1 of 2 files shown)


typhoonzero:
Please try resolve conflict by running `git pull upstream develop`, thanks

sql/expression_resolver.go
```diff
@@ -128,152 +141,114 @@ type embeddingColumn struct {
...
+	if p, ok := attrs["type"]; ok {
+		strVal, ok := p.Value.(string)
+		if ok {
+			engineType = strVal
 		}
-		delete(attrs, p.FullName)
+	} else if psNum > 0 || workerNum > 0 {
+		engineType = "k8s"
 	}
+	return engineSpec{
```
typhoonzero, line 169: Is there a way to let the submitter just return the yarn job URL and exit? then the status can be fetched on the web page and do not wait on the sqlflow client (e.g. jupyter notebook)

joyyoj: I can modify to exit immediately, but how the web page can get the status?

typhoonzero (approved):
LGTM++

PR #565, opened 2019-07-09
fix develop maxcompute CI

typhoonzero:
Fix develop CI error: https://travis-ci.com/sql-machine-learning/sqlflow/builds/118374096?utm_medium=notification&utm_source=email

### DIFF
--- a/sql/codegen.go
+++ b/sql/codegen.go
@@ -456,14 +456,10 @@ def insert(table_name, eval_input_dataset, feature_column_names, predictions, in
         {{end}}
         pred_rows.append(tuple(row))
         if len(pred_rows) == insert_batch_size:
-            insert_cursor = insert_values(driver, write_conn, table_name, column_names, pred_rows)
-            if driver != "maxcompute":
-                insert_cursor.close()
+            insert_values(driver, write_conn, table_name, column_names, pred_rows)
             pred_rows.clear()
     if len(pred_rows) > 0:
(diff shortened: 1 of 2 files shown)


sql/codegen.go
```diff
@@ -457,11 +457,13 @@ def insert(table_name, eval_input_dataset, feature_column_names, predictions, in
         pred_rows.append(tuple(row))
         if len(pred_rows) == insert_batch_size:
             insert_cursor = insert_values(driver, write_conn, table_name, column_names, pred_rows)
-            insert_cursor.close()
```
tonyyang-svail, line 460: How about we change it to `close(driver, insert_curser)` and in the `db.py`, we add
```python
def close(driver, insert_curser):
    if driver != "maxcompute":
        insert_cursor.close()
```
So that we can avoid `if driver != "maxcompute":` in the codegen_template.

sql/python/sqlflow_submitter/db.py
```diff
@@ -114,5 +114,4 @@ def insert_values(driver, conn, table_name, table_schema, values):
     cursor = conn.cursor()
     cursor.executemany(statement, values)
     conn.commit()
-
-    return cursor
+    cursor.close()
```
tonyyang-svail, line 117: ~~do we need~~
```python
if driver != "maxcompute":
    cursor.close()
```
~~here?~~

tonyyang-svail (approved):
LGTM.

PR #577, opened 2019-07-10
Alps fix column info

joyyoj:
allow specify column's dtype and fix column's shape.

### DIFF
--- a/sql/codegen_alps.go
+++ b/sql/codegen_alps.go
@@ -69,7 +69,9 @@ func engineCreatorCode(resolved *resolvedTrainClause) (string, error) {
 		return "LocalEngine()", nil
 	}
 	engine := resolved.EngineParams
-	return fmt.Sprintf("YarnEngine(ps = ResourceConf(memory=%d, num=%d),worker=ResourceConf(memory=%d, num=%d))",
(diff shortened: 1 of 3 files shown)


sql/codegen_alps.go
```diff
@@ -674,6 +698,8 @@ func (meta *metadata) getSparseColumnInfo() (map[string]*columnSpec, error) {
 		}
 		name := values[0].(*string)
 		ishape, _ := strconv.Atoi(*values[1].(*string))
+		ishape++
```
typhoonzero, line 701: Curious why need to `++` the shape?

sql/codegen_alps.go
```diff
@@ -629,12 +649,16 @@ func (meta *metadata) getDenseColumnInfo(keys []string) (map[string]*columnSpec,
...
 		}
-		for _, ct := range columnTypes {
-			denseValue := values[0].(*string)
+		for idx, ct := range columnTypes {
+			denseValue := values[idx].(*string)
 			fields := strings.Split(*denseValue, ",")
 			shape := make([]int, 1)
 			shape[0] = len(fields)
-			output[ct.Name()] = &columnSpec{ct.Name(), false, false, shape, "float", ",", *meta.featureMap}
+			if userSpec, ok := refColumns[ct.Name()]; ok {
```
typhoonzero, line 657: Maybe add comments that user column definitions are used firstly.

typhoonzero (approved):
LGTM++

PR #578, opened 2019-07-11
fix alps engine

joyyoj:


sql/expression_resolver.go
```diff
@@ -160,6 +160,9 @@ func getEngineSpec(attrs map[string]*attribute) engineSpec {
 		if p, ok := attrs[key]; ok {
 			strVal, ok := p.Value.(string)
 			if ok {
+				if strings.HasPrefix(strVal, "\"") && strings.HasSuffix(strVal, "\"") {
```
typhoonzero, line 164: Should add todo comment here, we can use the parser to do mostly these validations work.

typhoonzero (dismissed):

sql/expression_resolver.go
joyyoj: fixed

typhoonzero (approved):

PR #583, opened 2019-07-16
Split extended select via Lexer token

tonyyang-svail:
The first step on supporting arbitrary select clause.


typhoonzero (approved):
LGTM++

PR #585, opened 2019-07-17
customized specify feature map

uuleon:


tonyyang-svail:

PR #591, opened 2019-07-17
Design: Support arbitrary select statements

tonyyang-svail:


doc/support_arbitrary_select.md
```diff
@@ -0,0 +1,62 @@
...
+### Modification on SQLFlow parser
+
+The SQLFlow parser is only responsible for parsing the train clause, instead of the full extended SQL statement. So the
+[`standardSelect`](https://github.com/sql-machine-learning/sqlflow/blob/158b098cfecf7b12479171b09c12b877ad3fb00b/sql/sql.y#L66-L71)
+struct will be a string.
+
+### Splitting the extended SQL
+
+SQLFlow splits the extended SQL by looking for consecutive tokens returned by the lexer. If SQLFlow finds the following consecutive tokens 
+[`TRAIN`, `IDENT`, `WITH`] or [`PREDICT`, `IDENT`, `USING`], it splits the SQL string at the beginning of the first token
```
typhoonzero, line 34: Should we use `TO TRAIN` to minimize the possibility that `TRAIN IDENT WITH` are all column names? It's a very tiny chance that this could happen but what we need should be a 100% sure the statement is split as we expected.

typhoonzero: NOTE: related PR: https://github.com/sql-machine-learning/sqlflow/pull/583

tonyyang-svail: Good point.

@wangkuiyi could you please file an issue explaining why we need to change from `TRAIN` to `TO TRAIN`? Then I will submit a PR to fix that issue.

tonyyang-svail:
Close due to lack of activity. I will file a new PR for the design.

PR #592, opened 2019-07-18
add sqlflow sql with join test

typhoonzero:
Add test case for testing bypass SQL statements with `JOIN`.

### DIFF
--- a/cmd/sqlflowserver/main_test.go
+++ b/cmd/sqlflowserver/main_test.go
@@ -104,7 +104,7 @@ func AssertContainsAny(a *assert.Assertions, all map[string]string, actual *any.
 		b := wrappers.StringValue{}
 		ptypes.UnmarshalAny(actual, &b)
 		if _, ok := all[b.Value]; !ok {
-			a.Failf("string value %s not exist", b.Value)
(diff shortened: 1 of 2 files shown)


cmd/sqlflowserver/main_test.go
```diff
@@ -736,3 +739,52 @@ LABEL "label" INTO model_table;`, caseDB)
...
+	}
+	dbConnStr = "mysql://root:root@tcp/?maxAllowedPacket=0"
+	modelDir := ""
+
+	tmpDir, caCrt, caKey, err := generateTempCA()
+	defer os.RemoveAll(tmpDir)
+	if err != nil {
+		t.Fatalf("failed to generate CA pair %v", err)
+	}
+	go start("", modelDir, caCrt, caKey, true)
```
tonyyang-svail: `TestEnd2EndMySQL` also starts an SQLFlow server. They will conflict on port 50051.

tonyyang-svail (approved):

cmd/sqlflowserver/main_test.go
typhoonzero: Thanks for fixes!

PR #594, opened 2019-07-22
support configure async cluster job submit from user's session

typhoonzero:
Can generate `ALPS` code that choose from below mode according to user "session":
- exit immediately after the training job is submitted to the cluster.
- wait until the job is done.

Related PR for `pysqlflow` will be updated here.


Yancey0623 (approved):
LGTM

PR #606, opened 2019-07-26
Split table into two views for training&validation

### LINKED ISSUE #569: training and validation(2):  split temporary table for training and validation
Following design doc: training and validation
PART: How to split

We fetch the training/validation data using two different queries respectively：
`SELECT * FROM temp_table WHERE sqlflow_random < 0.8`
`SELECT * FROM temp_table WHERE sqlflow_random >= 0.8`
...

weiguoz:
Fix #569 

**Notice** 
It's different from the [design doc](https://github.com/sql-machine-learning/sqlflow/blob/develop/doc/training_and_validation.md).
`ALPS` doesn't support using a query to specify the dataset for training(or validation), it reads all records from a  table instead. Hence, SQLFlow creates 2 views to fulfill this.
Of cause, I will fix the design doc to follow the code.

### DIFF
--- a/sql/validation.go
+++ b/sql/validation.go
@@ -20,38 +20,82 @@ import (
 	"github.com/pkg/errors"
 )
 
+type trainingDataset struct {
+	// `supported` here identifies if SQLFlow is able to split dataset into
+	// training dataset and validation dataset.
+	// So, TODO(weiguo): Let's remove `supproted` if SQLFlow supports other
+	// drivers, like: MySQL, hive(specified in database.go:open()).
+	supported      bool
+	table          string
+	trainingView   string // view of table (<k)
+	validationView string // view of table (>=k)
+}
+
 const (
 	temporaryTableLifecycle = 14 // day(s)
 	randomColumn            = "sqlflow_rdm"
-	randomTablePrefix       = "sqlflow_tv_" // 'tv' = training & validation
+	tablePrefix             = "sqlflow_tv_" // 'tv' = training & validation
+	trainingViewPrefix      = "sqlflow_view_training_"
+	validationViewPrefix    = "sqlflow_view_validation_"
 )
 
-var errNotSupportYet = errors.New("not support yet")
+var (
+	errBadBoundary = errors.New("boundary should between (0.0, 1.0) exclude")
+)
 
 // SQLFlow generates a temporary table, + sqlflow_randowm column
-func tableWithRandomColumn(db *DB, standardSelect string) (string, error) {
+func createTrainingDataset(db *DB, slct string, trainingUpperbound float32) (trainingDataset, error) {
+	if trainingUpperbound <= 0 || trainingUpperbound >= 1 {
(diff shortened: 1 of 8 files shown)


sql/validation.go
```diff
@@ -20,38 +20,82 @@ import (
...
+	return nil
 }
 
-func randomTableName() string {
-	n := time.Now().UnixNano() / 1e3
-	return fmt.Sprintf("%s%d", randomTablePrefix, n)
+func namingTrainingDataset() trainingDataset {
+	uniq := time.Now().UnixNano() / 1e3
+	return trainingDataset{
+		supported:      true,
```
tonyyang-svail: Is `supported` field necessary?

- If a database DOES support `namingTrainingDataset`, `supported` field is always true.
- If a database DOES NOT support `namingTrainingDataset`, the function should raise an error.

sql/validation.go
```diff
@@ -20,38 +20,82 @@ import (
...
-	tbl := randomTableName()
-	createStmt := fmt.Sprintf("CREATE TABLE %s LIFECYCLE %d AS SELECT *, RAND() AS %s FROM (%s) AS %s_ori", tbl, temporaryTableLifecycle, randomColumn, standardSelect, tbl)
-	if _, e := db.Exec(createStmt); e != nil {
-		log.Errorf("create temporary table failed, statememnt:[%s], err:%v", createStmt, e)
-		return "", e
+func releaseTrainingDataset(ds trainingDataset) {
+	// TODO(weiguo): release resources for databases, like: "hive", "mysql"...
+}
+
+func createMaxcomputeDataset(db *DB, slct string, trainingUpperbound float32) (trainingDataset, error) {
```
tonyyang-svail: Minor suggestion: how about returning `(*trainingDataset, error)` with the pointer, so that we can use `return nil, e` on error.

tonyyang-svail (dismissed):
Generally LGTM. Left a few comments for potential improvements.

sql/validation.go
```diff
@@ -20,38 +20,82 @@ import (
 	"github.com/pkg/errors"
 )
 
+type trainingDataset struct {
```
typhoonzero: maybe use the name `trainAndValDataset` ?

typhoonzero: Also, this source file name `validation.go` seems is not what this file does.

tonyyang-svail: > Also, this source file name `validation.go` seems is not what this file does.

Good point. How about `split.go` or `create_train_val.go`?

sql/validation.go
weiguoz: `supported` is a temporary variable for `codegen`. Here it used to avoid damage to `CI`

https://github.com/weiguoz/sqlflow/blob/5618433e2490ed50b72d48af06d77eef6d18ca72/sql/validation.go#L24-L27

tonyyang-svail (approved):

PR #617, opened 2019-07-30
Customized module from gitlab

uuleon:


sql/expression_resolver.go
```diff
@@ -47,6 +47,15 @@ type engineSpec struct {
 	queue   string
 }
 
+type gitLabModule struct {
```
typhoonzero, line 50: `gitLabModel` ?

sql/expression_resolver.go
```diff
@@ -47,6 +47,15 @@ type engineSpec struct {
...
+	ModuleName   string
+	ProjectName  string
+	Sha          string
+	PrivateToken string
+	SourceRoot   string
+	GitLabServer string
```
typhoonzero, line 56: Maybe define some URI schema could reduce the number of parameters, like `token@https://gitlab.yourhost.com/your/repo.git:sha:/path/to/dir:ModelClassName`

uuleon: Is this a standard URL format for `gitlab`?

sql/expression_resolver.go
uuleon: Ok, will fix it

sql/expression_resolver.go
typhoonzero: Not a common standard, note we also use `maxcompute://xxx` to identify the database drivers. This is a method to define a pattern and reduce the number of paramters.

typhoonzero (dismissed):
LGTM, can merge when the CI of the test case passed.

typhoonzero (approved):

PR #622, opened 2019-07-30
Make verifier unaware of the table to be selected 

### LINKED ISSUE #621: Make verifier unaware of the table to be selected
Part of "Support arbitrary select statements" #620 

The current verifier implementation assumes the parser knows which table to select, which is no longer valid after the support of arbitrary select statements.

tonyyang-svail:
Fix #621.


typhoonzero (approved):
LGTM++

PR #627, opened 2019-07-31
Replace view by table while creating dataset for training and validation

weiguoz:
ALPS does not support the [view](https://dev.mysql.com/doc/refman/8.0/en/create-view.html) dataset for training or validation: `the specified method is not allowed against virtual view.`

### DIFF
--- a/sql/create_train_val.go
+++ b/sql/create_train_val.go
@@ -25,18 +25,18 @@ type trainAndValDataset struct {
 	// training dataset and validation dataset.
 	// So, TODO(weiguo): Let's remove `supproted` if SQLFlow supports other
 	// drivers, like: MySQL, hive(specified in database.go:open()).
-	supported      bool
-	table          string
-	trainingView   string // view of table (<k)
-	validationView string // view of table (>=k)
+	supported  bool
+	table      string
+	training   string // table for training: < k
+	validation string // table for validation: >= k
 }
 
 const (
 	temporaryTableLifecycle = 14 // day(s)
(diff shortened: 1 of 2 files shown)


sql/create_train_val.go
```diff
@@ -64,27 +64,27 @@ func releaseTrainAndValDataset(ds *trainAndValDataset) {
...
-	if e := createView(db, ds.table, ds.validationView, validationCond); e != nil {
+	if e := createMaxcomputeTable(ds.validation, ds.table, db, validationCond); e != nil {
 		return nil, e
 	}
 	return ds, nil
 }
 
-func createView(db *DB, table, view, where string) error {
-	stmt := fmt.Sprintf("CREATE VIEW %s AS SELECT * FROM %s WHERE %s", view, table, where)
+func createMaxcomputeTable(target, origin string, db *DB, cond string) error {
```
typhoonzero, line 84: Is there a way to drop the temp table after the training is finished?

typhoonzero (approved):
LGTM++

sql/create_train_val.go
weiguoz: PR([release resources](http://gitlab.alipay-inc.com/Arc/sqlflow/blob/develop/docker_alps/Dockerfile)) will take over it, but I would like to use [LIFECYCLE](https://github.com/sql-machine-learning/sqlflow/blob/develop/doc/training_and_validation.md#release-the-temporary-table) to handle it for Maxcompute at this time.

Issue #624, opened 2019-07-31
Train and val


weiguoz:
Following the design: [training_and_validation](https://github.com/sql-machine-learning/sqlflow/blob/develop/doc/training_and_validation.md)

SQLFlow supports train&val on [Maxcompute](https://www.alibabacloud.com/product/maxcompute) for `alps` through 4 PRs:
- [x] [generate a temporary](https://github.com/sql-machine-learning/sqlflow/issues/568)
- [x] [split temporary table](https://github.com/sql-machine-learning/sqlflow/issues/569) 
- [x] [codegen for alps](https://github.com/sql-machine-learning/sqlflow/issues/570)
- [x] [release temporary table](https://github.com/sql-machine-learning/sqlflow/issues/571)

Next, we are going to implement:
- [x] [support train&val on Maxcompute for TF](https://github.com/sql-machine-learning/sqlflow/pull/638)
- [x] [support train&val on Hive](https://github.com/sql-machine-learning/sqlflow/issues/644)
- [x] [support train&val on MySQL](https://github.com/sql-machine-learning/sqlflow/issues/684)

PR #638, opened 2019-08-07
Supporting train&val on Maxcompute for TF

weiguoz:
A part of [Train and val](https://github.com/sql-machine-learning/sqlflow/issues/624).


sql/codegen.go
```diff
@@ -248,309 +261,17 @@ func genTF(w io.Writer, pr *extendedSelect, ds *trainAndValDataset, fts fieldTyp
 	if e != nil {
 		return e
 	}
-	// TODO(weiguo): fix codegen to carry trainingDatase in the next PR
-	if e = codegenTemplate.Execute(w, r); e != nil {
-		return fmt.Errorf("genTF: failed executing template: %v", e)
+	if pr.train {
```
tonyyang-svail, line 264: How about
```go
if pr.train {
    return tfTrainTemplate.Execute(w, r)
}
return tfPredTemplate.Execute(w, r)
```

sql/template_tf.go
```diff
@@ -0,0 +1,388 @@
...
+feature_metas["{{$value.FeatureName}}"] = {
+    "feature_name": "{{$value.FeatureName}}",
+    "dtype": "{{$value.Dtype}}",
+    "delimiter": "{{$value.Delimiter}}",
+    "shape": {{$value.InputShape}},
+    "is_sparse": "{{$value.IsSparse}}" == "true"
+}
+{{end}}
+
+def get_dtype(type_str):
```
tonyyang-svail, line 96: shall we move `get_dtype`and `_parse_sparse_feature` to `sql/python` so that they be shared between templates?

tonyyang-svail (dismissed):
Generally LGTM.

sql/template_tf.go
weiguoz: `get_dtype` & `_parse_sparse_feature` are specified to `tensorflow`. I'm not sure if other templates would use this.

tonyyang-svail (approved):

PR #657, opened 2019-08-13
SQLFlow server should not panic and exit when encounter Lexer issues

### LINKED ISSUE #639: Sqlflow server should not panic and exit when encounter lexer issues
https://github.com/sql-machine-learning/sqlflow/blob/develop/sql/lexer.go#L94

Should log the error and return the error message to client but not panic here.

tonyyang-svail:
Fix #639

### DIFF
--- a/sql/executor.go
+++ b/sql/executor.go
@@ -58,6 +58,9 @@ func splitExtendedSQL(slct string) []string {
 	var pos []int
 	for {
 		t := l.Lex(&n)
+		if t == -1 {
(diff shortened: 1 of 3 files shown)


typhoonzero (commented):
Sorry I did not see this, my pr https://github.com/sql-machine-learning/sqlflow/pull/659 here also try to return the error message to the client.

tonyyang-svail:
@typhoonzero OK. Let's work on #659 then.

PR #659, opened 2019-08-14
No panic when lexer error

typhoonzero:


sql/executor.go
```diff
@@ -28,7 +28,17 @@ import (
...
 // Run executes a SQL query and returns a stream of rows or messages
 func Run(slct string, db *DB, modelDir string, session *pb.Session) *PipeReader {
-	if len(splitExtendedSQL(slct)) == 2 {
+	splitedSQL, err := splitExtendedSQL(slct)
+	if err != nil {
+		rd, wr := Pipe()
+		// return the lexer error message to client side
+		go func() {
+			defer wr.Close()
+			wr.Write(fmt.Sprintf("%v", err))
```
tonyyang-svail: Currently, the error is written directly to `wr`, for example,

https://github.com/sql-machine-learning/sqlflow/blob/5e7adc1ececc92995423097328eb3ca63c27f99b/sql/executor.go#L189-L193

So how about `wr.Write(fmt.Sprintf("%v", err))` => `wr.Write(err)`?

tonyyang-svail (dismissed):

sql/executor.go
typhoonzero: Done

tonyyang-svail (approved):

PR #678, opened 2019-08-16
Add design for feature derivation details

### LINKED ISSUE #589: [Feature Derivition] Thoughts about simpler column clause
For discussion, We intend to let `COLUMN` clause to define:

- how to decode the data of the column
- how to define feature_columns as inputs of the model

The definition is:

```
column_clause
: COLUMN columns
| COLUMN columns FOR IDENT
| COLUMN columns AS IDENT
| column_clause COLUMN columns FOR IDENT
;
```

columns with `FOR IDENT` are used as `feature_columns` for the model, else it stands for the column processors, e.g.

```
COLUMN numeric(col1, dtype) FOR features,
                category_id(csv(col2, size, dtype, delim, sparse), bucket_size) AS col2_multi_hot,
                embedding(col2_multi_hot, 128) FOR features,
                cross(col2_multi_hot, col1) FOR features,
                embedding(category_id(csv(col3, size, dtype, delm, dense), 1000), 32)
```

`col1` is a scalar input tensor and it will be fed into the model as a `feature_column`; `col2` is a `SparseTensor` and will be fed into the model using embedding feature column; `col3` is a dense tensor and will be fed into the model using embedding column.

Rules:

1. `FOR IDENT`, `IDENT` is the parameter that the model takes, like `tf.estimator.DNNLinearCombinedClassifier` takes `linear_feature_columns` and `dnn_feature_columns`
2. `AS IDENT` will declare an alias of the current column.
3. you can define multiple "feature processors` but only the last one will be used.

Supported processors:

- `numeric(col, dtype)` scalar column
- `csv(col, size, dtype, delim, [sparse/dense])`
- `enum(col, size)`

User customized processors:

You can define your own processor and use it in SQLFlow, to do that you must define the process using the SQLFlow processor register API:

```python
@register_sqlfow_processor(name="json")
def json_processor(cell_input):
    return json.loads(cell_input)
```

The above lines of code should appear in your custom model defination files, so that SQLFlow could run it.

typhoonzero:
Fix https://github.com/sql-machine-learning/sqlflow/issues/589

### DIFF
--- a/doc/feature_derivation.md
+++ b/doc/feature_derivation.md
@@ -8,25 +8,25 @@ Target: We need to know the below two information for each column after the
 feature type infer routine:
 
 1. How to transform the column data to tensors, including `tf.SparseTensor`.
-2. What type of feature column should adapt on the column and the parameters for the feature column call.
+2. What type of feature column should adapt to the column and the parameters for the feature column call.
 
 ## Simpler COLUMN Clause
 
 When we have a training table contains many columns that should be used
-for training like: https://www.kaggle.com/mlg-ulb/creditcardfraud, it's
+for training like https://www.kaggle.com/mlg-ulb/creditcardfraud, it's
 not friendly if we must provide all column names in `COLUMN` clause, we'd
 like to just use `COLUMN *` here.
 
 But, for cases when the columns are of quite different data format, like:
 
 ```
-data type:   (float, float, float, float, csv_string_for_dense_tensor, csv_string_for_sparse_tensor, int)
-column name: (a    , b    , c    , d    , e                          , f, label)
+data type: (float, float, float, float, csv_string_for_dense_tensor, csv_string_for_sparse_tensor, int)
+column name: (a , b , c , d , e , f, label)
 ```
 
 `COLUMN *` and plus the data in the table may not provide enough information we need:
 
-- whether the `csv_string_for_dense_tensor` should going through a embedding layer
+- whether the `csv_string_for_dense_tensor` should going through an embedding layer
 - what the original "dense shape" for the `csv_string_for_sparse_tensor` column to parse to a `SparseTensor`.
 
 So we need SQLFlow users to provide such information in the SQL statements to make the training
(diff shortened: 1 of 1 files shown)


doc/feature_derivation.md
```diff
@@ -0,0 +1,85 @@
...
+- what the original "dense shape" for the `csv_string_for_sparse_tensor` column to parse to a `SparseTensor`.
+
+So we need SQLFlow users to provide such information in the SQL statements to make the training
+network specific. We support the below SQL statements for the above case:
+
+```
+SELECT * FROM training_table
+TRAIN DNNClassifier
+WITH someattr=somevalue
+COLUMN *,EMBEDDING(e, 128, "sum"),EMBEDDING(SPARSE(f, [1000000]), 512, "sum")
```
tonyyang-svail: @typhoonzero Thanks for this syntax design.

Looks like the `*` in `select * from train_table` and the `*` in `COLUMN *` have different meanings. The first is the shorthand for all fields. And the later is rather complex, whose meaning depends on many other things.

1. Consider `select *, a from train_table`, the result columns are all fields in `train_table` plus a duplicated `a`. However, `COLUMN *, a` has the same columns as `COLUMN *`. The meaning of `*` is depending on `a`.
1. The `COLUMN *` in `COLUMN * LABEL label` doesn't include the label field. The meaning of `*` is depending on the label clause.

Do you think it is hard for the user to learn `*`?

How about introducing a function named `MATCH(regexp)` that will expand the regexp to match column names. So we can write `MATCH("[a-d]")` to represent columns `a`, `b`, `c` and `d`. So the train SQL would become
```SQL
SELECT * FROM training_table
TRAIN DNNClassifier
WITH someattr=somevalue
COLUMN MATCH("[a-d]"),EMBEDDING(e, 128, "sum"),EMBEDDING(SPARSE(f, [1000000]), 512, "sum")
LABEL label
INTO my_model_file;
```
In the case of [credit fraud](https://www.kaggle.com/mlg-ulb/creditcardfraud) dataset, we want to select `V1`, `V2`, ..., `V19`, we can write `MATCH("V\d+")`, which would exclude field `time` and `class`.

doc/feature_derivation.md
```diff
@@ -0,0 +1,85 @@
...
+
+## The Feature Derivation Routine
+
+We need to `SELECT` part of the training data, like 1000 rows and go through the below routine:
+
+1. If the column data type is numeric: int, bigint, float, double, can directly parse to a tensor of shape `[1]`.
+2. If the column data type is string: VARCHAR or TEXT:
+   1. If the string is not one of the supported serialized format (only support CSV currently):
+      1. If all the rows of the column's string data can be parsed to a float or int value,
+         treat it as a tensor of shape `[1]`.
```
tonyyang-svail, line 97: What if the column data type falls into this category (contents are "1", "2", "3", "4", ...), but a user write `SPARSE(column_name)`. Do we still categorize this column as dense tensor?

I believe the feature derivation logic should start with what feature column a user has written.

doc/feature_derivation.md
typhoonzero: As we discussed this morning, I am sure that what must be done by the SQL statements are:

1. Describe what columns to read from. It could be all columns in the table or several columns in some pattern.
1. Describe what are the columns used for training samples ( or input X ), and what is the columns used for the label (Y).
1. If there are a lot of X columns, we should have a fast way to match them all, e.g. credit card fraud dataset have time, v1~v18, amount features, we can match them by `time, MATCH(v[0-9][0-9]), amount`.
1. Describe how to parse the column data into tensors, e.g. parse a CSV string to a `tf.SparseTensor` or `tf.Tensor`, or preprocess the tensors like image random crop etc.
1. Describe how the feature column is assembled to the network, like whether to use embedding for the column, should the embedding column use shared weights etc.

doc/feature_derivation.md
typhoonzero: > I believe the feature derivation logic should start with what feature column a user has written.

You are right.

doc/feature_derivation.md
```diff
@@ -0,0 +1,108 @@
...
+When we have a training table contains many columns that should be used
+for training like https://www.kaggle.com/mlg-ulb/creditcardfraud, it's
+not friendly if we must provide all column names in `COLUMN` clause.
+Since we'd like to use all columns, when we write `SELECT *` then we can
+assume that we are using all columns to train and no longer need to write
+`COLUMN` anymore:
+
+```sql
+SELECT * FROM creditcardfraud
+TRAIN DNNClassifier
```
tonyyang-svail, line 25: which column is the label?

tonyyang-svail (dismissed):

tonyyang-svail (approved):
LGTM++

PR #681, opened 2019-08-19
add embedding support initializer param

typhoonzero:
Initializers difference may cause a model performance drop. We may set up model weights initializer using custom models, but not feature column of embedding.

Anyway, this is sort of some workaround, write initializer functions in SQL is not what we perfer.


Yancey0623 (approved):
lgtm

PR #682, opened 2019-08-19
Support train&val on HIVE

weiguoz:
Fix #644

PR #687, opened 2019-08-20
Support MySQL train&val

### LINKED ISSUE #684: Support train&val on MySQL
A part of [Train & val](https://github.com/sql-machine-learning/sqlflow/issues/624)

weiguoz:
Fix #684

### DIFF
--- a/sql/create_train_val.go
+++ b/sql/create_train_val.go
@@ -21,11 +21,6 @@ import (
 )
 
 type trainAndValDataset struct {
-	// `supported` here identifies if SQLFlow is able to split dataset into
-	// training dataset and validation dataset.
-	// So, TODO(weiguo): Let's remove `supproted` if SQLFlow supports other
-	// drivers, like: MySQL, hive(specified in database.go:open()).
-	supported  bool
 	database   string
 	table      string
 	training   string // table for training: < k
@@ -55,7 +50,7 @@ func newTrainAndValDataset(db *DB, slct string, origTable string, trainingUpperb
 		return createMaxcomputeDataset(db, slct, origTable, trainingUpperbound)
 	case "hive", "mysql":
 		return createDataset(db, slct, origTable, trainingUpperbound)
-	// TODO(weiguo) case "sqlite":
+	// TODO(weiguo) case "sqlite3":
 	default:
 		return nil, nil
 	}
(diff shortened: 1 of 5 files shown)


sql/create_train_val_test.go
```diff
@@ -39,20 +39,28 @@ func TestCreateTrainAndValDataset(t *testing.T) {
...
 		a.Error(e)
 		_, e = newTrainAndValDataset(testDB, testTrainAndValDataset, "orig", 0)
 		a.Error(e)
 		ds, e := newTrainAndValDataset(testDB, testTrainAndValDataset, "orig", 0.8)
 		a.NoError(e)
-		a.True(ds.supported)
-	case "hive":
-		_, e := newTrainAndValDataset(testDB, testHiveTrainAndValDataset, "orig", 1)
+		a.NotEmpty(ds.database)
+	case "mysql":
```
tonyyang-svail: Is the case `"mysql"` identical to the case `"hive"`, if so, how about using the shorthand `case "hive", "mysql":`?

tonyyang-svail (dismissed):
Excellent!

tonyyang-svail (approved):
LGTM.

sql/create_train_val.go
```diff
@@ -104,28 +98,29 @@ func createMaxcomputeTable(target, origin string, db *DB, cond string) error {
 	return e
 }
 
-func createHiveDataset(db *DB, slct string, origTable string, trainingUpperbound float32) (*trainAndValDataset, error) {
+// create dataset on Hive, MySQL
```
typhoonzero, line 101: function comments should be right above the function declaration and should be the format: `createDataset ....`

sql/create_train_val.go
```diff
@@ -174,7 +168,7 @@ func namingTrainAndValDataset(origTable string) *trainAndValDataset {
 
 func releaseTrainAndValDataset(db *DB, ds *trainAndValDataset) error {
```
typhoonzero, line 169: Note that we should not drop the validation table after training because we may still need to use that table to run a full evaluation. Can run this function before the training start yet. ( introduced in https://github.com/sql-machine-learning/sqlflow/pull/682/files#diff-d77fda4265f0434c48d4fe6252700e7dR284)

typhoonzero (changes requested):

sql/create_train_val.go
typhoonzero: I see that only hive will do the `DROP TABLE`, this does not affect current running cases.

typhoonzero (approved):

sql/create_train_val.go
weiguoz: It will be fixed!

PR #693, opened 2019-08-21
fix train val create in current project

typhoonzero:
Try create train and validation temporary tables in current database or MaxCompute project, because we may not have write access to the database where the training table located in.

### DIFF
--- a/sql/create_train_val.go
+++ b/sql/create_train_val.go
@@ -159,17 +159,15 @@ func namingTrainAndValDataset(origTable string) *trainAndValDataset {
 	// NOTE: We always create temporary table under current in using
 	// database or MaxCompute project, because the origin table may
 	// from another database which we may not have write access.
-	var origTableWithNoDBPrefix string
 	origTableParts := strings.Split(origTable, ".")
-	// FIXME(typhoonzero): use last part, if len != 2 should return some error
-	origTableWithNoDBPrefix = origTableParts[len(origTableParts)-1]
-	fmt.Printf("create table : %s_%s\n", tablePrefix, origTableWithNoDBPrefix)
+	// FIXME(typhoonzero): use last part, if len > 2 should return some error
+	origTableWithoutDBPrefix := origTableParts[len(origTableParts)-1]
 
 	return &trainAndValDataset{
 		database:   "sf_home",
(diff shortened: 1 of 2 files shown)


sql/create_train_val.go
```diff
@@ -156,13 +156,20 @@ func createTable(database, table, origin string, db *DB, cond string) (string, e
...
 func namingTrainAndValDataset(origTable string) *trainAndValDataset {
-	// hive returns a table with a database name
-	flattenTbl := strings.Replace(origTable, ".", "__", -1)
+	// NOTE: We always create temporary table under current in using
+	// database or MaxCompute project, because the origin table may
+	// from another database which we may not have write access.
+	var origTableWithNoDBPrefix string
+	origTableParts := strings.Split(origTable, ".")
+	// FIXME(typhoonzero): use last part, if len != 2 should return some error
+	origTableWithNoDBPrefix = origTableParts[len(origTableParts)-1]
```
weiguoz: Remove `var origTableWithNoDBPrefix string` ?
Tiny suggestion: how about renaming `origTableWithNoDBPrefix` to `origTableWithoutDBPrefix `

sql/create_train_val.go
```diff
@@ -156,13 +156,20 @@ func createTable(database, table, origin string, db *DB, cond string) (string, e
...
 
 func namingTrainAndValDataset(origTable string) *trainAndValDataset {
-	// hive returns a table with a database name
-	flattenTbl := strings.Replace(origTable, ".", "__", -1)
+	// NOTE: We always create temporary table under current in using
+	// database or MaxCompute project, because the origin table may
+	// from another database which we may not have write access.
+	var origTableWithNoDBPrefix string
+	origTableParts := strings.Split(origTable, ".")
+	// FIXME(typhoonzero): use last part, if len != 2 should return some error
```
weiguoz: I'm not sure if we can make this assertion. Is it possible the `len` equals to 1?

typhoonzero: We must support user write sqls like `select * from myproject.mytabe train ...`

weiguoz (dismissed):
LGTM.

sql/create_train_val.go
typhoonzero: Will update the comment, thanks.

weiguoz (approved):

Issue #664, opened 2019-08-14
[TODOs] ElasticDL on SQLFlow


terrytangyuan:
Initial setup and skeleton code has been added in https://github.com/sql-machine-learning/sqlflow/pull/663. Below is a list of additional work items for ElasticDL integration with SQLFlow:

# Training
- [x] Parse `SELECT ... FROM ...`, generate code to read from ODPS table and write it to [RecordIO](https://github.com/wangkuiyi/recordio) files.
- [x] Execute the above generated data conversion script before submitting `elasticdl` commands. This depends on https://github.com/wangkuiyi/elasticdl/issues/1039.
- [x] Parse `COLUMN ... LABEL ...` to figure out `<input_shape>` and `<num_classes>`.
- [x] Allow `custom_model()` to take `<input_shape>` and `<num_classes>` that will be passed to `elasticdl train` via `--params` later.
- [x] Pass additional parameters from `WITH` clause to `custom_model()`'s instantiation, such as `optimizer` and `loss`.
- [x] Construct `dataset_fn`: parse `COLUMN` clause and then pass column names, shapes, and types for features and labels to `dataset_fn`'s feature description that will be used in `tf.io.parse_single_example()`. For this MVP, column names can be obtained from `SELECT ... LABEL ...`. Each feature columns will be of shape `[1]` and of type `tf.float32` while label column is of shape `[1]` and of type `tf.int64` for classification problems and `tf.float32` for regression problems.
- [x] Skip support for feature transformation functions such as `NUMERIC` or `BUCKET` in `COLUMN` clause for now. Add some checks/warnings to handle this in codegen.
- [x] Parse `INTO` clause and pass the information to `--outputs` argument in ElasticDL high-level API.
- [x] Parse run-time configurations that can be passed to `elasticdl train` command later.
- [x] Pass generated model definition file name and other args to `elasticdl train` command.

# Prediction
- [x] Generate `PredictionOutputsProcessor` that writes prediction result to ODPS.
- [x] Parse `USING` clause to get the name to the trained model to be used to make predictions.
- [x] Pass the generated args and model definition file to `elastic predict` command.

# Others
- [x] Register allowed ElasticDL-specific keywords in `WITH` clause.
- [x] Register ElasticDL filler.
- [x] Add integration test, including the following steps: 
a. SQLFLlow generates model definition files and high-level `elasticdl train` command
[...]

PR #709, opened 2019-08-23
Add predict clause resolver

terrytangyuan:
This is a necessary change that will help the ElasticDL integration work in https://github.com/sql-machine-learning/sqlflow/issues/664. Currently predict clause uses the same resolver as train clause. It has several problems:
* It makes it difficult to tell which fields are required for prediction or training. 
* Since predict clause contains less keywords than train clause, `resolveTrainClause` would throw an error when resolving predict clause.
* It is currently not trivial to obtain additional parameters that is necessary for engine-specific parameters in predict clause via `WITH` (we have to go through `predAttrs` in `predictClause`), which is not easy for submitters to use directly.

This PR includes the following changes:

* Add `resolvedPredictClause` and `resolvePredictClause` that will be used to resolve fields from prediction clause.
* Move `trimQuotes` and renamed `resolveTrainAttribute` to `resolveAttribute` so they can be reused by both train and predict clause resolver.
* Removed `PredictCheckpointFilenameForInit` from `resolvedTrainClause` since predict now has its own resolver.

A subsequent PR will be submitted later to use this in ElasticDL prediction job submission.

### DIFF
--- a/sql/expression_resolver.go
+++ b/sql/expression_resolver.go
@@ -256,6 +256,15 @@ func getEngineSpec(attrs map[string]*attribute) engineSpec {
 	}
 }
 
+func trimQuotes(s string) string {
+	if len(s) >= 2 {
+		if s[0] == '"' && s[len(s)-1] == '"' {
+			return s[1 : len(s)-1]
+		}
+	}
+	return s
+}
+
 func resolveTrainClause(tc *trainClause) (*resolvedTrainClause, error) {
 	modelName := tc.estimator
 	preMadeModel := !strings.ContainsAny(modelName, ".")
@@ -266,14 +275,6 @@ func resolveTrainClause(tc *trainClause) (*resolvedTrainClause, error) {
 	err = ValidateAttributes(attrs)
 	if err != nil {
 		return nil, err
-	}
-	trimQuotes := func(s string) string {
-		if len(s) >= 2 {
-			if s[0] == '"' && s[len(s)-1] == '"' {
-				return s[1 : len(s)-1]
-			}
-		}
-		return s
 	}
 	getIntAttr := func(key string, defaultValue int) int {
 		if p, ok := attrs[key]; ok {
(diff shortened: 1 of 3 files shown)


sql/expression_resolver.go
```diff
@@ -70,33 +70,39 @@ type gitLabModule struct {
...
+	MaxSteps                      int
+	GradsToWait                   int
+	TensorboardLogDir             string
+	CheckpointSteps               int
+	CheckpointDir                 string
+	KeepCheckpointMax             int
+	EvalSteps                     int
+	EvalStartDelay                int
+	EvalThrottle                  int
+	EvalCheckpointFilenameForInit string
```
weiguoz, line 94: It looks this field renamed to `CheckpointFilenameForInit` in `resolvedPredictClause`.
Shall we do the same thing here?

sql/expression_resolver.go
```diff
@@ -386,33 +391,69 @@ func resolveTrainClause(tc *trainClause) (*resolvedTrainClause, error) {
...
+		return nil, err
+	}
+	getStringAttr := func(key string, defaultValue string) string {
+		if p, ok := attrs[key]; ok {
+			strVal, _ := p.Value.(string)
+			defer delete(attrs, p.FullName)
+			if err == nil {
+				return trimQuotes(strVal)
+			}
+			fmt.Printf("ignore invalid %s=%s, default is %v", key, p.Value, defaultValue)
```
weiguoz: Remove this line or use the `log.Infof(xxx)`   : )

sql/expression_resolver.go
terrytangyuan: I'd like to keep `Eval` here for now since currently training and evaluation is coupled and it would be good to keep `Eval` so that we know this is only used for evaluation. `resolvedPredictClause` is only related to prediction so it's unnecessary to name it `PredictCheckpointFilenameForInit `.

sql/expression_resolver.go
terrytangyuan: Thanks. Removed.

weiguoz (approved):
LGTM

PR #716, opened 2019-08-26
[ElasticDL] Register ElasticDL submitter in executor

terrytangyuan:
Part of #664. 

Note that ElasticDL submitter does not support `trainAndValDataset` yet so the `if` block is above the initialization code for `newTrainAndValDataset`.

Signed-off-by: terrytangyuan <terrytangyuan@gmail.com>

### DIFF
--- a/sql/executor.go
+++ b/sql/executor.go
@@ -276,6 +276,9 @@ func runExtendedSQL(slct string, db *DB, modelDir string, session *pb.Session) *
 			defer os.RemoveAll(cwd)
 
 			if pr.train {
+				if os.Getenv("SQLFLOW_submitter") == "elasticdl" {
+					return elasticDLTrain(wr, pr, db, cwd, session, nil)
+				}
 				// TODO(weiguo): fix the hard code 0.8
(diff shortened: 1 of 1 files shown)


sql/executor.go
```diff
@@ -276,6 +276,9 @@ func runExtendedSQL(slct string, db *DB, modelDir string, session *pb.Session) *
 			defer os.RemoveAll(cwd)
 
 			if pr.train {
+				if os.Getenv("SQLFLOW_submitter") == "elasticdl" {
+					return elasticDLTrain(wr, pr, db, cwd, session, nil)
+				}
```
tonyyang-svail, line 281: Does `elasticDLTrain` need the result of `newTrainAndValDataset`?

terrytangyuan: Nope not yet until ElasticDL supports reading from ODPS directly

tonyyang-svail: For the record, SQLFlow+EDL currently only supports training without evaluation.

tonyyang-svail (approved):

PR #720, opened 2019-08-27
impala reads hive session configuration from gohive

weiguoz:
TODO:  pass hive config(session, auth) to `db_generator()` in xgboost codegen


sperlingxx (approved):
LGTM

Issue #705, opened 2019-08-22
Implement feature derivation


typhoonzero:
Implement feature derivation following this doc: https://github.com/sql-machine-learning/sqlflow/blob/develop/doc/feature_derivation.md.

Should depend on IR work at: https://github.com/sql-machine-learning/sqlflow/issues/805

- [x] add `feature_derivation.go` to do the job, output a `*codegen.FeatureColumn` for all columns in the `SELECT` statement.
- [x] call `feature_derivation.go` in `ir_generator.go` to generate derevated feature columns.
- [x] Use `feature_derivator.go` in all codegen:
    - [x] tensorflow
    - [x] ~~ALPS~~
    - [x] xgboost
    - [x] ~~ElasticDL~~

tonyyang-svail:
@typhoonzero I believe the current checklist is outdated since all codegens will be depend on IR.

typhoonzero:
Done updated.

tonyyang-svail:
Done and closing.

PR #722, opened 2019-08-27
simplify expression resolver

typhoonzero:
part of work in https://github.com/sql-machine-learning/sqlflow/issues/705

### DIFF
--- a/cmd/sqlflowserver/main_test.go
+++ b/cmd/sqlflowserver/main_test.go
@@ -23,8 +23,6 @@ import (
 	"os"
 	"os/exec"
 	"path"
-	"strconv"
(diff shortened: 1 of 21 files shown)


Yancey0623 (commented):
How about creating a sub package `sql/columns` and move all the `*_column.go` files into `sql/columns`.

typhoonzero:
> How about creating a sub package sql/columns and move all the *_column.go files into sql/columns.

Will do that later, since all structs in `sql` packages is not exported, we must export things in `sql.y` to let other packages to use.

Yancey0623 (approved):
LGTM

PR #725, opened 2019-08-27
 More consistent uses of Maxcompute environment variables

terrytangyuan:
Currently we have mixed uses of `ODPS_xxx` and `MAXCOMPUTE_xxx` in the codebase. This PR addressed that. Also updated ElasticDL template so it's more consistent (related #664).


tonyyang-svail (approved):
LGTM.

PR #740, opened 2019-08-29
support multiple sql statements in one rpc call

### LINKED ISSUE #695: Support process multiple SQL statements in one RPC call
To avoid writing `%%sqlflow` in each notebook block to create a full workflow of data processing and training, we need to support writing multiple SQL statements in the notebook like:

```sql
%%sqlflow
CREATE TABLE copied_table AS SELECT * FROM origin_table where random > 0.8;
-- NOTE: this comment line is not going to run.
SELECT * FROM copied_table limit 100;
```

typhoonzero:
Fix https://github.com/sql-machine-learning/sqlflow/issues/695

Design: https://github.com/sql-machine-learning/sqlflow/pull/779


server/sqlflowserver.go
```diff
@@ -55,34 +56,60 @@ func (s *Server) Run(req *pb.Request, stream pb.SQLFlow_RunServer) error {
...
 		defer db.Close()
 	}
-	var pr *sf.PipeReader
-	if s.enableSession == true {
-		pr = s.run(req.Sql, db, s.modelDir, req.Session)
-	} else {
-		pr = s.run(req.Sql, db, s.modelDir, nil)
+
+	// FIXME(typhoonzero): split by ; can not deal with situations like
+	// "SELECT * from mytable where col <> ';';", should be fixed.
```
tonyyang-svail, line 61: To support more accurate splitting, we need the lexer in package `github.com/sql-machine-learning/sqlflow/sql`. So we need to export the lexer? If so, I would prefer splitting on the client-side...

server/sqlflowserver_test.go
```diff
@@ -119,7 +117,9 @@ func TestSQL(t *testing.T) {
 	stream, err := c.Run(ctx, &pb.Request{Sql: testErrorSQL})
 	a.NoError(err)
 	_, err = stream.Recv()
-	a.Equal(status.Error(codes.Unknown, fmt.Sprintf("run error: %v", testErrorSQL)), err)
+	a.Error(err)
+	// FIXME(typhoonzero): find out why the error changed.
```
tonyyang-svail: The cause is the `testErrorSQL` is appended with a `;`. So it won't go into the mockRun `case testErrorSQL`. Therefore no error is returned. 

https://github.com/sql-machine-learning/sqlflow/blob/229937ac9a6ed26101bff9552b269f8855abe2b7/server/sqlflowserver_test.go#L52-L53

server/sqlflowserver.go
wangkuiyi, line 61: I support to allow SQLFlow server to run a sequence of SQL statements to form a workflow. But I have a few questions:

1. I cannot find a design doc. However, I think we do need one, and it should explain the interactions between Jupyter Notebook server/magic command and the SQLFlow server.
   - What would happen if the user presses Ctrl-Enter in a Jupyter Notebook cell with one or more SQL statements? Will there be one or more gRPC call(s)?
   - What would happen if the user clicks the Run button (a triangle pointing to the right)? Will there be one or more gRPC call(s)?

- I am very worried that the splitting of SQL statements in a cell by `;` is too hacky and would lead to many bugs. I would do the right things in the right order -- get the general parser (connecting SQLFlow with various database systems) done before parsing multiple statements from a cell.

tonyyang-svail: One possible solution described at https://github.com/sql-machine-learning/sqlflow/pull/779: expose this function as `SplitSQLStatements(s string) []string` for the `server` package.

server/sqlflowserver_test.go
typhoonzero: Done

tonyyang-svail (approved):
LGTM++

PR #748, opened 2019-08-30
add boston dataset

sperlingxx:


tonyyang-svail (approved):
LGTM. It would be great if the XgBoost tutorial can use this dataset.

PR #777, opened 2019-09-04
can config server listen port

typhoonzero:
Can set server listen port using `sqlflowserver --port=1234`


Yancey0623 (approved):
LGTM, thanks for fixing the apt-get mirror.

PR #811, opened 2019-09-11
use lexer to split multiple sql

### LINKED ISSUE #787: Use lexer to split multiple SQL statements on server
As mentioned in https://github.com/sql-machine-learning/sqlflow/blob/develop/server/sqlflowserver.go#L60 we must use the lexer to split multiple SQL statements in one RPC call to make sure it's accurate.

typhoonzero:
Fix https://github.com/sql-machine-learning/sqlflow/issues/787

### DIFF
--- a/server/sqlflowserver.go
+++ b/server/sqlflowserver.go
@@ -21,7 +21,6 @@ package server
 
 import (
 	"fmt"
-	"strings"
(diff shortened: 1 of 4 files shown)


tonyyang-svail (approved):
Excellent!

PR #815, opened 2019-09-12
polish feature derivation return value

typhoonzero:
Polish feature derivation to return a `map[target][columnName]FeatureColumn`


tonyyang-svail (approved):
LGTM.

PR #819, opened 2019-09-12
[Python] Avoid passing session configuration in data generator

### LINKED ISSUE #818: [Python] Avoid session configuration in data generator
The session configuration is only used for Hive, so it shouldn't be part of the data generator API.

The driver name and the session info can be stored at the connection struct so that we can avoid passing them to data generator.
https://github.com/sql-machine-learning/sqlflow/blob/5653d093825f15938faaa9ecaf51687a9d6a96b9/sql/python/sqlflow_submitter/db.py#L47-L49

tonyyang-svail:
Fix #818


Yancey0623 (approved):
LGTM

PR #844, opened 2019-09-18
Remove head picture from /README.md

wangkuiyi:
It is not easy to create a concise illustration of the idea behind SQLFlow.

The current figure merges the concept of SQLFlow server and submitter program into a single node labeled SQLFlow. I am afraid that such merge might mislead the audience to ignore the very important distinction between SQLFlow server and the submitters.

Also, as part of the codebase, `/README.md` doesn't have to contain an illustrative figure. Such a figure, if we can create one, is alright to be on the homepage of sqlflow.org.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,10 +4,6 @@
 
 ## What is SQLFlow?
 
-<p align="center">
-    <img src="doc/figures/SQLFlow_overview.png" width="500px">
-</p>
-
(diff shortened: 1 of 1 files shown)


wangkuiyi:
Duplicated with https://github.com/sql-machine-learning/sqlflow/pull/843

PR #852, opened 2019-09-18
set default batch size to 1

typhoonzero:
The sample SQL to train iris dataset ends with low accuracy:
```
%%sqlflow
SELECT *
FROM iris.train
TRAIN DNNClassifier
WITH
  model.n_classes = 3,
  model.hidden_units = [10, 20],
  train.epoch = 100,
LABEL class
```

it seems that it's the default batch size configuration effect this.


Yancey0623 (approved):
LGTM

PR #862, opened 2019-09-19
test reduce ci docker build time

typhoonzero:
Try the tricks mentioned here: https://andrewlock.net/caching-docker-layers-on-serverless-build-hosts-with-multi-stage-builds---target,-and---cache-from/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,10 +35,8 @@ install:
 
 script:
   - set -e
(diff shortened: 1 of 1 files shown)


typhoonzero:
Seems this should work in the next build.

Yancey0623 (approved):
LGTM

Issue #805, opened 2019-09-10
[Intermediate Representation] XGBoost codegen refactor TODO list


tonyyang-svail:
As discussed in the design doc #785, we will generate the Python submitter program from the intermediate representation.
 
1. Transfer the training logic to `TrainIR`.
	1. Generate the training program from mocked `TrainIR`. #827 
		1. Create a Python database connector from the data source string. #806
	1. Generate `TrainIR` from SQL
		1. Unify feature column and column spec. #799 
		1. Make expression resolver realize type. #800 
	1. Enable SQLFlow to execute the XGB train code generated by IR. #980 
        1. Use pkg/sql/codegen/attribute in XGBoost TrainIR. #1002
        1. Add XGBoost training documentation. #1003
1. Transfer Analysis to `AnalyzeIR`. #978
1. Transfer the predict logic to `PredictIR`. #1051

typhoonzero:
After we finish above work, we should add a validator to validate attributes and generate documents for each codegen models: https://github.com/sql-machine-learning/sqlflow/issues/774

PR #865, opened 2019-09-19
auto parse attribute type in ir

typhoonzero:
Part work of https://github.com/sql-machine-learning/sqlflow/issues/805

### DIFF
--- a/sql/ir_generator.go
+++ b/sql/ir_generator.go
@@ -102,7 +102,7 @@ func parseExpression(e interface{}) (interface{}, error) {
 	if expr, ok := e.(*expr); ok {
 		if expr.typ != 0 {
 			// TODO(typhoonzero): infer the element expression type like int, float, string
-			return expr.val, nil
+			return inferStringValue(expr.val), nil
 		}
 		return parseExpression(&expr.sexp)
 	}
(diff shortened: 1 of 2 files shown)


sql/ir_generator.go
```diff
@@ -142,13 +142,30 @@ func parseExpression(e interface{}) (interface{}, error) {
 	return nil, fmt.Errorf("not supported expr: %v", el)
 }
 
+func inferStringValue(expr string) interface{} {
+	ret, err := strconv.Atoi(expr)
```
tonyyang-svail: Maybe change to more concised statements
```go
if ret, err := strconv.Atoi(expr); err == nil {
    return ret
}
```

sql/ir_generator.go
```diff
@@ -142,13 +142,30 @@ func parseExpression(e interface{}) (interface{}, error) {
...
+	if err == nil {
+		return ret
+	}
+	retFloat, err := strconv.ParseFloat(expr, 32)
+	if err == nil {
+		// NOTE: always use float32 for attributes, we may never use a
```
tonyyang-svail: `NOTE` => `NOTE(typhoonzero)`. It is better to indicate the owner of the note.

sql/ir_generator.go
```diff
@@ -142,13 +142,30 @@ func parseExpression(e interface{}) (interface{}, error) {
...
+}
+
 func parseFeatureColumn(el *exprlist) (codegen.FeatureColumn, error) {
 	head := (*el)[0].val
 	if head == "" {
 		return nil, fmt.Errorf("column description expects format like NUMERIC(key) etc, got %v", el)
 	}
 
 	switch strings.ToUpper(head) {
+	// TODO(typhoonzero): support FieldMeta configurations in column clause
```
tonyyang-svail: I don't quite understand. How do we support FieldMeta configurations in column clause?

sql/ir_generator.go
typhoonzero: `golint` seems is not checking `// NOTE` comments, so I just removed `NOTE` here.

sql/ir_generator.go
typhoonzero: Updated comment to:

```
// TODO(typhoonzero): support FieldMeta configurations (DENSE/SPARSE) in column clause, e.g.
// CATEGORY_ID(DENSE(...))
```

tonyyang-svail (dismissed):

tonyyang-svail (approved):

PR #905, opened 2019-09-23
[Intermediate Representation] parse FieldMeta

typhoonzero:
Part work of: https://github.com/sql-machine-learning/sqlflow/issues/805

Probably we should find a way to automatically document each feature column grammar.

### DIFF
--- a/sql/ir_generator.go
+++ b/sql/ir_generator.go
@@ -162,15 +162,6 @@ func parseFeatureColumn(el *exprlist) (codegen.FeatureColumn, error) {
 	}
 
 	switch strings.ToUpper(head) {
-	// TODO(typhoonzero): support FieldMeta configurations (DENSE/SPARSE) in column clause, e.g.
-	// CATEGORY_ID(DENSE(...))
-	//
-	// case dense:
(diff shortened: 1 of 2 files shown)


sql/ir_generator.go
```diff
@@ -367,6 +400,74 @@ func parseEmbeddingColumn(el *exprlist) (*codegen.EmbeddingColumn, error) {
...
+	if len(*el) >= 5 {
+		dtypeStr, err := expression2string((*el)[4])
+		if err != nil {
+			return nil, err
+		}
+		if dtypeStr == "float" {
+			dtype = codegen.Float
+		} else if dtypeStr == "int" {
+			dtype = codegen.Int
+		}
```
tonyyang-svail, line 463: Maybe add handling for unrecognized data types
```go
else {
    return nil, fmt.Errorf("bad FieldMeta data type ...")
}
```

tonyyang-svail (dismissed):
LGTM.

tonyyang-svail (approved):

sql/ir_generator.go
typhoonzero: Done

PR #912, opened 2019-09-23
Enable CI for clustering model

weiguoz:
The root cause is the default value of `model.pretrain_lr` is too large for a small dataset.
Error message:

image


shendiaomo (approved):
lgtm.

typhoonzero (approved):

PR #916, opened 2019-09-23
Avoid docker build twice on deployment

tonyyang-svail:
Avoid docker build twice on deployment. Some develop branch builds are failed due to exceeding 50 mins time limit. [ref](https://travis-ci.com/sql-machine-learning/sqlflow/builds/128681234)

After merging a pull request, TravisCI will test the latest develop branch. If the test passes, TravisCI will build a new Docker image and push it to DockerHub. However, we could reuse the Docker image built for the tests.

### DIFF
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -32,12 +32,12 @@ if [[ "$TRAVIS_BRANCH" == "develop" ]]; then
 
     echo "docker push sqlflow/sqlflow:$DOCKER_TAG"
     echo "$DOCKER_PASSWORD" | docker login --username "$DOCKER_USERNAME" --password-stdin
-    docker build -t sqlflow/sqlflow:$DOCKER_TAG -f ./Dockerfile .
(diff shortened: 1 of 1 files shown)


wangkuiyi (approved):

scripts/deploy.sh
```diff
@@ -32,12 +32,12 @@ if [[ "$TRAVIS_BRANCH" == "develop" ]]; then
 
     echo "docker push sqlflow/sqlflow:$DOCKER_TAG"
     echo "$DOCKER_PASSWORD" | docker login --username "$DOCKER_USERNAME" --password-stdin
-    docker build -t sqlflow/sqlflow:$DOCKER_TAG -f ./Dockerfile .
+    docker tag sqlflow:latest sqlflow/sqlflow:$DOCKER_TAG
```
wangkuiyi, line 35: I believe we need more comments here explaining why we don't need to build before tagging.

We also need comments explaining variable TRAVIS_ environment variables to make this configuration comprehensive.  So, I am good to merge this PR before a successive one that adds the comments.

PR #933, opened 2019-09-25
update document title format

typhoonzero:
Update title format of capitalization in all documents.
Remove `Design: ` in all design doc so that sqlflow.org's document tree will be cleaner.


weiguoz (approved):
LGTM

PR #863, opened 2019-09-19
change ir attributes to a map

typhoonzero:
Fix the issue mentioned here: https://github.com/sql-machine-learning/sqlflow/pull/837#discussion_r325839271

PR #945, opened 2019-09-27
Add back attribute check in parser

tonyyang-svail:
Related: https://github.com/sql-machine-learning/sqlflow/pull/863/files#r329254257

### DIFF
--- a/pkg/sql/sql.y
+++ b/pkg/sql/sql.y
@@ -102,6 +102,9 @@
 
 	func attrsUnion(as1, as2 attrs) attrs {
 		for k, v := range as2 {
+			if _, ok := as1[k]; ok {
+				log.Panicf("attr %q already specified", as2)
+			}
 			as1[k] = v
 		}
 		return as1
--- a/pkg/sql/parser.go
+++ b/pkg/sql/parser.go
@@ -121,12 +121,15 @@ var parseResult *extendedSelect
 
 func attrsUnion(as1, as2 attrs) attrs {
 	for k, v := range as2 {
+		if _, ok := as1[k]; ok {
+			log.Panicf("attr %q already specified", as2)
+		}
 		as1[k] = v
 	}
 	return as1
 }
 
-//line sql.y:111
+//line sql.y:114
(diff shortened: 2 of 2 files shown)


pkg/sql/sql.y
```diff
@@ -102,6 +102,9 @@
 
 	func attrsUnion(as1, as2 attrs) attrs {
 		for k, v := range as2 {
+			if _, ok := as1[k]; ok {
+				log.Panicf("attr %q already specified", as2)
```
typhoonzero, line 106: See https://github.com/golang/go/wiki/CodeReviewComments#dont-panic, do not panic, maybe just log the error and overwrite the attribute?

tonyyang-svail: I prefer using panic. We need to return the error of duplicated attributes to the user. Logging the error only shows up on the SQLFlow server.

typhoonzero: Well, `panic` will cause the server process exit, which we do not prefer because the server is a long-running process, we can not assume that user always input valid SQL statements. When the server exits, we need to manually restart it or auto restart it by Kubernetes, even it's restarted by Kubernetes, it will cause:

- temporary service down-grade
- broke currently handling incoming requests in other goroutines
- give more load to the Kubernetes master

and this kind of error better to return to the client, so maybe just put this in `ir_generator.go`?

tonyyang-svail: > Well, panic will cause the server process exit, which we do not prefer because the server is a long-running process, we can not assume that user always input valid SQL statements. 

There is a `recover` function for the parsing. So the server will not exit.

https://github.com/sql-machine-learning/sqlflow/blob/1cc07f0e3d611adec889c59773236873ee04898b/pkg/sql/parser.go#L316-L324

> this kind of error better to return to the client, so maybe just put this in ir_generator.go

Is there a way other than panic that can get the duplication error?

typhoonzero: I see, approving.

typhoonzero (approved):

PR #964, opened 2019-10-08
ir generator with feature derivation

typhoonzero:
Update work for https://github.com/sql-machine-learning/sqlflow/issues/705


pkg/sql/codegen/intermediate_representation.go
```diff
@@ -40,17 +40,18 @@ const (
...
+	Shape      []int             `json:"shape"`      // e.g. [1], [1 2 3]
+	IsSparse   bool              `json:"is_sparse"`  // e.g. false
+	Vocabulary map[string]string `json:"vocabulary"` // use a map to generate a list without duplication
 }
 
 // FeatureColumn indicates the feature column to be applied on the field. Please refer to
 // sqlflow.org/sqlflow/pkg/sql/codegen/feature_column.go for detailed list of all feature columns.
 type FeatureColumn interface {
-	isFeatureColumn()
+	GetFieldMeta() *FieldMeta
```
tonyyang-svail: Hi @typhoonzero, thanks for this PR. I am afraid `GetFieldMeta() *FieldMeta` may not cover all the cases. For example, some feature columns, like `cross`, may have multiple `FieldMeta`s.

tonyyang-svail (changes requested):

pkg/sql/codegen/intermediate_representation.go
typhoonzero: Should it be like `GetFieldMeta() []*FieldMeta` ?

tonyyang-svail: Sounds good.

typhoonzero: Done

pkg/sql/codegen/feature_column.go
```diff
@@ -66,6 +91,15 @@ type EmbeddingColumn struct {
...
 	Initializer    string
+	// only used when EMBEDDING(col_name, ...) this will set CategoryColumn = nil
+	// will fill the feature column details using feature_derivation
+	Name string
 }
 
-func (EmbeddingColumn) isFeatureColumn() {}
+// GetFieldMeta returns FieldMeta member
+func (ec *EmbeddingColumn) GetFieldMeta() []*FieldMeta {
+	if ec.CategoryColumn == nil {
```
tonyyang-svail, line 101: `nil` point is likely to cause panic.

Instead of returning `nil`, why not
- return an empty slice? 
- return `[]*FieldMeta, error`

pkg/sql/codegen/intermediate_representation.go
```diff
@@ -40,17 +40,18 @@ const (
...
-	DType     FieldType `json:"dtype"`     // e.g. "float", "int32"
-	Delimiter string    `json:"delimiter"` // e.g. ","
-	Shape     []int     `json:"shape"`     // e.g. [1], [1 2 3]
-	IsSparse  bool      `json:"is_sparse"` // e.g. false
+	Name       string            `json:"name"`       // e.g. "spetal_length"
+	DType      FieldType         `json:"dtype"`      // e.g. "float", "int32"
+	Delimiter  string            `json:"delimiter"`  // e.g. ","
+	Shape      []int             `json:"shape"`      // e.g. [1], [1 2 3]
+	IsSparse   bool              `json:"is_sparse"`  // e.g. false
+	Vocabulary map[string]string `json:"vocabulary"` // use a map to generate a list without duplication
```
tonyyang-svail, line 48: Maybe add a comment explaining what is a `Vocabulary`?

tonyyang-svail (approved):
Generally LGTM. Left a few comments for possible improvements.

pkg/sql/codegen/feature_column.go
typhoonzero: Thanks, I was considering exactly the same thing, will do in next PR.

pkg/sql/codegen/intermediate_representation.go
typhoonzero: Will do in the next PR.

PR #969, opened 2019-10-09
[ElasticDL] Enable end-to-end test for ElasticDL submitter with Maxcompute

terrytangyuan:
This enables the end-to-end test for ElasticDL submitter with Maxcompute. This also includes:
* A bunch of issues in the codegen template, `main_test.go`, and the parsing of numeric args.
* Printing out the detailed error messages if `elasticdl` command failed.

Note that there is still some work needed to check the success status of each pod created by ElasticDL, which will be done in a follow-up PR later.


pkg/sql/codegen_elasticdl.go
```diff
@@ -279,10 +285,17 @@ func elasticDLPredict(w *PipeWriter, pr *extendedSelect, db *DB, cwd string, ses
...
-	if e := cmd.Run(); e != nil {
-		return fmt.Errorf("code %v failed %v", modelDefCode, e)
+	stderr, _ := cmd.StderrPipe()
+	if err := cmd.Start(); err != nil {
+		log.Fatal(err)
+	}
+
+	scanner := bufio.NewScanner(stderr)
+	for scanner.Scan() {
+		fmt.Println(scanner.Text())
```
tonyyang-svail: Why `fmt.Println`? Shouldn't we send the logs to the user via `cw`?

terrytangyuan: Thanks. Fixed.

tonyyang-svail (approved):
LGTM

PR #983, opened 2019-10-11
add tensorflow codegen using IR

typhoonzero:
Part work of: https://github.com/sql-machine-learning/sqlflow/issues/805


pkg/sql/codegen/tensorflow/codegen.go
```diff
@@ -0,0 +1,251 @@
...
+
+// Train generates a Python program for train a TensorFlow model.
+func Train(ir codegen.TrainIR) (string, error) {
+	trainParams := make(map[string]interface{})
+	modelParams := make(map[string]interface{})
+	for attrKey, attr := range ir.Attributes {
+		if strings.HasPrefix(attrKey, "train.") {
+			trainParams[strings.Replace(attrKey, "train.", "", 1)] = attr
+		}
+		if strings.HasPrefix(attrKey, "model.") {
```
tonyyang-svail, line 136: `model.hidden_units = [10, 10]` or `hidden_units = [10, 10]`? Shall we make the model parameters consistent with XGBoost?

tonyyang-svail (approved):
As the first PR on TF over IR, it LGTM.

pkg/sql/codegen/tensorflow/codegen.go
typhoonzero: Should be `model.hidden_units`, to be consistent with all current docs: https://github.com/sql-machine-learning/sqlflow/blob/develop/README.md#quick-overview

tonyyang-svail:
TODO:

1. More codegen test coverage using `feature_derivation_case.train`.
2. EndToEnd test like `TestEnd2EndMySQLIR` in `cmd/sqlflowserver/main_test.go`.
3. Use `pkg/sql/codegen/attribute`.

pkg/sql/codegen/tensorflow/codegen.go
tonyyang-svail: I see. Let's keep the current way and discuss this in an issue.

PR #985, opened 2019-10-12
Add attribute checker

### LINKED ISSUE #774: Enhance WITH statement attributes parse for each submitter

tonyyang-svail:
Fix #774 

Next step:
1. make all IRs use attribute checker.
1. Implement `GenerateTableInMarkdown` for document generation.

### DIFF
--- a/pkg/sql/codegen/attribute/attribute.go
+++ b/pkg/sql/codegen/attribute/attribute.go
@@ -1,0 +1,101 @@
+// Copyright 2019 The SQLFlow Authors. All rights reserved.
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package attribute
+
+import "fmt"
+
+// Type indicates the attribute type of an attribute in the WITH clause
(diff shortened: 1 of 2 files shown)


pkg/sql/codegen/attribute/attribute.go
```diff
@@ -0,0 +1,101 @@
...
+	// Float indicates the corresponding attribute is a float
+	Float
+	// String indicates the corresponding attribute is a string
+	String
+	// IntList indicates the corresponding attribute is a list of integers
+	IntList
+)
+
+// Dictionary contains the description of all attributes
+type Dictionary map[string]*Description
```
typhoonzero, line 38: Maybe use a more meaningful name than `Dictionary`?

typhoonzero (approved):
LGTM

pkg/sql/codegen/attribute/attribute.go
tonyyang-svail: I was struggling with the naming as well. Any suggestions? :)

typhoonzero: Like `HyperParamsDefDict` ?

PR #1001, opened 2019-10-14
[Design] Support arbitrary select

tonyyang-svail:


doc/design/design_support_arbitrary_select.md
```diff
@@ -0,0 +1,57 @@
...
+TRAIN DNNClassifier
+WITH n_class = 3
+COLUMN c1, c2, EMBEDDING(id)
+LABEL class
+INTO my_dnn_model;
+```
+
+SQLFlow does the following steps.
+
+1. SQLFlow splits the extended SQL statement into its select clause and its train clause. In the above example, the select clause is `SELECT ... a` and the train clause is `TRAIN DNNClassifier ... INTO my_dnn_model`. For the train clause, we check the syntax by our parser at `pkg/sql/parser.go`.
```
typhoonzero, line 24: Do we need to check the arbitrary select statement using third-party parsers like calcite, or directly send the statement to the DB engine? It's easy to implement if we can directly send the SQL statement to the DB engine, and if grammar error occurs, the DB engine should return some error message before the SQL is executed.

doc/design/design_support_arbitrary_select.md
```diff
@@ -0,0 +1,57 @@
...
+
+1. SQLFlow verifies the column in the train clause.
+    1. SQLFlow executes the select clause and retrieves the column names and column types of the result. For example, the result of `SELECT ... a` has four columns with names `c1`, `c2`, `id`, and `class`. `c1`, and `c2` are of float types. And `id` and `class` are of integer types.
```
typhoonzero, line 27: Is it vital to use APIs like `describe table` to retrieve column names and types? Running the arbitrary select statement maybe resource consuming.

doc/design/design_support_arbitrary_select.md
```diff
@@ -0,0 +1,57 @@
...
+    1. SQLFlow checks the columns of train clause exist in the select result. For example, `c1`, `c2`, `id`, and `class` in `COLUMN` and `LABEL` are all in the select result. (Please be aware that `select expression` without an alias might give system-generated names that the user doesn't know in advance. In this case, we suggested using alias such as `select expression as my_column_name`.)
```
typhoonzero: A little confused about when to use the alias. If we write `SELECT c1,c2,c3 FROM table TRAIN model COLUMN c1,c2,c3`, this seems clear enough for the column clause. If we write `SELECT *`, the user may already know the column names, or just let SQLFlow to use them all.

doc/design/design_support_arbitrary_select.md
```diff
@@ -0,0 +1,57 @@
...
+    1. SQLFlow checks the columns have the desired types. The type is either suitable to explicit feature column transformation such as `EMBEDDING(id)`, or derived from [feature derivation](/doc/design/design_feature_derivation.md) such as `c1` of float type will be derived as a numerical column.
+
+1. SQLFlow generates a Python submitter program that forwards the select clause to a particular SQL engine to fetch the training data.
```
typhoonzero: fetch the training data -> fetch the training log?

doc/design/design_support_arbitrary_select.md
```diff
@@ -0,0 +1,57 @@
...
+
+Please be aware that the SQLFlow parser does not parse the nested select due to the difficulty in handling different syntax across different SQL engines. Instead, it follows the UNIX's pipeline philosophy: forwarding the complexity to various SQL engines, while retrieves the data via unified database API.
+
+## Implementation
+
+### Splitting the Extended SQL
+
+SQLFlow splits the extended SQL by looking for consecutive tokens returned by the lexer. If SQLFlow finds the following consecutive tokens [`TRAIN`, `IDENT`, `WITH`], [`PREDICT`, `IDENT`, `WITH`] or [`ANALYZE` `IDENT`, `WITH`], it splits the SQL string at the beginning of the first token in the list.
```
typhoonzero: If WITH is optional?

typhoonzero: Maybe add some examples to show how SQLFlow should use lexer to split the train clause?

doc/design/design_support_arbitrary_select.md
tonyyang-svail: I prefer directly sending the SQL statement to the DB engine.

The parser can check the syntax error, while the DB engine already does that.

doc/design/design_support_arbitrary_select.md
tonyyang-svail: Just to clarify, you are referring to create a temporary table using `CREATE TABLE {.TempTable} AS ({.SelectClause});`, then use `describe table` to retrieve column names and types. By doing so we can avoid running the arbitrary select in the later feature derivation stage and training stage.

If the user has the write permission to the DB, I do believe this method can have performance improvement. However, I would like to save these details until we find performance is the major issue.

doc/design/design_support_arbitrary_select.md
tonyyang-svail: Thanks for pointing it out. For example:

```
mysql> select class + class from iris.train limit 1;
+---------------+
| class + class |
+---------------+
|             4 |
+---------------+
```

The name `class + class` is system-generated, and the user may not know in advance.

I will add this example to the design doc.

doc/design/design_support_arbitrary_select.md
tonyyang-svail: Sorry for the confusion. I was thinking the submitter is running the training job locally, and the training job is fetching the training data, so the submitter is fetching the training data. However, the submitter is likely to run the training job remotely.

I will remove ~~"to fetch the training data"~~.

doc/design/design_support_arbitrary_select.md
tonyyang-svail: > Maybe add some examples to show how SQLFlow should use lexer to split the train clause?

Thanks for the suggestion. I will add an example.

> If WITH is optional?

Currently, our parser still requires the `WITH` clause. The design doc illustrates the main idea. And we should reference the code for the specific implementation. I am changing 

If SQLFlow finds the following consecutive tokens ...

=>

If SQLFlow finds the consecutive tokens like ...

doc/design/design_support_arbitrary_select.md
typhoonzero: I see, that's cool.

doc/design/design_support_arbitrary_select.md
typhoonzero: > However, I would like to save these details until we find performance is the major issue.

Sounds good.

typhoonzero (approved):
LGTM++

PR #1009, opened 2019-10-15
Repl supports `-e -f` options

### LINKED ISSUE #993: repl supports -e -f
A clear description of the problem

sqlflow -e "sql"
sqlflow -f filename

A clear description of any alternative solutions or features

**Additional Notes**

shendiaomo:
Fix #993 .
- Handle EOF correctly to make redirection work correctly.
- `-f` is  implemented in the same way as redirection, i. e. `repl -f test.sql` is the same as `repl < test.sql` or  `cat test.sql | repl`. This behavior is the same as `hive -f test.sql`.


cmd/repl/repl.go
```diff
@@ -119,6 +136,31 @@ func main() {
...
+		sqlRun(*cliStmt)
+	} else {
+		sqlFile := os.Stdin
+		if flagPassed("file", "f") {
+			sqlFile, err = os.Open(*sqlFileName)
+			if err != nil {
+				log.Fatal(err)
+			}
+		}
+		defer sqlFile.Close()
```
weiguoz, line 164: I don't think we need to `Close()` the `os.Stdin`.

shendiaomo: Done.

cmd/repl/repl.go
```diff
@@ -81,9 +83,25 @@ func render(rsp interface{}, table *tablewriter.Table) bool {
...
+			}
+		})
+	}
+	return found
+}
+
 func main() {
 	ds := flag.String("datasource", "", "database connect string")
 	modelDir := flag.String("model_dir", "", "model would be saved on the local dir, otherwise upload to the table.")
+	cliStmt := flag.String("execute", "", "execute SQLFlow from command line.  e.g. --execute 'select * from table1'")
```
tonyyang-svail, line 139: FYI, Go's default "flag" package doesn't differentiate between `-` and `--`, so `-execute` is the same as `--execute`.

cmd/repl/repl.go
```diff
@@ -97,15 +115,15 @@ func main() {
 		}
 	}
 
-	for {
-		fmt.Print("sqlflow> ")
-		slct := readStmt()
-		fmt.Println("")
-
+	isTerminal := !flagPassed("execute", "e", "file", "f") && terminal.IsTerminal(syscall.Stdin)
```
tonyyang-svail, line 156: ~~I am not sure if we need `terminal.IsTerminal()`. If the user doesn't type in `-e/-f`, we should start the program as a REPL.~~ Never mind, I see the code also supports reading from stdin.

cmd/repl/repl.go
```diff
@@ -97,15 +115,15 @@ func main() {
...
+	sqlRun := func(stmt string) {
+		if !isTerminal {
+			fmt.Println("sqlflow>", stmt)
```
tonyyang-svail: I am not sure why we need to print `sqlflow>` in a non-terminal mode such as `go run repl.go -e "select 1;"`.

tonyyang-svail (commented):
Hi @shendiaomo, Thanks for submitting this PR.

(1) According to the description in #993, looks like we still need to implement

> The log information is output to the standard error stream, and the result information is output to the standard output stream

So please avoid `fmt.Println` in this package, for example, there are some `fmt.Println` in `render()` fucntion.

(2) Since we are supporting reading SQL statements from a file, and users tend to put multiple SQL statements in one file. Shall we add this support to REPL?

(3) This package was intended for debugging for SQLFlow developers. If we are to use it as an official REPL, we should guard it with unit tests.

cmd/repl/repl.go
```diff
@@ -97,15 +115,15 @@ func main() {
 		}
 	}
 
-	for {
-		fmt.Print("sqlflow> ")
-		slct := readStmt()
-		fmt.Println("")
-
+	isTerminal := !flagPassed("execute", "e", "file", "f") && terminal.IsTerminal(syscall.Stdin)
+	sqlRun := func(stmt string) {
```
wangkuiyi: I propose to separate `sqlRun` out from `main` to be a function definition as the following

```go
func runStmt(stmt string, isTerminal bool)
```

so to make `main` shorter and more comprehensive.

cmd/repl/repl.go
```diff
@@ -119,6 +137,31 @@ func main() {
...
+		if flagPassed("file", "f") {
+			sqlFile, err = os.Open(*sqlFileName)
+			if err != nil {
+				log.Fatal(err)
+			}
+			defer sqlFile.Close()
+		}
+		scanner := bufio.NewScanner(sqlFile)
+
+		for {
```
wangkuiyi: I propose to extract this `for` loop out from `main` to be a function definition

```go
def repl(inputScanner bufio.Scanner, isTerminal bool)
```

wangkuiyi (commented):
It looks to me that users could pass in multiple SQL statements from the `-e` flag, just like they could do with `-f`. So, it makes sense for us to run the REPL anyway -- with `-e`, with `-f`, and with stdin.

- With `-e`, the input is a [`strings.Reader`](https://golang.org/pkg/strings/#Reader).
- With `-f`, the input is a [`os.File`](https://golang.org/pkg/os/#File).
- With stdin, the input is `os.Stdin`.

cmd/repl/repl.go
shendiaomo: Yep. The reason of introducing `-execute` and `-file` is to be consistent with `-model_dir` and `-datasource`, while `-e` and `-f` is to be compatible with hive.

cmd/repl/repl.go
shendiaomo: In my own experience, user may use redirection to save information in the training process,  such as `go run repl.go -f 'train_and_evaluate.sql' **> train.log 2>&1**`, `sqlflow> ` can be used as an indicator for different output blocks.

cmd/repl/repl.go
shendiaomo: Done.

cmd/repl/repl.go
shendiaomo: Done.

wangkuiyi (approved):

wangkuiyi:
Thanks to @shendiaomo for this contribution.

it would be great to have a follow up PR to add unit tests.

Issue #992, opened 2019-10-14
Tensorflow codegen using IR


typhoonzero:
Related: https://github.com/sql-machine-learning/sqlflow/issues/805

1. [x] `ir_generator.go` generate `PredictionIR` for codegen to use.
1. [x] EndToEnd test like TestEnd2EndMySQLIR in cmd/sqlflowserver/main_test.go.
1. [x] More codegen test coverage using feature_derivation_case.train.
1. [x] Use pkg/sql/codegen/attribute.

typhoonzero:
Done and closing.

PR #1018, opened 2019-10-16
generate predict IR

typhoonzero:
Part work in: https://github.com/sql-machine-learning/sqlflow/issues/992


tonyyang-svail (approved):
LGTM. I am wondering what is breaking on the Hive tests?

typhoonzero:
Seems `iris.train` table is not created for the test, will find a way to test this for Hive.

PR #1022, opened 2019-10-17
config hive_location and hdfs_namenode_addr from session

### LINKED ISSUE #1007: Default database in SQLFLOW_DATASOURCE
**Description**
Even has defined environment parameter `SQLFLOW_DATASOURCE` at client side, the `TRAIN` clause cannot fetch access to right database when creating temporary table for train and validation dataset.

The same problem also occurred at `predict` clause.

**Screenshots**
46294C2B-9CC8-4B39-B326-56FF30A450C0

typhoonzero:
Fix https://github.com/sql-machine-learning/sqlflow/issues/1007


pkg/server/proto/sqlflow.proto
```diff
@@ -24,6 +24,9 @@ message Session {
     string db_conn_str = 2;
     bool exit_on_submit = 3;
     string user_id = 4;
+    // for loading CSV to hive
+    string hive_location = 5;
+    string hdfs_namenode_addr = 6;
```
tonyyang-svail, line 29: Maybe group these to a message called `HiveConfig`?

tonyyang-svail (approved):
LGTM.

pkg/server/proto/sqlflow.proto
typhoonzero: hdfs_namenode_addr is actually an HDFS config not for only hive.

PR #1042, opened 2019-10-20
Add the model definition and model zoo design

wangkuiyi:


doc/design/model_zoo.md
```diff
@@ -0,0 +1,104 @@
...
+1. A **model definition**, for example, `MyDNNRegressor`, is a Python class or other kinds of code that specify a model.
+
+1. The SQL data to model inputs conversion, or **data conversion**, is the COLUMN and optionally LABEL clauses.
+
+1. Hyperparameters
+   1. **Model hyperparameters** are some key-value pairs specified in the WITH clause. They are arguments to the constructor of the model definition.
+   1. **Training hyperparameters** appear in WITH, as model constructors do. They affect the training process.
+   1. **Data conversion hyperparameters** appear in COLUMN and optionally LABEL. For example, the scale range and bucket size in the first example.
+
+1. A **trained model** consists of all the above concepts and the estimated model parameters.
```
typhoonzero, line 51: Should also include model parameters ( weights ).

typhoonzero: We should make clear what the identifier is used for follows `TRAIN`, `USING`, `INTO`:

`TRAIN [IDENT]`: IDENT is a model def.
`USING [IDENT]`: IDENT is the location stores a model parameter set.
`INTO [IDENT]`: IDENT is the location stores a model parameter set

doc/design/model_zoo.md
```diff
@@ -0,0 +1,104 @@
...
+1. To describe dependencies, we require a Dockerfile at the root directory of the repository.
+1. To release a repository, we checkout the specific version and run `docker run` with the Dockerfile.
+1. Each release has a globally unique ID (GUID) represented by the hash ID of the Docker image.
+1. The Docker image must support a protocol of command-line arguments so that SQLFlow can pass in required information like hyperparameters from the `docker run` command.
+
+### Argument Protocol
+
+For a SELECT program using the SQLFlow syntax extension, the SQLFlow server converts it into a *submitter* program, usually in Python.
+
+Consider the training statement at the beginning of this document. Suppose that the Python class `MyDNNRegressor` is in source file `my_dnn_regressor.py` in the repository https://github.com/a_data_scientist/regressors, which builds into Docker image `a_data_scientist/regressors`.  Let us also assume that the submitter program is `/var/sqlflow/submitters/an_analyst/my_first_model.py`.  SQLFlow can run the submitter using the Docker image as follows:
```
typhoonzero, line 71: How can `MyDNNRegressor` distinguish it's a model in the model zoo or a pre-made estimator or some self-defined model in `PYTHONPATH`? And how to convert the name `MyDNNRegressor ` to docker image `a_data_scientist/regressors`? We can not assume all the model name is in the model zoo, we need to run the model before it's published to some docker repository.

doc/design/model_zoo.md
```diff
@@ -0,0 +1,104 @@
...
+
+```bash
+docker run --rm -it \
+  -v /var/sqlflow/submitters:/submitters \
+  a_data_scientist/regressors \
+    python /submitters/an_analyst/my_first_model.py
```
typhoonzero, line 77: For distributed cases, the command should be calling ElasticDL. Also how to put generated `my_first_model.py` to the runtime containers.

doc/design/model_zoo.md
```diff
@@ -0,0 +1,104 @@
...
+If we run SQLFlow in a Docker container, to allow it to run another Docker container, we must enable the Docker-in-Docker feature, say, following [this blog post](https://itnext.io/docker-in-docker-521958d34efd).  Suppose that the SQLFlow server runs in a Kubernetes cluster, where we cannot enable Docker-in-Docker. In such a case, we can make the SQLFlow server call Kubernetes API to run a Pod that executes the above `docker run` command line.
+
+ There could be more than one model definition repositories and Docker images. We need to be able to choose the right one that contains the model `MyDNNRegressor`.  We might want to require that the Docker image, when passed the specific argument `list`, prints the model definition names it contains.  For example, the following command prints a name in the form of `a_data_scientist.MyDNNRegressor`.
+
+```bash
+docker run --rm -it a_data_scientist/regressors list
+```
+
+The submitter program `my_first_model.py`, running in the model definition Docker container, should be able to submit a (distributed) training job to a preconfigured cluster.  Once the job completes, the submitter program adds, or edits, a row in a **model zoo table**.
+
```
typhoonzero: Use a database to store the saved model is:

1. hard to publish or share the weights to other users.
1. users are using different databases, it can not be commonly used for all open source users.
1. for some database engines it's not suitable to store model parameters in the database like hive. Most of the saved models are of "small size" for HDFS

doc/design/model_zoo.md
wangkuiyi: I think we don't need to enable the use of pre-made Estimators or self-defined models in PYTHONPATH. How about disabling these possibilities?

wangkuiyi: I propose the solution as described in https://github.com/sql-machine-learning/sqlflow/pull/1042/files#diff-0e016e244fea1579c23458c190308b87R82 for Docker images to list the model definitions it contains.

doc/design/model_zoo.md
wangkuiyi: In my mind, in order to submit an ElasticDL training job, we need the following actions.

1. The SQLFlow server invokes codegen_elasticdl.go to generates the submitter program `my_first_model.py`.
1. `my_first_model.py` calls the ElasticDL client API to submit a job.
1. The SQLFlow server then runs `my_first_model.py` in a Docker container (if SQLFlow service is not running in a Kubernetes cluster), or a Pod, which has ElasticDL client library installed.
1. As this container needs to contain ElasticDL and the model definition, it must be derived from `a_data_scientist/regressor`, like from the following Dockerfile:

   ```dockerfile
   FROM a_data_scientist/regressors
   RUN pip install -y elasticdl_api
   ```

Please correct me or improve the above idea.

doc/design/model_zoo.md
```diff
@@ -0,0 +1,137 @@
...
+Such design flaw roots from the ignoring of the fact that model definitions are code that has versions. Another important aspect is that we often build the source code into some release-ready form.  Once we noticed these facts, we can use version management tools and Git and release engineering tools like Docker.  Here follows our proposal.
+
+1. A collection of model definitions is a Git repository of source files. 
+1. To describe dependencies, we require a Dockerfile at the root directory of the repository.
+1. To release a repository, we checkout the specific version and run `docker run` with the Dockerfile.
+1. Each release has a globally unique ID (GUID) represented by the hash ID of the Docker image.
+1. The Docker image must support a protocol of command-line arguments so that SQLFlow can pass in required information like hyperparameters from the `docker run` command.
```
tonyyang-svail, line 65: ~~SQLFlow passes the argument to `docker run` and starts the training. And the `docker run` is supported by the model repo, so the model repo is responsible for the distributed training?~~

doc/design/model_zoo.md
typhoonzero: Thanks, I see. Another question, we can release one or more docker images, where can we find all docker images built for SQLFlow models? Maybe we can use a [docker registry](https://docs.docker.com/registry/deploying/)

doc/design/model_zoo.md
typhoonzero: I see, that's clear enough. One question, maybe we can just build all model zoo Docker images with ElasticDL API ( model Dockerfile should depend on our base image like `sqlflow/sqlflow_model_base`), then the submitted job can directly use the model's Docker image, or else it may require to build an image and push to the registry before submitting the job.

typhoonzero (dismissed):

doc/design/model_zoo.md
```diff
@@ -0,0 +1,137 @@
...
+```
+
+### Model Publication
+
+Publishing a model definition, we hope all users of all SQLFlow deployments can train the model using their datasets. To achieve this goal, data scientists need to set up a continuous deployment system to automatically build their model definition repositories into Docker images and push to a public Docker registry like DockerHub.com.  In the above example, the Docker image `a_data_scientist/regressors` is one hosted on DockerHub.com.
+
+It requires more steps to publish a trained model. We need a public registry like DockerHub.com, but hosts rows from the model zoo table and corresponding model parameters.  We propose the following extended SQL syntax for analysts `an_analyst` to publish her trained model.
+
+```sql
+SQLFLOW PUBLISH my_first_model
```
typhoonzero, line 164: Can we also use the syntax like `SQLFLOW LIST` to list all models available and `SQLFLOW DESCRIBE my_first_model` to show the model's document?

doc/design/model_zoo.md
wangkuiyi: I see. It is like requiring all model definition repositories have their Dockefile's all derived from a common base image. I think it is a viable solution.

doc/design/model_zoo.md
wangkuiyi: I think the model zoo database is part of the SQLFlow server. It could be embedding DBs like SQLite.

doc/design/model_zoo.md
wangkuiyi: Good idea! I might not have sufficient time today to follow your suggestion in this PR. Please feel free to merge this PR and submit successive PRs to complete these ideas.

doc/design/model_zoo.md
```diff
@@ -0,0 +1,137 @@
...
+
+Such design flaw roots from the ignoring of the fact that model definitions are code that has versions. Another important aspect is that we often build the source code into some release-ready form.  Once we noticed these facts, we can use version management tools and Git and release engineering tools like Docker.  Here follows our proposal.
```
mhaoli, line 59: `tools and Git` -> `tools like Git`?

QiJune:
There is a project called [dvc](https://dvc.org), which is an Open-source Version Control System for Machine Learning Projects. [Here](https://github.com/iterative/dvc) is the github repo. 

It claims to provide the following three features:

- Version control of machine learning models, data sets and intermediate files.
- ML experiment management, along with first-class citizen metrics and ML pipelines. 
- Deployment & Collaboration

Maybe we could learn something from it.

doc/design/model_zoo.md
```diff
@@ -0,0 +1,174 @@
...
+
+Documatation for my_first_model.MyDNNRegressor
+...
+...
+```
+
+Use a published model to predict some data:
+
+```sql
+SELECT ... TO PREDICT employee.predicted_salary USING model.sqlflow.com/an_analyst/my_first_model
```
QiJune, line 154: Do we need a `docker pull` or `git pull` like command here, to pull the model in the model zoo to local?

doc/design/model_zoo.md
```diff
@@ -0,0 +1,174 @@
...
+
+1. A **data source** is a table, for example, `employee`.  Please be aware that we can catch and reuse a dataset until its data sources are changed.
+
+1. A **model definition**, for example, `MyDNNRegressor`, is a Python class or other kinds of code that specify a model.
+
+1. The SQL data to model inputs conversion, or **data conversion**, is the COLUMN and optionally LABEL clauses.
+
+1. Hyperparameters
+   1. **Model hyperparameters** are some key-value pairs specified in the WITH clause. They are arguments to the constructor of the model definition.
+   1. **Training hyperparameters** appear in WITH, as model constructors do. They affect the training process.
```
terrytangyuan, line 48: Training might be mixed with evaluation sometimes too. We should probably also store evaluation hyperparameters. They can affect the resulting trained model (in case techniques like early-stopping is used).

doc/design/model_zoo.md
```diff
@@ -0,0 +1,174 @@
...
+
+The training submitter program `my_first_model.py`, running in the model definition Docker container, should be able to submit a (distributed) training job to a preconfigured cluster.  Once the job completes, the submitter program adds, or edits, a row in a **model zoo table**.
+
+### Model Zoo Data Schema
+
+The model zoo table is in a database deployed as part of the SQLFlow service. This database might not be the one that holds data sources.  The only requirement of the model zoo table is to have a particular data schema that contains at least the following fields.
+
+1. model ID (key), specified by the INTO clause, or `an_analyst/my_first_model` in the above example.
+1. Docker image ID, the Docker commit ID of the image `a_data_scientist/regressors` in the above example.
```
terrytangyuan, line 89: Also need the docker repository location where this particular docker image resides

doc/design/model_zoo.md
```diff
@@ -0,0 +1,174 @@
...
+| MyDNNClassifier      |
+
+SQLFLOW DESCRIBE model.sqlflow.com/an_analyst/my_first_model.MyDNNRegressor;
+
+Documatation for my_first_model.MyDNNRegressor
+...
+...
+```
+
+Use a published model to predict some data:
```
terrytangyuan, line 151: predict some data -> make some predictions on new data

doc/design/model_zoo.md
```diff
@@ -0,0 +1,174 @@
...
+
+If we run SQLFlow in a Docker container, to allow it to run another Docker container, we must enable the Docker-in-Docker feature, say, following [this blog post](https://itnext.io/docker-in-docker-521958d34efd).  Suppose that the SQLFlow server runs in a Kubernetes cluster, where we cannot enable Docker-in-Docker. In such a case, we can make the SQLFlow server call Kubernetes API to run a Pod that executes the above `docker run` command line.
+
+The training submitter program `my_first_model.py`, running in the model definition Docker container, should be able to submit a (distributed) training job to a preconfigured cluster.  Once the job completes, the submitter program adds, or edits, a row in a **model zoo table**.
+
+### Model Zoo Data Schema
+
+The model zoo table is in a database deployed as part of the SQLFlow service. This database might not be the one that holds data sources.  The only requirement of the model zoo table is to have a particular data schema that contains at least the following fields.
+
+1. model ID (key), specified by the INTO clause, or `an_analyst/my_first_model` in the above example.
```
weiguoz, line 88: I would suggest separating the `an_analyst/my_first_model` into 2 fields(`an_analyst ` and `my_first_model`) to support  SQL like `SQLFLOW LIST an_analyst` easier.

doc/design/model_zoo.md
```diff
@@ -0,0 +1,174 @@
...
+Within a deployment, it is straightforward to share a trained model.  If the analyst, `an_analyst`, in the above example wants to use her own trained model `my_first_model` for prediction, she could use the short name `my_first_model`.
+
+```sql
+SELECT ... TO PREDICT ... USING my_first_model
+```
+
+If another analyst wants to use the trained model, he would need to use the full name as `an_analyst/my_first_model`.
+
+```sql
+SELECT ... TO PREDICT ... USING an_analyst/my_first_model
```
shendiaomo, line 111: I think using dot rather than slash  is more nature in SQL, such as `an_analyst.my_first_model`.
If `an_analyst/my_first_model` is a path, maybe a quoted string is more preferable, a quoted string implies that the slash is not an operator, i.e. `an_analyst/    my_first_model` is different to `an_analyst  /my_first_model`, but `an_analyst.    my_first_model` and `an_analyst  .my_first_model` has the same meaning.

doc/design/model_zoo.md
```diff
@@ -0,0 +1,174 @@
...
+The training submitter program `my_first_model.py`, running in the model definition Docker container, should be able to submit a (distributed) training job to a preconfigured cluster.  Once the job completes, the submitter program adds, or edits, a row in a **model zoo table**.
+
+### Model Zoo Data Schema
+
+The model zoo table is in a database deployed as part of the SQLFlow service. This database might not be the one that holds data sources.  The only requirement of the model zoo table is to have a particular data schema that contains at least the following fields.
+
+1. model ID (key), specified by the INTO clause, or `an_analyst/my_first_model` in the above example.
+1. Docker image ID, the Docker commit ID of the image `a_data_scientist/regressors` in the above example.
+1. submitter program, the source code of the submitter program, `my_first_model.py` in the above example, or its MD5 hash.
+1. data converter, the COLUMN and LABEL clauses.
```
shendiaomo, line 91: It seems we should save the extended SQL itself instead of the two above fields, for example: if the user trained two different datasets at different time (daily analysis e. g.), the model zoo should know which model comes from which dataset. 
As a supplement, the SQLFlow version and training date/time can be present too, just like the build number in software versioning.
At last, there should be a description field that records training result (test accuracy etc.) by default, or users cannot distingish different models if they have too many.

doc/design/model_zoo.md
```diff
@@ -0,0 +1,174 @@
...
+
+There could be more than one model definitions in each model's Docker image. We need to be able to find out which model definition is used to train current saved model. Also if we only want to reuse the model definition to train a new model we need to know the model definition class name, so that we can pass it to the `TRAIN` clause.
+
+To list all trained models of one user, you can do:
+
+```sql
+SQLFLOW LIST an_analyst
+```
+
+This should output a table showing the saved models and which model definition it was using.
```
shendiaomo, line 122: Is this table the same as the model zoo table mentioned above in line #86? It seems they have different schemas.

doc/design/model_zoo.md
```diff
@@ -0,0 +1,174 @@
...
+1. To describe dependencies, we require a Dockerfile at the root directory of the repository.
+1. To release a repository, we checkout the specific version and run `docker run` with the Dockerfile.
+1. Each release has a globally unique ID (GUID) represented by the hash ID of the Docker image.
+1. The Docker image must support a protocol of command-line arguments so that SQLFlow can pass in required information like hyperparameters from the `docker run` command.
+
+### Submitter Programs
+
+For a SELECT program using the SQLFlow syntax extension, the SQLFlow server converts it into a *submitter* program, usually in Python.
+
+Consider the training statement at the beginning of this document. Suppose that the Python class `MyDNNRegressor` is in source file `my_dnn_regressor.py` in the repository https://github.com/a_data_scientist/regressors, which builds into Docker image `a_data_scientist/regressors`.  Let us also assume that the submitter program is `/var/sqlflow/submitters/an_analyst/my_first_model.py`.  SQLFlow can run the submitter using the Docker image as follows:
```
tonyyang-svail, line 71: Please be clear `/var/sqlflow/submitters/an_analyst/my_first_model.py` is generated by SQLFlow. I was wondering what is `my_first_model.py` until I read the conversation at 

> The SQLFlow server invokes codegen_elasticdl.go to generates the submitter program my_first_model.py.

doc/design/model_zoo.md
tonyyang-svail: I also had the same question as @typhoonzero while I am reading the design. Please add "the following actions" to the design doc.

doc/design/model_zoo.md
wangkuiyi: The model zoo is a collection of Docker images. When the SQLFlow server runs `docker run`, the Docker client would pull if needed.

doc/design/model_zoo.md
wangkuiyi: Yes. For simplicity, I just mentioned all hyperparameters related to training by "training hyperparameters".

tonyyang-svail (approved):
Generally LGTM. @wangkuiyi Please also take a look like other's comments. :)

doc/design/model_zoo.md
tonyyang-svail: No, SQLFlow will generate a submitter program according to
1. the model.py provided by the model repo
1. the `input_fn` generated according to `COLUMN` clause.

doc/design/model_zoo.md
wangkuiyi: I see. Yes. Just like the full name of a trained model like `model.sqlflow.com/an_analyst/my_first_model`, we need `dockerhub.com/a_data_scientist/regressors` here to name a Docker image.

doc/design/model_zoo.md
wangkuiyi: Do you mean making `SQLFLOW LIST a_data_scientist` to list all model definitions in all Docker images built from code repositories owned by `a_data_scientist`?

doc/design/model_zoo.md
wangkuiyi: Both the full name of Docker images and trained models like the following are inspired by the convention used by Docker to name images.

- `dockerhub.com/a_data_scientist/regressors` or `a_data_scientist/regressors` names the Docker image hosted on `dockerhub.com`. This convention is from the Docker community.

- `models.sqlflow.com/an_analyst/my_first_model` or `an_analyst/my_first_model` names a model trained by `an_analyst` and hosted on `models.sqlflow.com`. I prefer to use the same convention of Docker image names, rather than inventing a new one like `an_analyst.my_first_model`.

doc/design/model_zoo.md
wangkuiyi: I agree that we might need more fields here. Each exists for some reason. We could add these fields together with the reasons later.

doc/design/model_zoo.md
wangkuiyi: I don't see relation to #86.

doc/design/model_zoo.md
wangkuiyi: I agree that the identifier after `TRAIN` is the name of a model definition. However, the identifiers after `USING` and `INTO` are not the local of model parameters; instead, it is the value in the `model ID` field of the model zoo table, as described later in this document.

doc/design/model_zoo.md
```diff
@@ -0,0 +1,174 @@
...
+SQLFLOW PUBLISH my_first_model
+    [TO https://models.sqlflow.com/user_name]
+```
+
+This statement uploads the model parameters and other information to the registry service, which defaults to https://models.sqlflow.com, and under the account `user_name`, which defaults to `an_analyst` in the above example.
+
+Then, another analyst should be able to use the trained model by referring to it in its full name.
+
+```sql
+SELECT ... TO PREDICT employee.predicted_salary USING model.sqlflow.com/an_analyst/my_first_model
```
ywskycn, line 173: We should use `sqlflow.org`?

wangkuiyi: I am afraid that the model service would be a commercialized one; as we need fundings for storage and computation resource.

wangkuiyi:
I merged this PR as I will have to rush into other development work and will not have sufficient time to fix all comments here. I remember that are a few comments that would lead to fixes. Please commenters feel free to file PRs to fix them. Thanks!

doc/design/model_zoo.md
```diff
@@ -0,0 +1,137 @@
...
+
+Users don't have to write the `COLUMN` clause in the above example, as SQLFlow reuses the one in the training statement.
+
+## Concepts
+
+The above training and prediction example reveals some concepts related to models.
+
+1. A **dataset** is defined and generated by a SQL SELECT statement. For example, `SELECT * FROM employee WHERE onboard_year < 2019`.
+
+1. A **data source** is a table, for example, `employee`.  Please be aware that we can catch and reuse a dataset until its data sources are changed.
```
byzhang, line 40: s/catch/cache

doc/design/model_zoo.md
```diff
@@ -0,0 +1,137 @@
...
+
+1. A **model definition**, for example, `MyDNNRegressor`, is a Python class or other kinds of code that specify a model.
+
+1. The SQL data to model inputs conversion, or **data conversion**, is the COLUMN and optionally LABEL clauses.
```
byzhang, line 44: data conversion or "feature transformation"?

doc/design/model_zoo.md
byzhang: > In my mind, in order to submit an ElasticDL training job, we need the following actions.
> 
> 1. The SQLFlow server invokes codegen_elasticdl.go to generates the submitter program `my_first_model.py`.
> 2. `my_first_model.py` calls the ElasticDL client API to submit a job.
> 3. The SQLFlow server then runs `my_first_model.py` in a Docker container (if SQLFlow service is not running in a Kubernetes cluster), or a Pod, which has ElasticDL client library installed.
> 4. As this container needs to contain ElasticDL and the model definition, it must be derived from `a_data_scientist/regressor`, like from the following Dockerfile:
>    ```dockerfile
>    FROM a_data_scientist/regressors
>    RUN pip install -y elasticdl_api
>    ```
> 
> Please correct me or improve the above idea.

Are you talking about submitting a model to the zoo, or submitting a training job? A little bit lost in the discussion.

doc/design/model_zoo.md
```diff
@@ -0,0 +1,137 @@
...
+
+ There could be more than one model definition repositories and Docker images. We need to be able to choose the right one that contains the model `MyDNNRegressor`.  We might want to require that the Docker image, when passed the specific argument `list`, prints the model definition names it contains.  For example, the following command prints a name in the form of `a_data_scientist.MyDNNRegressor`.
+
+```bash
+docker run --rm -it a_data_scientist/regressors list
+```
+
+The training submitter program `my_first_model.py`, running in the model definition Docker container, should be able to submit a (distributed) training job to a preconfigured cluster.  Once the job completes, the submitter program adds, or edits, a row in a **model zoo table**.
+
```
byzhang, line 83: Is `my_first_model.py` written by the developer or generated by the system? Does it contain just the model definition, or also include the data pipeline?

doc/design/model_zoo.md
byzhang, line 65: Can we support the tag similar with docker images?

doc/design/model_zoo.md
typhoonzero: We should probably save both the training SQL statement as well as the "data converter", because we are doing "feature derivation" when training, the "data converter" is generated based on the training data. If the data used for predict is of different format of training data, go through "feature derivation" again might cause some error.

doc/design/model_zoo.md
typhoonzero: @tonyyang-svail does that mean we still need to generate some python source code and run it in the docker container?

typhoonzero: We will delete this line because we all need to run `codegen` to generate a python program containing all hyperparameters to submit the job.

PR #1044, opened 2019-10-21
[Intermediate Representation] Add TF end to end tests

typhoonzero:
part work of https://github.com/sql-machine-learning/sqlflow/issues/992

### DIFF
--- a/pkg/sql/codegen/tensorflow/template_pred.go
+++ b/pkg/sql/codegen/tensorflow/template_pred.go
@@ -30,43 +30,12 @@ type predFiller struct {
 }
 
 const tfPredTemplateText = `
-import os
-# Disable Tensorflow INFO and WARNING logs
-os.environ['TF_CPP_MIN_LOG_LEVEL'] = '3'
-
-import sys, json
-import tensorflow as tf
-import functools
-try:
-    import sqlflow_models
-except:
-    pass
-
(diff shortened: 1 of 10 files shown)


pkg/sql/codegen/tensorflow/template_pred.go
```diff
@@ -78,148 +47,29 @@ feature_metas["{{$value.Name}}"] = {
...
+model_params["{{$k}}"]={{$v | attrToPythonValue}}
 {{end}}
 
-print("Done predicting. Predict table : {{.ResultTable}}")
+pred(is_keras_model="{{.IsKerasModel}}" == "true",
+    datasource="{{.DataSource}}",
+    estimator="{{.Estimator}}",
+    select="""{{.Select}}""",
+    result_table="{{.ResultTable}}",
+    feature_column_code="""{{.FeatureColumnCode}}""",
```
tonyyang-svail, line 68: Is it possible to avoid `feature_column_code` and `estimator_code`? So that we don't need to call `eval`.

tonyyang-svail (approved):

pkg/sql/codegen/tensorflow/template_pred.go
typhoonzero: Sounds good, will try to refine this in next PR.

PR #1050, opened 2019-10-22
support pass hdfs user and pass

### LINKED ISSUE #1048: Send user hadoop credentials using session
When running SQLFlow predict statements with Hive, SQLFlow intend to write the prediction result to a temporary CSV file first and then use the `hdfs` command to upload the CSV file to HDFS, then load the CSV file into the result Hive table.

The directory where the temporary CSV file will write to can be configured by the client-side environment variable `SQLFLOW_HIVE_LOCATION`, and the HDFS name node address can be configured using `SQLFLOW_HDFS_NAMENODE_ADDR`.

However in some deployment, the `hdfs` command needs user credential environment variables (e.g. `HDFS_USERNAME`, `HDFS_PASSWD`) to proceed, we still need to pass the user's HDFS credentials to the server.

typhoonzero:
Fix: https://github.com/sql-machine-learning/sqlflow/issues/1048

### DIFF
--- a/pkg/server/proto/sqlflow.proto
+++ b/pkg/server/proto/sqlflow.proto
@@ -27,6 +27,8 @@ message Session {
     // for loading CSV to hive
     string hive_location = 5;
     string hdfs_namenode_addr = 6;
+    string hdfs_user = 7;
+    string hdfs_pass = 8;
 }
 
 // SQL statements to run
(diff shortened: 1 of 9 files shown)


Yancey0623 (approved):
LGTM

PR #1055, opened 2019-10-22
Correct model zoo links in the design doc

ywskycn:
As titled.


tonyyang-svail (approved):
@ywskycn Thanks for the fix. :)

PR #1067, opened 2019-10-24
add feature derivation case e2e test

typhoonzero:
part work of https://github.com/sql-machine-learning/sqlflow/issues/992


pkg/sql/feature_derivation_test.go
```diff
@@ -159,8 +160,7 @@ func TestFeatureDerivation(t *testing.T) {
 	nc, ok = fc3.(*codegen.NumericColumn)
 	a.True(ok)
 
-	// trainIR.Features["feature_columns"][3] is the column class
```
tonyyang-svail, line 162: Was this a bug? Why was the label column in `trainIR.Features`?

tonyyang-svail (approved):
Excellent!

pkg/sql/feature_derivation_test.go
typhoonzero: The previous implementation includes the label in `trainIR.Features`. Removed now.

PR #1069, opened 2019-10-24
[DO NOT MERGE] test ci enhancement

typhoonzero:


typhoonzero:
Closing, docker images is hard to share between stages. We can push the docker image to some where, but it still takes time.

PR #1073, opened 2019-10-24
do not use hdfs namenode config

### LINKED ISSUE #1072: hive.py do not use hdfs:// and use the hadoop configuration on the server
hive.py should not use `hdfs://` to making directories and copy files, we should use the hadoop configurations on the `sqlflowserver` in cases there are multiple `namenodes` deployed. Then the `hdfs dfs -copyFromLocal` command can still succeed when one of the `namenodes` fails.

typhoonzero:
Fix https://github.com/sql-machine-learning/sqlflow/issues/1072


tonyyang-svail:
Hi @typhoonzero, would you mind explaining the motivation of this PR?

Yancey0623 (approved):
LGTM

PR #1094, opened 2019-10-28
codegen tensorflow use attr dict

typhoonzero:
Last part of https://github.com/sql-machine-learning/sqlflow/issues/992

### DIFF
--- a/pkg/sql/codegen/attribute/attribute.go
+++ b/pkg/sql/codegen/attribute/attribute.go
@@ -36,6 +36,8 @@ const (
 	String
 	// IntList indicates the corresponding attribute is a list of integers
 	IntList
+	// Unknown type indicates that the attribute type is dynamically determined.
+	Unknown
 )
 
 // Dictionary contains the description of all attributes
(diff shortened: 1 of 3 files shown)


pkg/sql/codegen/attribute/attribute.go
```diff
@@ -68,9 +70,27 @@ func (t Type) String() string {
 //   2. Customer checker
 func (d Dictionary) Validate(attrs map[string]interface{}) error {
 	for k, v := range attrs {
+		var desc *Description
 		desc, ok := d[k]
 		if !ok {
-			return fmt.Errorf(errUnsupportedAttribute, k)
+			// Support attribute defination like "model.*" to match attributes start with "model"
```
tonyyang-svail: "defination" => "definition"

pkg/sql/codegen/attribute/attribute.go
```diff
@@ -68,9 +70,27 @@ func (t Type) String() string {
...
+					return fmt.Errorf(errUnsupportedAttribute, k)
+				}
+			} else {
+				return fmt.Errorf(errUnsupportedAttribute, k)
+			}
+
+		}
+		// unknown type of attribute do not need to run validate
+		if desc.Type == Unknown {
+			return nil
```
tonyyang-svail: Potential bug: should be `continue`?

typhoonzero: Unknown type do not need to check the actual value's type, just added checker call here.

tonyyang-svail: `Validate` should validate all attributes. One attribute being `Unknown` doesn't mean we don't need to check the rest attributes. So we should change `return nil` to `continue`.

tonyyang-svail (approved):

pkg/sql/codegen/attribute/attribute.go
typhoonzero: I see, sorry my brain wa ta le.

PR #1108, opened 2019-10-31
default use ir to generate tf/xgboost code to run

typhoonzero:
- Default use IR codegen to generate local Tensorflow/XGBoost training code.
- Remove legacy codegen.


pkg/sql/codegen/attribute/checker.go
```diff
@@ -100,6 +100,28 @@ func IntRangeChecker(lower, upper *int, includeLower, includeUpper bool) func(in
 	return checker
 }
 
+// BoolIntChecker returns a checker function that do **not** check the input.
```
shendiaomo: The comment is not inconsistent with the following code, BoolIntChecker **do** check the input.

pkg/sql/codegen/tensorflow/codegen.go
```diff
@@ -39,6 +40,9 @@ range: [0,Infinity]`, attribute.IntRangeChecker(newInt(0), nil, false, false)},
 	"train.epoch": {attribute.Int, `[default=1]
 Number of epochs the training will run.
 range: [1, Infinity]`, attribute.IntRangeChecker(newInt(0), nil, false, false)},
+	"train.verbose": {attribute.Int, `[default=0]
+Show verbose logs when training.
+possible values: 0, 1`, attribute.IntChoicesChecker([]int{0, 1})},
```
weiguoz, line 37: IntChoiceChecker returns an error message like `fmt.Errorf("attribute value out of range(%v), actual: %d", choices, i)`, but it's not clear for the user to find out the relative attribute name.

shendiaomo: > IntChoiceChecker returns an error message like `fmt.Errorf("attribute value out of range(%v), actual: %d", choices, i)`, but it's not clear for the user to find out the relative attribute name.

This is an existing problem in other checkers, maybe I can fix this later in another commit

typhoonzero: @weiguoz I'll try polish the error message in the next PR.

Yancey0623 (approved):
LGTM, also please @Echo9573 and @BlackPoint-CX  try to use this refactor version. And please tell us if any questions.

PR #1120, opened 2019-11-02
Add details for how to find published models and saved models

typhoonzero:
Add some details about how to find published models and saved models and how to use it in sqlflow.


tonyyang-svail (approved):
LGTM.

PR #1125, opened 2019-11-04
Correct more misspells and typos

shendiaomo:
Correct serval misspells or typos or simply syntax errors I found.

### DIFF
--- a/pkg/sql/sql.y
+++ b/pkg/sql/sql.y
@@ -82,7 +82,7 @@
 
 	/* If no FOR in the COLUMN, the key is "" */
 	type columnClause map[string]exprlist
-	type filedClause  exprlist
+	type fieldClause  exprlist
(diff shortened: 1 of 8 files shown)


pkg/sql/sql.y
```diff
@@ -82,7 +82,7 @@
 
 	/* If no FOR in the COLUMN, the key is "" */
 	type columnClause map[string]exprlist
-	type filedClause  exprlist
+	type fieldClause  exprlist
```
weiguoz, line 85: It seems the old `filedClause` is  useless.

shendiaomo: Yes, I keep it just in case. Maybe we need another commit to clean such stale syntax rules.

typhoonzero (approved):

Issue #1111, opened 2019-10-31
[train&val] Support VALIDATE while running TRAIN


weiguoz:
According to #675, we are going to support the user specify dataset for validation by adding a  `VALIDATE ON` in `TRAIN` statement:
```SQL
SELECT ... 
TO TRAIN DNNClassifier
...
INTO my_model
VALIDATE ON (SELECT ...)    -- validation dataset
```

At this moment,  we plan to implement a workaround solution first due to the technical schedule.
```SQL
SELECT ... 
TO TRAIN DNNClassifier
WITH
    validation.dataset="SELECT ...",         -- validation dataset
    validation.loss_fun="loss_function"
...
INTO my_model;
```

- ~~Add keyword `VALIDATE ON` in `sql.y`.~~ PR: #1112  
- [x] Support specify `VALIDATE` dataset in `WITH` clause. #1124
- [x] Switch the old approach `train&val` to the `VALIDATE` mode #1124
- [x] Bring back validation for XGBoost #1130 
- [x] Verify the schema of `validation dataset` specified by `VALIDATE ON`

tonyyang-svail:
Hi @weiguoz, thanks for submitting this TODO list.

We need to make several design choices before rushing into implementation
1. Since we will be changing `TRAIN` to `TO TRAIN`, do we need to change `VALIDATE` to `TO VALIDATE`? Why?
1. Are parentheses `(` `)` required for specifying the validation select? Why?
1. Are we supporting arbitrary select statements? Why?
1. Should `VALIDATE` support the `WITH` clause in case we need to specify some configurations?

weiguoz:
Thanks for reminding me. 
I'm aware that this direct solution may have some technical risks. Considering this feature(supporting `validate`) has a higher priority. I would propose to specify the validation dataset in the `WITH` clause.

When more user suggestions are collected, we will back here adding the keyword `VALIDATE `.

weiguoz:
> Hi @weiguoz, thanks for submitting this TODO list.
> 
> We need to make several design choices before rushing into implementation
> 
> 1. Since we will be changing `TRAIN` to `TO TRAIN`, do we need to change `VALIDATE` to `TO VALIDATE`? Why?
> 2. Are parentheses `(` `)` required for specifying the validation select? Why?
> 3. Are we supporting arbitrary select statements? Why?
> 4. Should `VALIDATE` support the `WITH` clause in case we need to specify some configurations?

Thanks for reminding me. 
[...]

PR #1130, opened 2019-11-05
[train&val] Bring back validation for XGBoost

weiguoz:
Part of #1111

### DIFF
--- a/pkg/sql/codegen/xgboost/docgen.go
+++ b/pkg/sql/codegen/xgboost/docgen.go
@@ -36,8 +36,9 @@ const docTemplateText = `# XGBoost Parameters
 SELECT * FROM boston.train
 TRAIN xgboost.gbtree
 WITH
-    objective="reg:squarederror",
-    train.num_boost_round = 30
+    objective ="reg:squarederror",
+    train.num_boost_round = 30,
+    validation.select = "SELECT * FROM boston.train LIMIT 8"
 COLUMN crim, zn, indus, chas, nox, rm, age, dis, rad, tax, ptratio, b, lstat
 LABEL medv
 INTO sqlflow_models.my_xgb_regression_model;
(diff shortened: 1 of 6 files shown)


pkg/sql/codegen/xgboost/docgen.go
```diff
@@ -36,8 +36,9 @@ const docTemplateText = `# XGBoost Parameters
 SELECT * FROM boston.train
 TRAIN xgboost.gbtree
 WITH
-    objective="reg:squarederror",
-    train.num_boost_round = 30
+    objective ="reg:squarederror",
+    train.num_boost_round = 30,
+    validation.select = "SELECT * FROM boston.train LIMIT 8"
```
tonyyang-svail, line 41: The validation data is usually not the same as the training data. So maybe rename `boston.train` to `boston.val`?

tonyyang-svail (approved):

python/sqlflow_submitter/xgboost/train.py
```diff
@@ -33,8 +33,12 @@ def train(datasource, select, model_params, train_params, feature_field_meta, la
...
+        dvalidate = xgb_dataset(conn, 'validate.txt', validation_select, feature_column_name, label_name, feature_spec)
+        watchlist.append((dvalidate, "validate"))
 
-    bst = xgb.train(model_params, dtrain, **train_params)
-    bst.save_model("my_model")
\ No newline at end of file
+    re = dict()
+    bst = xgb.train(model_params, dtrain, **train_params, evals=watchlist, evals_result=re)
+    bst.save_model("my_model")
+    print(re)
```
typhoonzero, line 44: maybe add some message: `print("Validation result: %s" % re)`?

weiguoz: This will be fixed in the next `polish code` pr.

pkg/sql/codegen/xgboost/docgen.go
weiguoz: sure

PR #1097, opened 2019-10-28
[wip] Setup Argo test environment

tonyyang-svail:
Part of #1066

PR #1132, opened 2019-11-05
init go mod to manage dependencies

typhoonzero:
Document: https://blog.golang.org/using-go-modules

Previous issue: https://github.com/sql-machine-learning/sqlflow/pull/986

NOTE: if we use go mod, the `go get` command may upgrade the dependencies. Just run `go build` or `go test` only.

TODO: update build document.


go.mod
```diff
@@ -0,0 +1,21 @@
+module sqlflow.org/sqlflow
+
+require (
+	github.com/apache/thrift v0.13.0 // indirect
```
shendiaomo, line 4: Is there any mechanism of go mod that could handle such indirect modules automatically?

typhoonzero: Please check out the document link above. `go.mod` and `go.sum` are generated automatically.

wangkuiyi (approved):

scripts/test_hive.sh
```diff
@@ -30,14 +30,15 @@ set -e
 hdfs dfs -rm -r -f hdfs://localhost:8020/sqlflow
 hdfs dfs -mkdir -p hdfs://localhost:8020/sqlflow
 export SQLFLOW_HIVE_LOCATION_ROOT_PATH=/sqlflow
+# enable go mod
+export GO111MODULE=on
```
tonyyang-svail, line 34: Add `ENV GO111MODULE on` in Dockerfile also? You can refer to the changes on Dockerfile in https://github.com/sql-machine-learning/sqlflow/pull/1097

go.mod
```diff
@@ -0,0 +1,21 @@
...
+	github.com/asaskevich/govalidator v0.0.0-20190424111038-f61b66f89f4a
+	github.com/beltran/gohive v1.0.0 // indirect
```
wangkuiyi, line 6: @typhoonzero How come such `//indirect` dependency entries? Did you manually add them? If so, why did you do that, instead of leaving these indirect dependencies using their default versions?

scripts/test_hive.sh
wangkuiyi: @typhoonzero @tonyyang-svail Can we remove the settings of `GO111MODULE=on`  if we install Go 1.13 into our Docker image by editing the following code snippet?

https://github.com/sql-machine-learning/sqlflow/blob/2dc6657bc4095852c3cc666aadfbf0b6efca175f/scripts/build_docker_image.sh#L47-L51

I ask because it seems from https://blog.golang.org/using-go-modules that since 1.13, the module mode is by default on.

go.mod
typhoonzero: `go.mod` and `go.sum` is generated automatically by running `go test`.

PR #1138, opened 2019-11-05
Refactorize Dockerfile

wangkuiyi:
`scripts/build_docker_image.sh` uses ARG, VOLUME, and ENV instructions in Dockerfile. However, `scripts/build_docker_image.sh` is a long script, whose various parts rely on different Dockerfile instructions.  This PR splits `scripts/build_docker_image.sh` into small pieces, each focuses on installing a group of related tools. It then interleaves the small pieces with related Dockerfile instructions to make the Dockerfile comprehensive.


terrytangyuan (dismissed):
Neat!

typhoonzero (approved):

PR #1161, opened 2019-11-08
Refactor executor

tonyyang-svail:
1. Make `RunSQLProgram` the only entrance function. Currently, we have `ProgramToIR` and `RunAsync`.
1. Remove extra closure function definition in `runTrainIR/runPredictIR/runAnalyzeIR` like [here](https://github.com/sql-machine-learning/sqlflow/blob/60e00d7b9cea8e2ac73877abfd1725a0f7452686/pkg/sql/executor_ir.go#L99-L108). We can avoid extra indentation which makes the code more easy to read.
1. Combine the following code snippet in `runTrainIR/runPredictIR/runAnalyzeIR` in one place.
```
if err != nil {
		log.Errorf("runExtendedSQL error:%v", err)
		if err != ErrClosedPipe {
			if err := wr.Write(err); err != nil {
				log.Errorf("runExtendedSQL error(piping):%v", err)
			}
		}
		return err
	}
```

### DIFF
--- a/cmd/repl/repl.go
+++ b/cmd/repl/repl.go
@@ -116,13 +116,8 @@ func runStmt(stmt string, isTerminal bool, modelDir string, db *sql.DB, ds strin
 		return
 	}
 	defer os.RemoveAll(cwd)
-	programIR, err := sql.ProgramToIR([]string{stmt}, ds, modelDir)
-	if err != nil {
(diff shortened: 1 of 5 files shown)


typhoonzero (approved):
LGTM

PR #1171, opened 2019-11-12
Support calling Java parser in SQLFlow

tonyyang-svail:
This PR depends on #1170.

This PR is a part of #1126.


tonyyang-svail:
Calling Java parser takes about 1100ms for `HiveQL` parser and 800ms for `Calcite` parser. I am not sure if it's normal for Java command-line programs. If it is not normal, we'd better do some profiling on the Java program. @Yancey1989 @weiguoz Do you have experience in profiling? 

You can reproduce the result the following command.
```
$ echo "select 1" > /tmp/input.sql
$ time java -cp /opt/sqlflow/parser/parser-1.0-SNAPSHOT-jar-with-dependencies.jar org.sqlflow.parser.ParserAdaptorCmd -p calcite -i /tmp/input.sql -o /tmp/output.json

real	0m0.922s
user	0m0.690s
sys	0m0.530s
```

typhoonzero:
@tonyyang-svail can you write a test case that starts jvm once and run the parsing for 1000 times to see if the major time consumed is caused by starting the jvm?

tonyyang-svail:
@typhoonzero I have tested it by changing the command-line argument `calcite` to `i_dont_exists`, which would let the program exit before doing anything useful. And it takes about 137ms, which the original time is 900+ms. So should I assume the parsing algorithm is the bottleneck?

```
$ time java -cp /opt/sqlflow/parser/parser-1.0-SNAPSHOT-jar-with-dependencies.jar org.sqlflow.parser.ParserAdaptorCmd -p i_dont_exist -i /tmp/input.sql -o /tmp/output.json

invalid parser type i_dont_exist
real	0m0.137s
user	0m0.060s
sys	0m0.080s
```

typhoonzero:
@tonyyang-svail seems so. Maybe we can merge this first and do more profiling or performance fixes in another PR.

pkg/sql/tpp/parser.go
```diff
@@ -91,3 +86,50 @@ func tiDBParseAndSplit(sql string) ([]string, int, error) {
...
+	outputFile := filepath.Join(cwd, "output.json")
+	if err := ioutil.WriteFile(inputFile, []byte(sql), 755); err != nil {
+		return nil, -1, err
+	}
+
+	cmd := exec.Command("java",
+		"-cp", "/opt/sqlflow/parser/parser-1.0-SNAPSHOT-jar-with-dependencies.jar",
+		"org.sqlflow.parser.ParserAdaptorCmd",
+		"-p", typ,
+		"-i", inputFile,
```
typhoonzero, line 114: Maybe we can refine this later using the Linux pipe to avoid writing temporary files on the disk.

typhoonzero (approved):

weiguoz:
> Calling Java parser takes about 1100ms for `HiveQL` parser and 800ms for `Calcite` parser. I am not sure if it's normal for Java command-line programs. If it is not normal, we'd better do some profiling on the Java program. @Yancey1989 @weiguoz Do you have experience in profiling?
> 
> You can reproduce the result the following command.
> 
> ```
> $ echo "select 1" > /tmp/input.sql
> $ time java -cp /opt/sqlflow/parser/parser-1.0-SNAPSHOT-jar-with-dependencies.jar org.sqlflow.parser.ParserAdaptorCmd -p calcite -i /tmp/input.sql -o /tmp/output.json
> 
> real	0m0.922s
> user	0m0.690s
> sys	0m0.530s
> ```

I think the time is mainly costed in the startup phase after running some tests on my mac.
```CalciteParserAdaptor.ParseAndSplit(string sql)```

|# (test cases)|costs: millisecond|
|---|----|
|30,000|16,253|
|3,000|4,576|

https://github.com/weiguoz/sqlflow/blob/testParsePerf/java/parser/src/test/java/org/sqlflow/parser/CalciteParserAdaptorTest.java#L118-L120

To avoid startup the slow `JVM`:
1. [Calling by gRPC](https://redbyte.eu/en/blog/calling-java-from-go-using-grpc/)
1. [Java Native Interface Golang Interface](https://github.com/timob/jnigi) -- I didn't take a try.

weiguoz (approved):
LGTM.
Let's improve the `calcite` calling performance in the next PR

tonyyang-svail:
> I think the time is mainly costed in the startup phase after running some tests on my mac.
CalciteParserAdaptor.ParseAndSplit(string sql)

@weiguoz Wooow, you are awesome!

PR #1172, opened 2019-11-12
Alps code dev

ly120983:
change alps code gen to ir


typhoonzero:
To apply the new alps codegen, you need to update `executor_ir.go` to insert below code above https://github.com/sql-machine-learning/sqlflow/blob/develop/pkg/sql/executor_ir.go#L157:

```go
if os.Getenv("SQLFLOW_submitter") == "alps" {
    code, err := alps.Train(trainIR)
}
```

Then the generated alps python code will be executed.

tonyyang-svail:
Close to due lack of activity.

PR #1182, opened 2019-11-13
Make RunSQLProgram takes a single string

tonyyang-svail:
Part of #1126.

The RunSQLProgram is the only entrance of the `sql` package, it should take a SQL program and the subsequent parsing will split the program into a list of statements.

### DIFF
--- a/pkg/server/sqlflowserver_test.go
+++ b/pkg/server/sqlflowserver_test.go
@@ -46,9 +46,9 @@ const (
 
 var testServerAddress string
 
-func mockRun(sql []string, db *sf.DB, modelDir string, session *pb.Session) *sf.PipeReader {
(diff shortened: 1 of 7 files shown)


pkg/server/sqlflowserver_test.go
```diff
@@ -125,12 +125,13 @@ func TestSQL(t *testing.T) {
 	_, err = stream.Recv()
 	a.Equal(status.Error(codes.Unknown, "Lex: Unknown problem ..."), err)
 
-	testMultipleSQL := fmt.Sprintf("%s %s", testQuerySQL, testExtendedSQL)
```
tonyyang-svail, line 128: @typhoonzero I don't think we need to test multiple SQL in the unit test since we are using `mockRun`.

typhoonzero (commented):
LGTM, please fix CI then merge.

typhoonzero (approved):

PR #1184, opened 2019-11-13
katib xgboost codegen and submitter

samplise:
Include source codes of codegen module for katib xgboost, which generates katib_xgboost.py. Executing katib_xgboost.py will submit the xgboost experiment to katib.


tonyyang-svail:
This is PR is duplicated to #1186.

PR #1187, opened 2019-11-13
Switch to sqlflow parser

tonyyang-svail:
Part of #1126.

Depends on #1202.


pkg/sql/codegen/intermediate_representation.go
```diff
@@ -70,10 +70,6 @@ type SingleSQLIR interface {
 }
 
 // TrainIR is the intermediate representation for code generation of a training job.
-//
-// Please be aware that the TrainIR intentionally excludes the model table name in the
-// INTO clause. The sql package will save the output files of a generated Python program.
-// For prediction and analysis jobs, the sql will restore an identical working directly.
```
tonyyang-svail, line 76: SQLFlow generates a workflow program. The workflow program will save the model. So the `TrainIR` should contain `Into`.

pkg/sql/executor_ir.go
```diff
@@ -120,26 +120,35 @@ func runSingleSQLIR(wr *PipeWriter, ir codegen.SingleSQLIR, db *DB, modelDir str
...
 	}
 	defer os.RemoveAll(cwd)
 
-	if os.Getenv("SQLFLOW_submitter") == "elasticdl" {
-		return elasticDLTrain(wr, pr, db, cwd, session)
-	}
-	// FIXME(weiguo): temporary branch to alps
-	if os.Getenv("SQLFLOW_submitter") == "alps" {
-		return alpsTrain(wr, pr, db, cwd, session)
+	if os.Getenv("SQLFLOW_submitter") != "" {
```
typhoonzero, line 148: Also, need to update `runPredictIR` or do this in another PR?

pkg/sql/executor_ir.go
```diff
@@ -59,7 +59,7 @@ func RunSQLProgram(sqlProgram string, db *DB, modelDir string, session *pb.Sessi
 }
 
 func runSQLProgram(wr *PipeWriter, sqlProgram string, db *DB, modelDir string, session *pb.Session) error {
-	sqls, err := SplitMultipleSQL(sqlProgram)
+	sqls, err := parse(db.driverName, sqlProgram)
```
typhoonzero, line 62: `parse` and below `newParser().Parse(sql)` seems confusing, and the filename `parse.go` and `parser.go` also confusing, a better name is needed.

tonyyang-svail: Thanks for your suggestion. I have renamed `newParser().Parse(sql)` to `newExtendedSyntaxParser().Parse(sql)`.

typhoonzero (approved):

PR #1197, opened 2019-11-17
Add Couler design

wangkuiyi:
After the merge of this design, we could add the Couler Core file.


typhoonzero (approved):

typhoonzero:
Maybe put to `doc/design`?

PR #1200, opened 2019-11-18
Support history file in REPL

shendiaomo:


tonyyang-svail:
Hi @shendiaomo, could you please add tests for this package?

shendiaomo:
> Hi @shendiaomo, could you please add tests for this package?

Sure. There is already a `repl_test.go` in the package that covers most of the complex functions. I'll add some test cases later.

cmd/repl/prompt.go
```diff
@@ -167,6 +173,32 @@ func (p *promptState) initCompleter() {
...
+	defer f.Close()
+	scanner := bufio.NewScanner(f)
+	for scanner.Scan() {
+		p.history = append(p.history, scanner.Text())
+	}
+}
+
+func (p *promptState) updateHistory() {
+	if p.statement != "" {
+		f, err := os.OpenFile(p.historyFileName, os.O_CREATE|os.O_WRONLY|os.O_APPEND, 0644)
```
shendiaomo, line 191: Posix seems to guarantee that writes with O_APPEND are atomic. The APUE book also describes the behavior in detail.  So we don't need a file lock here.

weiguoz (approved):
LGTM

PR #1202, opened 2019-11-18
Fix Hive parser error position on multiple line statement

tonyyang-svail:
Part of #1126.


java/parser/src/main/java/org/sqlflow/parser/HiveQLParserAdaptor.java
```diff
@@ -61,7 +61,7 @@ public ParseResult ParseAndSplit(String sql) {
           // Consider select 1 to train, Calcite parser raise error at letter t of "to",
           // while HiveQL parser raise error at the white space before "to". As a result,
           // we put `+ 1` on the `epos`.
-          epos = posToIndex(sql, re.line, re.charPositionInLine) + 1;
+          epos = posToIndex(sql, re.line, re.charPositionInLine + 1);
```
tonyyang-svail, line 64: On parsing
```
SELECT *
FROM some_table
TO TRIAN ...
```

The Hive Parser gives error position as `line: 3, column: 0`. To make the error position at the character `T` of `TO`, we should add 1 to the column number.

typhoonzero: Also update the comment above?

typhoonzero (approved):

java/parser/src/main/java/org/sqlflow/parser/HiveQLParserAdaptor.java
tonyyang-svail: Sure, will do.

Issue #1063, opened 2019-10-23
[Model Zoo] TODO list of implementing model zoo


typhoonzero:
Design doc: https://github.com/sql-machine-learning/sqlflow/blob/develop/doc/design/model_zoo.md

1. [ ] unify SQLFlow and ElasticDL custom models https://github.com/sql-machine-learning/models/issues/22.
1. [x] support keras functional API models: https://github.com/sql-machine-learning/sqlflow/issues/1093
1. [x] support parse model name like `a_data_scientist/regressors:v0.2/MyDNNRegressor`
1. [ ] support pytorch custom model in model zoo.
1. [x] base Docker image for building all model Docker images, with python, ElasticDL, and other common dependencies.
1. [x] save model weights in a database table with columns of metadatas. (save in MySQL, Hive, MaxCompute)
1. [ ] HTTP service `models.sqlflow.org` to serve publishing and listing models.
1. [ ] submitter program runs a model in a docker image.
1. [ ] support using `models.sqlflow.org` in SQL statements.
1. [ ] support `SQLFLOW LIST|PUBLISH` command.

typhoonzero:
Related with couler: https://github.com/sql-machine-learning/sqlflow/issues/1066

typhoonzero:
Model zoo is implemented in latest SQLFlow.

PR #1217, opened 2019-11-20
support parse model zoo model name

typhoonzero:

support parse model name like `a_data_scientist/regressors:v0.2/MyDNNRegressor`

### DIFF
--- a/pkg/sql/lexer.go
+++ b/pkg/sql/lexer.go
@@ -99,7 +99,8 @@ func (l *lexer) lexIdentOrKeyword(lval *sqlSymType) int {
 	// lexToken ensures that the first rune is a letter.
 	r := l.next()
 	for {
-		for unicode.IsLetter(r) || unicode.IsNumber(r) || r == '_' {
+		// model IDENT may be like: a_data_scientist/regressors:v0.2/MyDNNRegressor
+		for unicode.IsLetter(r) || unicode.IsNumber(r) || r == '_' || r == '/' || r == ':' {
 			r = l.next()
 		}
 		if r != '.' { // The dot cannot be the last rune.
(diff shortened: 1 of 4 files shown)


pkg/sql/lexer.go
```diff
@@ -99,7 +99,8 @@ func (l *lexer) lexIdentOrKeyword(lval *sqlSymType) int {
 	// lexToken ensures that the first rune is a letter.
 	r := l.next()
 	for {
-		for unicode.IsLetter(r) || unicode.IsNumber(r) || r == '_' {
+		// model IDENT may be like: a_data_scientist/regressors:v0.2/MyDNNRegressor
+		for unicode.IsLetter(r) || unicode.IsNumber(r) || r == '_' || r == '/' || r == ':' {
```
tonyyang-svail, line 103: For the record, SQL syntax usually doesn't accept `/` or `:` being part of the identifier. Take MySQL for example, we can create table named `temp1`, but not `tem/p1` or `tem:p1`.

```SQL
mysql> create table temp1 (a integer);
Query OK, 0 rows affected (0.02 sec)

mysql> create table tem/p1 (a integer);
ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '/p1 (a integer)' at line 1
mysql> create table tem:p1 (a integer);
ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near ':p1 (a integer)' at line 1
```

However, `repo/image:tag` is the native syntax in Docker, so I think it is worthwhile to make the naming consistent by using `repo/image:tag`.

https://github.com/sql-machine-learning/sqlflow/pull/1042/files#r337120122 also mentioned this issue.

tonyyang-svail (approved):
LGTM.

pkg/sql/lexer.go
typhoonzero: Agree with keep the same naming with Docker image.

PR #1224, opened 2019-11-21
Add intermediate representation Protocol Buffers definition

typhoonzero:
implementing design: https://github.com/sql-machine-learning/sqlflow/pull/1222


pkg/sql/codegen/ir_to_proto_test.go
```diff
@@ -0,0 +1,65 @@
...
+		HiveLocation:     "/sqlflowtmp",
+		HdfsNamenodeAddr: "192.168.1.1:8020",
+		HdfsUser:         "sqlflow_admin",
+		HdfsPass:         "sqlflow_pass",
+	}
+	pbIR, err := TrainIRToProto(sampleTrainIR, sampleSession)
+	a.NoError(err)
+	pbtxt := proto.MarshalTextString(pbIR)
+	pbIRToTest := &irpb.TrainIR{}
+	err = proto.UnmarshalText(pbtxt, pbIRToTest)
```
tonyyang-svail, line 59: I am not sure why we need to `MarshalTextString/UnmarshText`, because we shouldn't test these two functions.

Can we have an exhaustive `a.Equal` between fields of `sampleTrainIR` and `pbIR`(not `pbIRToTest`)?

pkg/sql/codegen/proto/intermediate_representation.proto
```diff
@@ -0,0 +1,85 @@
...
+message FeatureColumnList {
+    repeated FeatureColumn feature_columns = 1;
+}
+
+message TrainIR {
+    string datasource = 1;
+    string select = 2;
+    string validation_select  = 3;
+    string estimator = 4;
+    map<string, string> attributes = 5;
```
tonyyang-svail, line 81: Can we make attribute type aware? Maybe add a message like
```proto
message Attribute {
  oneof attribute {
    int32 i =1;
    ...
  }
}

File: pkg/sql/codegen/proto/intermediate_representation.proto [hunk 3]
```diff
@@ -0,0 +1,85 @@
...
+        NumericColumn nc = 1;
+        BucketColumn bc = 2;
+        CrossColumn cc = 3;
+        CategoryIDColumn catc = 4;
+        SeqCategoryIDColumn seqcatc = 5;
+        EmbeddingColumn embc = 6;
+    }
+}
+
+// TODO(typhoonzero): directly import Session from pkg/server/proto
```
tonyyang-svail, line 60: I am not sure how to do the direct import. Do we need to include `pkg/server/proto` during `protoc`? If so, is it more convenient to organize all the proto files in one directory?

pkg/sql/codegen/ir_to_proto.go
```diff
@@ -0,0 +1,258 @@
...
+
+func toInt32List(il []int) []int32 {
+	ret := []int32{}
+	for _, i := range il {
+		ret = append(ret, int32(i))
+	}
+	return ret
+}
+
+func fieldMetaToPbMeta(fm *FieldMeta) *irpb.FieldMeta {
```
tonyyang-svail, line 85: Naming: `Pb` or `PB`? I feel `PB` is better since it is the acronym of Protocol Buffers.

pkg/sql/codegen/proto/intermediate_representation.proto
```diff
@@ -0,0 +1,85 @@
+syntax = "proto3";
```
tonyyang-svail, line 1: The proto file needs more comments like `TrainIR` since it will be read by many developers.

tonyyang-svail (approved):
LGTM. Please refer to the comments for suggested improvements.

pkg/sql/codegen/ir_to_proto_test.go
typhoonzero: We are going to use the IR at the python side by deserializing the message, so maybe it's better to test the fields after deserializing it.

pkg/sql/codegen/proto/intermediate_representation.proto
typhoonzero: Sounds good, will update in next pull request.

pkg/sql/codegen/proto/intermediate_representation.proto
typhoonzero: You are probably right, I still need to check the details at https://developers.google.com/protocol-buffers/docs/proto#importing-definitions

pkg/sql/codegen/proto/intermediate_representation.proto
typhoonzero: I'll merge this first and do these polishments in another PR.

PR #1230, opened 2019-11-22
[Parser] Remove extract space for original SQL

tonyyang-svail:
This PR fixes the following TODO.

https://github.com/sql-machine-learning/sqlflow/blob/49aa11d593b42b9ad6dfc02e7249f19d7a7aed00/pkg/sql/parse.go#L76

Part of #1126.

### DIFF
--- a/pkg/sql/parse.go
+++ b/pkg/sql/parse.go
@@ -58,6 +58,9 @@ func parse(dbms, sqlProgram string) ([]statementParseResult, error) {
 		return make([]statementParseResult, 0), nil
 	}
 
+	// SELECT * FROM my_table TO TRAIN ...
+	//                        ^
+	//                        i
 	sqls, i, err := thirdPartyParse(dbms, sqlProgram)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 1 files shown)


pkg/sql/parse.go
```diff
@@ -66,17 +69,24 @@ func parse(dbms, sqlProgram string) ([]statementParseResult, error) {
 		return sqls, nil
 	}
 
+	left := sqlProgram[:i]
```
typhoonzero, line 72:
```go
sqlProgram := sqlProgram[:i]
extendedSQLProgram = sqlProgram[i:]
```

tonyyang-svail: Hi @typhoonzero, I am assuming you are suggesting a better naming, which I don't agree.

Please be aware that the `sqlProgram` may contain multiple statements. For example, let's consider the following scenario.

```go
sqlProgram := "select ... to train ...; select ... to train ...;"

sqlProgram := sqlProgram[:i]         // "select ...", which is not a SQL program
extendedSQLProgram = sqlProgram[i:]  // "to train ...; select ... to train", which is not a extendedSQLProgram
```

typhoonzero: I see. I'm approving this.

typhoonzero (approved):

PR #1248, opened 2019-11-25
REPL: provide more comprehensive auto-completion for models and parameters

shendiaomo:
1. Use a python script `extract_docstring.py` to extract and parse docstrings for each parameter of the models. The result is written in `JSON` format into a constant string in `model_parameters.go`. The constant can be used as attributes for models, or to generate documents for SQLFlow, or to help with implementing auto-completion of SQLFlow IDEs (e. g. DataWorks)
2. Implement an advanced auto-completion for the SQLFlow REPL, including auto-completion of estimators and model parameters:
complete
This is also a demo for what the `model_parameters.go` file can do.
3. `extract_docstring.py` is only used to generate a static file because its stability in a server call still needs to be thought over.


python/extract_docstring.py
```diff
@@ -0,0 +1,103 @@
...
+import re
+import inspect
+import six
+import json
+import xgboost
+import sqlflow_models
+from tensorflow.estimator import *
+
+def parse_ctor_args(f, prefix=''):
+  r"""Given an class or function, parse the class constructor/function details
```
weiguoz, line 24: redundant `r` at the beginning?

python/extract_docstring.py
```diff
@@ -0,0 +1,103 @@
...
+  Args:
+    f: The class or function whose docstring to parse
+    prefix: The prefix of parameters in the docstring
+
+  Returns:
+    A dict with parameters as keys and splitted docstring as values
+  """
+
+  try:
+      doc = f.__init__.__doc__
```
weiguoz, line 57: inconsistent indent

weiguoz (approved):
LGTM
Excellent PR!

python/extract_docstring.py
shendiaomo: Thanks for finding this, I'll fix it in the next commit.

PR #1249, opened 2019-11-25
Fix parser bug

tonyyang-svail:
Part of #1126.


pkg/sql/parse.go
```diff
@@ -69,7 +69,7 @@ func parse(dbms, sqlProgram string) ([]statementParseResult, error) {
 		return sqls, nil
 	}
 
-	left := sqlProgram[:i]
+	left := sqls[len(sqls)-1].standard
```
tonyyang-svail, line 72: In the case of `SELECT ...; SELECT * FROM my_table TO TRAIN ...`, `sqlProgram[:i]` is `SELECT ...; SELECT * FROM my_table`, while it should be `SELECT * FROM my_table`.

typhoonzero (approved):

PR #1255, opened 2019-11-26
Rename several members to be more consistent

shendiaomo:
Rename the `TrainIR` members to `TrainClause` to be consistent with the renaming of `TrainIR` struct to `TrainClause`

### DIFF
--- a/pkg/sql/codegen/pai/codegen_test.go
+++ b/pkg/sql/codegen/pai/codegen_test.go
@@ -113,7 +113,7 @@ func TestWrapperCodegen(t *testing.T) {
 
 func TestTrainCodegen(t *testing.T) {
 	a := assert.New(t)
-	ir := mockTrainIR()
+	ir := mockTrainClause()
 
 	paiTfCode, err := doTrain(ir, "my_dnn_model")
 	a.NoError(err)
@@ -152,7 +152,7 @@ func TestPredictCodegen(t *testing.T) {
 	a.False(hasUnknownParameters(tfCode, knownPredictParams))
 }
 
-func mockTrainIR() *ir.TrainClause {
+func mockTrainClause() *ir.TrainClause {
 	_ = `SELECT * FROM iris_train TO TRAIN DNNClassifier
          WITH train.batch_size=4,
 		      train.epoch=3,
@@ -190,6 +190,6 @@ func mockPredIR() *ir.PredictClause {
 		Select:      "select * from iris_test;",
 		ResultTable: "iris_predict",
 		Attributes:  make(map[string]interface{}),
-		TrainIR:     mockTrainIR(),
+		TrainClause: mockTrainClause(),
 	}
 }
--- a/pkg/sql/codegen/tensorflow/codegen.go
+++ b/pkg/sql/codegen/tensorflow/codegen.go
@@ -237,7 +237,7 @@ func Train(trainIR *ir.TrainClause) (string, error) {
 // Pred generates a Python program for predict using a TensorFlow model.
 func Pred(predIR *ir.PredictClause, session *pb.Session) (string, error) {
 	modelParams := make(map[string]interface{})
-	for attrKey, attr := range predIR.TrainIR.Attributes {
+	for attrKey, attr := range predIR.TrainClause.Attributes {
 		if strings.HasPrefix(attrKey, "model.") {
 			modelParams[strings.Replace(attrKey, "model.", "", 1)] = attr
 		}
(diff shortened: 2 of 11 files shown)


pkg/sql/codegen/pai/codegen_test.go
```diff
@@ -190,6 +190,6 @@ func mockPredIR() *ir.PredictClause {
 		Select:      "select * from iris_test;",
 		ResultTable: "iris_predict",
 		Attributes:  make(map[string]interface{}),
-		TrainIR:     mockTrainIR(),
+		TrainClause: mockTrainClause(),
```
typhoonzero, line 193: We can keep the name `TrainIR` because `TrainClause` do not have enough message that it is an `ir.TrainClause` struct.

wangkuiyi: @typhoonzero What is the difference between TrainIR and TrainClause?

Do you mean here TrainIR means the parsing result of the SQL statement consisting of the SELECT statement and the TRAIN clause?

typhoonzero: > @typhoonzero What is the difference between TrainIR and TrainClause?
> 
> Do you mean here TrainIR means the parsing result of the SQL statement consisting of the SELECT statement and the TRAIN clause?

I mean the object name `TrainClause` we actually don't know it's a struct or a `TO TRAIN ...` string, when we use the object name `TrainClause`, the type `ir.TrainClause` can tell it's an IR struct. But the name `TrainIR` can tell it's an IR struct.

We can compare the below two lines of code, and pick up a better naming though:

```go
trainIR := &ir.TrainClause{...}
trainClause := &ir.TrainClause{...}
irTrainClause := &ir.TrainClause{...}
```

shendiaomo: > > @typhoonzero What is the difference between TrainIR and TrainClause?
> > Do you mean here TrainIR means the parsing result of the SQL statement consisting of the SELECT statement and the TRAIN clause?
> 
> I mean the object name `TrainClause` we actually don't know it's a struct or a `TO TRAIN ...` string, when we use the object name `TrainClause`, the type `ir.TrainClause` can tell it's an IR struct. But the name `TrainIR` can tell it's an IR struct.
> 
> We can compare the below two lines of code, and pick up a better naming though:
> 
> ```go
> trainIR := &ir.TrainClause{...}
> trainClause := &ir.TrainClause{...}
> irTrainClause := &ir.TrainClause{...}
> ```

The naming convention that prefix/suffix a variable name with its type such as `irTrainClause` is known as Systems Hungarian, which is not recommended nowadays because the type prefix/suffix doesn't provide more information. See https://www.joelonsoftware.com/2005/05/11/making-wrong-code-look-wrong/ for details and https://stackoverflow.com/questions/111933/why-shouldnt-i-use-hungarian-notation for short. In this perspective, `TrainIR` is not better than `TrainClause`.
Maybe `ModelBuildingContext` is more informative because the member is actually representing the original model building code.

typhoonzero: Agree that not using "Systems Hungarian". Yet comparing `TrainIR` and `TrainClause`, as mentioned by @wangkuiyi  above, we call the `ir.TrainClause` an "Intermediate Representation" for now, and it consists of `Datasource`, `Select` and `ValidationSelect` etc, which apparently not the part of `TrainClause`, so I'd prefer `TrainIR`.

Yet if we change the naming `IR` to some other names like `TrainContext`, then the name should be `trainCxt` as an example.

By the way, the naming `ir.TrainClause` seems not describing what it is since it contains `Datasource` and so on, it really should be changed to `ir.TrainContext` or `ir.TrainSpec`, @shendiaomo what's your suggestion? 

I prefer `ir.TrainContext` in these two names because `TrainSpec` reminds people of `tf.estimator.TrainSpec` while they are very different.

typhoonzero: Also cc @Yancey1989 @weiguoz @tonyyang-svail

tonyyang-svail: I agree that we need to change the `TrainClause` since it contains more information than a clause. This struct is originally named `TrainIR`, and since it belongs to `package ir`, so why not simply named it `Train`, so we have `trainIR := & ir.Train{...}`.

typhoonzero: I vote for `trainIR := & ir.Train{...}`.

PR #1269, opened 2019-11-27
Add regex for lower case limit in feature derivation

### LINKED ISSUE #1263: lowercase keywords 'limit' report an error when training model
There is an error reported when I use the lowercase keyword 'limit', shown as follows:
image
image

tonyyang-svail:
Fix #1263.

### DIFF
--- a/pkg/sql/feature_derivation.go
+++ b/pkg/sql/feature_derivation.go
@@ -230,7 +230,20 @@ func InferFeatureColumns(trainIR *ir.TrainClause) error {
 	fcMap := makeFeatureColumnMap(trainIR.Features)
 	fmMap := makeFieldMetaMap(trainIR.Features)
 
-	q := fmt.Sprintf("SELECT * FROM (\n%s\n) a LIMIT %d", trainIR.Select, featureDerivationRows)
+	// TODO(typhoonzero): find a way to using subqueries like select * from (%s) AS a LIMIT 100
+	q := trainIR.Select
+	re, err := regexp.Compile("(LIMIT [0-9]+|limit [0-9]+)")
+	if err != nil {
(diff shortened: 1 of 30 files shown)


pkg/sql/feature_derivation.go
```diff
@@ -230,20 +230,7 @@ func InferFeatureColumns(trainIR *ir.TrainClause) error {
...
-	limitClauseIndexes := re.FindStringIndex(q)
-	if limitClauseIndexes == nil {
-		q = fmt.Sprintf("%s LIMIT %d", q, featureDerivationRows)
-	} else {
-		// TODO(typhoonzero): there may be complex SQL statements that contain multiple
-		// LIMIT clause, using regex replace will replace them all.
-		re.ReplaceAllString(q, fmt.Sprintf("LIMIT %d", featureDerivationRows))
-	}
-
+	q := fmt.Sprintf("SELECT * FROM (\n%s\n) a LIMIT %d", trainIR.Select, featureDerivationRows)
```
typhoonzero: Seems using subqueries here will cause error: https://travis-ci.com/sql-machine-learning/sqlflow/jobs/260837168, I'll change it back to the regex way for a fast fix. Remove the `LIMIT 10` from train SQL, it works well.

Running the subquery SQL using hive command line also works. Need to find out a solution.

Yancey0623 (approved):
LGTM.

tonyyang-svail:
@typhoonzero @Yancey1989 Please update the commit message and PR title if we are using a different solution. Even after reading https://github.com/sql-machine-learning/sqlflow/pull/1269#discussion_r351132750, I still took a while to realize the solution has been updated.

PR #1272, opened 2019-11-27
Refactor executor_ir.go using the visitor pattern

### LINKED ISSUE #1239: Too many if branches for PAI submitter
`SubmitterPAI` is one of many supported ML engines for SQLFlow. However, in the main logic of `runTrainIR`, there are three if statements on `SubmitterPAI`, which hurts the readability. 

https://github.com/sql-machine-learning/sqlflow/blob/e907cada17489c650802884dafc233b1ed039a21/pkg/sql/executor_ir.go#L178

https://github.com/sql-machine-learning/sqlflow/blob/e907cada17489c650802884dafc233b1ed039a21/pkg/sql/executor_ir.go#L194

https://github.com/sql-machine-learning/sqlflow/blob/e907cada17489c650802884dafc233b1ed039a21/pkg/sql/executor_ir.go#L221

@shendiaomo Could you please try to refactor this? If it is hard to refactor, then there might be a design flaw.

shendiaomo:
Wikipedia:
> The visitor design pattern is a way of separating an algorithm from an object structure on which it operates. A practical result of this separation is the ability to add new operations to _existing object structures_ without modifying the structures

In our scenario, the above _existing object structure_ is the `ir.SQLStatement` interface and the four types implemented that interface, because there're always requirements to add new operations to the object structure(i.e. add a new submitter), but the structure itself is relatively stable.
This pull request extracts two new concepts: an `Executor` is a visitor to `SQLStatement`, a `Submitter` extends `Executor` and represents a submitter platform such as _"pai"_, _"alps"_ or _"elasticdl"_. 

`defaultSubmitter` encapsulates the core functionality of the former `TrainIR`, `PredictIR` and `AnalyzeIR` functions and exposes several utility methods. As a result, it is a piece of cake to implement `alpsSubmitter` and `elasticdlSubmitter`.

The pull request uses a `map[string]Submmiter` as a type registry.

`paiSubmitter` demos how a new submitter can be implemented as a new file without touching any other existing source files

This also fixes #1239 

**TODO**
1. Move `submitter.go` and all its dependencies into a new package `submitter`
1. A better design to replace the `if` checking `xgboost` or `tensorflow`

### DIFF
--- a/pkg/sql/executor_ir.go
+++ b/pkg/sql/executor_ir.go
@@ -15,21 +15,11 @@ package sql
 
 import (
 	"bytes"
-	"encoding/base64"
 	"fmt"
-	"io"
-	"io/ioutil"
-	"os"
-	"os/exec"
-	"path"
 	"strings"
-	"sync"
 	"time"
 
 	pb "sqlflow.org/sqlflow/pkg/proto"
-	"sqlflow.org/sqlflow/pkg/sql/codegen/pai"
-	"sqlflow.org/sqlflow/pkg/sql/codegen/tensorflow"
-	"sqlflow.org/sqlflow/pkg/sql/codegen/xgboost"
 	"sqlflow.org/sqlflow/pkg/sql/ir"
 )
 
@@ -43,35 +33,6 @@ type EndOfExecution struct {
 // WorkflowJob indicates the Argo Workflow ID
 type WorkflowJob struct {
 	JobID string
-}
-
-var envSubmitter = os.Getenv("SQLFLOW_submitter")
-
-// SubmitterType is the type of SQLFlow submitter
-type SubmitterType int
-
-const (
-	// SubmitterPAI indicates that SQLFlow uses the PAI platform as submitter
-	SubmitterPAI = iota
-	// SubmitterEDL indicates that SQLFlow uses ElasticDL as submitter
-	SubmitterEDL
-	// SubmitterALPS indicates that SQLFlow uses ALPS as submitter
-	SubmitterALPS
-	// SubmitterDefault indicates that SQLFlow uses the default submitter
-	SubmitterDefault
-)
-
-func submitter() SubmitterType {
-	switch envSubmitter {
-	case "pai":
-		return SubmitterPAI
-	case "elasticdl":
-		return SubmitterEDL
-	case "alps":
-		return SubmitterALPS
-	default:
-		return SubmitterDefault
-	}
 }
 
 // RunSQLProgram run a SQL program.
@@ -155,9 +116,9 @@ func runSQLProgram(wr *PipeWriter, sqlProgram string, db *DB, modelDir string, s
 			if parsed.train {
 				r, err = generateTrainIRWithInferredColumns(parsed, connStr)
 			} else if parsed.analyze {
-				r, err = generateAnalyzeIR(parsed, connStr, modelDir, submitter() != SubmitterPAI)
+				r, err = generateAnalyzeIR(parsed, connStr, modelDir, submitter().GetTrainIRFromModel())
 			} else {
-				r, err = generatePredictIR(parsed, connStr, modelDir, submitter() != SubmitterPAI)
+				r, err = generatePredictIR(parsed, connStr, modelDir, submitter().GetTrainIRFromModel())
 			}
 		} else {
 			standardSQL := ir.StandardSQL(sql.standard)
@@ -171,7 +132,6 @@ func runSQLProgram(wr *PipeWriter, sqlProgram string, db *DB, modelDir string, s
 			return e
 		}
 	}
-
 	return nil
 }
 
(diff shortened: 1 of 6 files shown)


pkg/sql/executor_ir.go
```diff
@@ -187,242 +147,11 @@ func runSingleSQLIR(wr *PipeWriter, sqlIR ir.SQLStatement, db *DB, modelDir stri
...
-	imgBytes, err := ioutil.ReadAll(imgFile)
-	if err != nil {
-		return err
-	}
-	imgBase64Str := base64.StdEncoding.EncodeToString(imgBytes)
-	img2html := fmt.Sprintf("<div align='center'><img src='data:image/png;base64,%s' /></div>", imgBase64Str)
-	wr.Write(img2html)
-	return nil
+	defer submitter().Teardown()
+	return sqlIR.Execute(submitter())
```
typhoonzero, line 154: We can execute a program, execute(interpret) the intermediate representation, but not execute a submitter.

typhoonzero (commented):
It's an excellent idea to use polymorphism to implement multiple execution branches.

Yet I have several questions to be discussed:

image
1. It seems the lines of code are not reducing after this refactor.
1. The name "submitter" is referred to `python/sqlflow_submitter` for now.
1. `Executor` and `submitter` is a little bit confusing

shendiaomo:
> It's an excellent idea to use polymorphism to implement multiple execution branches.
> 
> Yet I have several questions to be discussed:
> 
> image
> 
> 1. It seems the lines of code are not reducing after this refactor.

Two reasons:
1. The core idea behind this pull request (and OOP itself) is to make the code more maintainable rather than to reduce code size. In this pull request, we can see the source of each `submitter` is much shorter than they were before, and each submitter only needs to depend on itself rather than intervene with code from other submitters or `executor_ir.go`. This is an obvious sign of better maintainability. Better maintenance means a lot, for example, there will be much fewer code conflicts in pull requests after this commit.

1. In fact, the pull request does reduce the **real** code size by about 1/4 to 1/3, the extra lines added are either boilerplate code such as `import` and `package` because there are more files now, or comments such as the copyright sections and exported name comments because there are more names exported now.

> 2. The name "submitter" is referred to `python/sqlflow_submitter` for now.
> 3. `Executor` and `submitter` is a little bit confusing

`Submitter` is named after the existing environment variable `SQLFLOW_submitter`, and `Executor` is named after `executor_ir.go`. `Submitter` is a species of `Executor`, anyway, better suggestions are welcomed.

pkg/sql/executor_ir.go
shendiaomo: > We can execute a program, execute(interpret) the intermediate representation, but not execute a submitter.

Execute is an intransitive and transitive verb.
Transitive verb example:
> A system that executes a program is called an interpreter of the program. ✅

Intransitive verb example:
> Very few programs execute on a bare machine.✅

Both the examples are quoted from https://en.wikipedia.org/wiki/Execution_(computing)
So we can say _The intermediate representation executes **on** a submitter_.

pkg/sql/submitter.go
```diff
@@ -0,0 +1,240 @@
...
+func (s *defaultSubmitter) ExecuteTrain(cl *ir.TrainClause) (e error) {
+	var code string
+	if isXGBoostModel(cl.Estimator) {
+		code, e = xgboost.Train(cl)
+	} else {
+		code, e = tensorflow.Train(cl)
+	}
+	if e == nil {
+		if e = s.runCommand(code); e == nil {
+			e = s.SaveModel(cl)
```
tonyyang-svail, line 156: Style guide: "Try to keep the normal code path at a minimal indentation, and indent the error handling, dealing with it first. This improves the readability of the code by permitting visually scanning the normal path quickly." https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

So maybe rewrite this function to following in the next PR. This comment also applies to other places in this PR.
```go
func (s *defaultSubmitter) ExecuteTrain(cl *ir.TrainClause) (e error) {
	var code string
	if isXGBoostModel(cl.Estimator) {
		if code, e = xgboost.Train(cl); e != nil {
			return e
		}
	} else {
		if code, e = tensorflow.Train(cl); e != nil {
			return e
		}
	}
	
	if e = s.runCommand(code); e != nil {
		return e
	}
	
	return s.SaveModel(cl)
}
```

I haven't seen `if e == nil` much in Go project, so please avoid it as much as possible.

pkg/sql/executor_ir.go
tonyyang-svail: > So we can say The intermediate representation executes on a submitter.

@typhoonzero @shendiaomo How about changing `sqlIR.Execute(submitter())` to `sqlIR.ExecuteOn(submitter())`.

tonyyang-svail (commented):
@shendiaomo Thanks for submitting this excellent PR.

@typhoonzero I tend to approve it since it decouples the logic in each Go submitters, therefore it improves the quality and maintainability of the codebase.

> The name "submitter" is referred to python/sqlflow_submitter for now.

@typhoonzero The submitter contains the execution logic of an ML job. However, in our current implementation, some of the logics are in Go (saving trained model into a database table, loading TrainSQL for PredictSQL to do feature derivation), and some of the logics are in Python (the model training, prediction and explaining). Therefore, the submitter concept exists in both Go and Python. We can't avoid it unless we port all the functionalities (feature derivation, model saving together with the train SQL statement, model loading with the train SQL statement, basically everything besides the parsing) to Python. Before we do the porting, we may keep the name submitter in both Go and Python.

typhoonzero (approved):

pkg/sql/submitter.go
shendiaomo: I'll fix this later in another commit.

PR #1278, opened 2019-11-28
Restrict TensorFlow model attributes to the available set

### LINKED ISSUE #1275: SQLFlow doesn't recognize an optimizer name
**Description**
Input:
```sql
SELECT * FROM iris.train TO TRAIN DNNClassifier WITH model.hidden_units=[300,200], model.n_classes=3, model.optimizer="RMSProp" LABEL class INTO sqlflow_models.test_dnn_model;
```
Output:
```
==========Output==========
  File "<stdin>", line 68, in <module>
NameError: name 'RMSProp' is not defined
```

shendiaomo:
Restrict TensorFlow model attributes to the available parameter set as mentioned in https://github.com/sql-machine-learning/sqlflow/pull/1248#issue-345298608


pkg/sql/codegen/attribute/attribute.go
```diff
@@ -158,3 +159,35 @@ func (d Dictionary) GenerateTableInHTML() string {
...
+// PremadeModelParamsDocs stores parameters and documents of all known models
+var PremadeModelParamsDocs map[string]map[string]string
+
+func init() {
+	if err := json.Unmarshal([]byte(ModelParameterJSON), &PremadeModelParamsDocs); err != nil {
+		panic(err) // assertion
+	}
+	// The following parameters of canned estimators are already supported in the COLUMN clause.
+	for _, v := range PremadeModelParamsDocs {
+		delete(v, "feature_columns")
```
typhoonzero, line 186: `attribute.go` is not engine specific and these attributes seems only used in Tensorflow.

pkg/sql/codegen/attribute/attribute.go
```diff
@@ -158,3 +159,35 @@ func (d Dictionary) GenerateTableInHTML() string {
 	l = append(l, `</table>`)
 	return strings.Join(l, "\n")
 }
+
+// Update updates `d` by copyint from `other` key by key
```
typhoonzero: What's "copyint"?

pkg/sql/codegen/attribute/attribute.go
```diff
@@ -158,3 +159,35 @@ func (d Dictionary) GenerateTableInHTML() string {
...
+func (d Dictionary) Update(other Dictionary) Dictionary {
+	for k, v := range other {
+		d[k] = v
+	}
+	return d
+}
+
+// NewDictionary create a new Dictionary according to `estimator`
+func NewDictionary(estimator, prefix string) Dictionary {
```
typhoonzero, line 172: `estimator` is Tensorflow specific, and we should not add engine specific code in `attribute.go`

pkg/sql/codegen/tensorflow/codegen.go
```diff
@@ -159,9 +157,19 @@ func IsKerasModel(estimator string) (bool, string) {
 	return false, fmt.Sprintf("tf.estimator.%s", estimator)
 }
 
+func validateAttributes(trainIR *ir.TrainClause) error {
+	modelAttr := attribute.NewDictionary(trainIR.Estimator, "model.")
+	for name, attr := range modelAttr {
+		if strings.HasSuffix(name, "optimizer") {
```
typhoonzero: I think this is kind of temporary. The `optimizer` argument in Tensorflow Estimators can be a string as well as an `Optimizer` object, we need also pass things like `learning_rate`, `learning_rate_decay` etc. to construct the optimizer object.

Seems it's better to use a struct:

```
type Optimizer struct {
    Name string
    params map[string]interface{}
}
```
To store the construction details of the optimizer object. If `params` is empty, then pass the optimizer name as string to the Estimator, else construct an object and pass it to Estimator.

And, it's useful to update `python/sqlflow_submitters/tensorflow/train.py` to have an argument `optimizer` to accept it.

typhoonzero (changes requested):

pkg/sql/codegen/tensorflow/codegen.go
typhoonzero: For reference: https://www.tensorflow.org/api_docs/python/tf/keras/optimizers

pkg/sql/codegen/attribute/attribute.go
shendiaomo: Done.

pkg/sql/codegen/attribute/attribute.go
shendiaomo: `estimator` is not TensorFlow specific, sklearn proposed the original concept and it is widely accepted by the AI community, XGBoost implemented the `estimator` interface too.
For another example, we also have `ir.TrainClause.Estimator` with the same meaning as the `estimator` here.

pkg/sql/codegen/tensorflow/codegen.go
shendiaomo: > I think this is kind of temporary. The `optimizer` argument in Tensorflow Estimators can be a string as well as an `Optimizer` object, we need also pass things like `learning_rate`, `learning_rate_decay` etc. to construct the optimizer object.
> 
> Seems it's better to use a struct:
> 
> ```
> type Optimizer struct {
>     Name string
>     params map[string]interface{}
> }
> ```
> 
> To store the construction details of the optimizer object. If `params` is empty, then pass the optimizer name as string to the Estimator, else construct an object and pass it to Estimator.
> 
> And, it's useful to update `python/sqlflow_submitters/tensorflow/train.py` to have an argument `optimizer` to accept it.

Yes, this is a transitional plan. To allow constructing an `Optimizer` in SQL statement requires a more comprehensive design on which I'm still working.

pkg/sql/codegen/attribute/attribute.go
shendiaomo: They are used both in TensorFlow and PAI TensorFlow, that's why I put them here. Though `codegen/pai` shares part of `codegen/tensorflow` at the moment, they may diverge someday.

pkg/sql/codegen/tensorflow/codegen.go
tonyyang-svail: > To allow constructing an Optimizer in SQL statement requires a more comprehensive design on which I'm still working.

Maybe something like the following?

```
WITH
    optimizer.name = "adam"
    optimizer.param.lr = 0.1
```

shendiaomo: Let's refine this pull request to only restrict the TensorFlow model attributes. Another pull request about optimizer will be issued later.

go.mod
```diff
@@ -5,21 +5,32 @@ require (
 	github.com/asaskevich/govalidator v0.0.0-20190424111038-f61b66f89f4a
 	github.com/beltran/gohive v1.0.0 // indirect
 	github.com/c-bata/go-prompt v0.0.0-20190826134812-0f95e1d1de2e
+	github.com/cosiner/argv v0.0.1 // indirect
```
typhoonzero, line 8: Did you ran `go get ./...` before commiting? we are using gomod now, and `go get` will try to update all dependencies.

typhoonzero (approved):

go.mod
shendiaomo: I ran `go test ./...` which I thought is the reason for this.

PR #1279, opened 2019-11-28
Use a case insensitive regexp to make feature derivation robust to typos

shendiaomo:
A slight improvement for #1269.


typhoonzero (approved):
cool!

PR #1280, opened 2019-11-29
parse only output proto for predict and analyze

typhoonzero:


pkg/sql/ir/ir_to_proto.go
```diff
@@ -199,17 +199,25 @@ func featureColumnToPb(fc FeatureColumn) (*pb.FeatureColumn, error) {
 	}
 }
 
-// TrainIRToProto convert parsed TrainIR to a protobuf format
-// TODO(typhoonzero): add PredictIR, AnalyzeIR
-func TrainIRToProto(trainIR *TrainClause, sess *pb.Session) (*pb.TrainIR, error) {
+// AttributesToProto convert attributes from IR to protobuf format
+func AttributesToProto(attrsIR map[string]interface{}) (map[string]*pb.Attribute, error) {
```
shendiaomo, line 203: Can we use the messages from `proto/intermediate_representation.proto` to replace the original structs in the `sql/ir/ir.go`? In that case, we don't need to maintain these conversion functions.

typhoonzero: Protobuf IR struct is harder to use because the `oneof` type will construct many layers to call, but Go's `interface{}` is simpler to use in codegens.

shendiaomo (approved):

pkg/sql/ir/ir_to_proto.go
shendiaomo: LGTM. Let's keep the status quo until we find a better way.

PR #1281, opened 2019-11-29
add modeluri for model zoo to save

typhoonzero:

### DIFF
--- a/pkg/sql/model.go
+++ b/pkg/sql/model.go
@@ -19,22 +19,69 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"path"
 	"path/filepath"
+	"strings"
 
 	pb "sqlflow.org/sqlflow/pkg/proto"
 	"sqlflow.org/sqlflow/pkg/sqlfs"
 )
+
+const modelZooDB = "sqlflow"
+const modelZooTable = "sqlflow.trained_models"
 
 type model struct {
 	workDir     string // We don't expose and gob workDir; instead we tar it.
 	TrainSelect string
 }
 
-// save creates a sqlfs table if it doesn't yet exist, and writes the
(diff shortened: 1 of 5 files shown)


pkg/sqlfs/writer.go
```diff
@@ -23,7 +23,8 @@ import (
 	pb "sqlflow.org/sqlflow/pkg/proto"
 )
 
-const bufSize = 4 * 1024
+// TEXT/STRING field support 64KB maximumn storage size
+const bufSize = 32 * 1024
```
shendiaomo, line 27: Can we make this `bufSize` configurable? If my memory serves me right, the `string` type in MaxCompute has a much bigger size than 64KB.

pkg/sql/model.go
```diff
@@ -19,22 +19,69 @@ import (
...
+				dir, file := path.Split(uriParts[1])
+				return m.saveTar(dir, file)
+			} else if uriParts[0] == "oss" {
+				return fmt.Errorf("save model to oss is not supported now")
+			}
+		} else {
+			return fmt.Errorf("error modelURI format: %s", modelURI)
+		}
+	}
+	db, err := NewDB(session.DbConnStr)
```
shendiaomo, line 53: Is it possible to unify `sqlfs` with `modelURI`? For example, can we make `sqlfs://database_name/table_name` behave the same as `table_name`? This seems to improve consistency.

pkg/sqlfs/writer.go
typhoonzero: Better to keep it constant for better saved model migrations.

pkg/sql/model.go
typhoonzero: `oss://` and `file://` are well know URI prefixes. We do not intend to invent a new one.

shendiaomo (approved):

PR #1285, opened 2019-12-02
[ModelZoo] save model meta to sqflow.trained_models

typhoonzero:

Should merge https://github.com/sql-machine-learning/sqlflow/pull/1281 first

TODO: support hive and maxcompute


pkg/sql/model.go
```diff
@@ -183,3 +198,86 @@ func readGob(filePath string, object interface{}) error {
...
+	createDBSQL := fmt.Sprintf("CREATE DATABASE IF NOT EXISTS %s;", modelZooDB)
+	_, err := db.Exec(createDBSQL)
+	if err != nil {
+		return err
+	}
+	// schema design: https://github.com/sql-machine-learning/sqlflow/blob/a98218ef8bee57e2a45357d7ee5721e1c6dfeb35/doc/design/model_zoo.md#model-zoo-data-schema
+	// NOTE(typhoonzero): submitter program size may exceed TEXT size(64KB)
+	// NOTE(typhoonzero): train_ir_pb contains all information how the submitter program is generated, so not saving submitter program now
+	// NOTE(typhoonzero): model_uri can be:
+	//    1. file:///path/to/model/dir
```
tonyyang-svail, line 214: `file://` or `file:///`?

pkg/sql/model.go
```diff
@@ -183,3 +198,86 @@ func readGob(filePath string, object interface{}) error {
...
+	return "", into, nil
+}
+
+func dbStringEscape(src string) string {
+	ret := strings.ReplaceAll(src, "\"", "\\\"")
+	return strings.ReplaceAll(ret, "'", "\\'")
+}
+
+func addTrainedModelsRecord(db *DB, trainIR *ir.TrainClause, modelURI string, sess *pb.Session) error {
+	fmt.Println("in addTrainedModelsRecord")
```
tonyyang-svail: ~~`fmt.Println`~~

pkg/sql/model.go
```diff
@@ -183,3 +198,86 @@ func readGob(filePath string, object interface{}) error {
...
+	q := fmt.Sprintf("SELECT * FROM %s WHERE model_id='%s'", modelZooTable, modelID)
+	fmt.Println(q)
+	res, err := db.Query(q)
+	if err != nil {
+		return err
+	}
+	defer res.Close()
+	isInsert := false
+	if !res.Next() {
+		isInsert = true
```
tonyyang-svail, line 257: Maybe add a test examing the `isInsert` logic.

tonyyang-svail (dismissed):
LGTM. Are we enabling model zoo by default?

pkg/sql/model.go
```diff
@@ -183,3 +198,86 @@ func readGob(filePath string, object interface{}) error {
...
+	//    1. file:///path/to/model/dir
+	//    2. db.table
+	//    3. oss://path/to/oss
+	createTableSQL := fmt.Sprintf(`CREATE TABLE IF NOT EXISTS %s (
+model_id VARCHAR(255),
+author VARCHAR(255),
+model_image VARCHAR(255),
+model_def VARCHAR(255),
+train_ir_pb TEXT,
+model_uri VARCHAR(255)
```
weiguoz, line 223: Do we need timestamp fields, like `createdAt`, `updatedAt`?
By timestamp, we can eliminate the overdue data.

pkg/sql/model.go
typhoonzero: The prefix is `file://` and the path is `/path/to/model/dir` started by `/`.

pkg/sql/model.go
typhoonzero: `CaseTrainFeatureDerevation` has two train SQL which have the same `INTO` clause, this is testing this if branch. Adding comments in that test case.

pkg/sql/model.go
typhoonzero: Sounds good, will add in next commit.

tonyyang-svail (approved):

PR #1288, opened 2019-12-02
Return feature derivation message to notebook

### LINKED ISSUE #1270: An error reported when the column is not "*" in  validation.select
some error will be reported when the column is not "*" in  the `validation.select` , the detail is shown as following figure:
image

typhoonzero:
Fix https://github.com/sql-machine-learning/sqlflow/issues/1270


pkg/sql/feature_derivation.go
```diff
@@ -400,5 +401,16 @@ func InferFeatureColumns(trainIR *ir.TrainClause) error {
...
+	// Write back to client information about feature derivation result
+	if wr != nil {
+		for target, fclist := range trainIR.Features {
+			for _, fc := range fclist {
+				for _, fm := range fc.GetFieldMeta() {
+					wr.Write(fmt.Sprintf("Using column (%s) in feature column (%T) as model construct param (%s)", fm.Name, fc, target))
+				}
+			}
+		}
+		wr.Write("\n")
```
weiguoz, line 417: What's the '\n' for?

pkg/sql/feature_derivation.go
```diff
@@ -400,5 +401,16 @@ func InferFeatureColumns(trainIR *ir.TrainClause) error {
 			}
 		}
 	}
+	// Write back to client information about feature derivation result
+	if wr != nil {
+		for target, fclist := range trainIR.Features {
```
tonyyang-svail, line 410: Can we put this code snippet as a separate function `func LogFeatureDerivationResult(wr *PipeWriter, trainIR *ir.TrainClause) error` and call this function at 

https://github.com/sql-machine-learning/sqlflow/blob/a5986ba188feea1081627f32d17383b499f66d35/pkg/sql/submitter.go#L131-L142

so that
1. We can avoid changing `InferFeatureColumns(trainIR)` to `InferFeatureColumns(wr, trainIR)`.
1. We can reuse this function for our debugging due to `w := io.MultiWriter(cw, &output)`.

pkg/sql/feature_derivation.go
typhoonzero: To separate feature derivation logs and training logs.

pkg/sql/feature_derivation.go
typhoonzero: Good point, yet it's better to put it in runSingleSQLIR, where trainIR can be easily got.

pkg/sql/feature_derivation.go
```diff
@@ -221,6 +221,7 @@ func fillFieldMeta(columnTypeList []*sql.ColumnType, rowdata []interface{}, fiel
 
 // InferFeatureColumns fill up featureColumn and columnSpec structs
 // for all fields.
+// if wr is not nil, then write
```
tonyyang-svail, line 224: ~~`// if wr is not nil, then write`~~

tonyyang-svail (approved):

PR #1316, opened 2019-12-03
Add -fs flag in hdfs command

ggaaooppeenngg:
Allow remote namenode of hdfs.

Signed-off-by: Peng Gao <peng.gao.dut@gmail.com>

### DIFF
--- a/python/sqlflow_submitter/db_writer/hive.py
+++ b/python/sqlflow_submitter/db_writer/hive.py
@@ -20,7 +20,7 @@ import subprocess
 CSV_DELIMITER = '\001'
 
 class HiveDBWriter(BufferedDBWriter):
-    def __init__(self, conn, table_name, table_schema, buff_size=10000, 
+    def __init__(self, conn, table_name, table_schema, buff_size=10000,
                  hdfs_namenode_addr="", hive_location="",
                  hdfs_user="", hdfs_pass=""):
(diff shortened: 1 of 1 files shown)


python/sqlflow_submitter/db_writer/hive.py
```diff
@@ -96,9 +96,9 @@ def write_hive_table(self):
             hdfs_envs.update({"HADOOP_USER_NAME": self.hdfs_user})
         if self.hdfs_pass != "":
             hdfs_envs.update({"HADOOP_USER_PASSWORD": self.hdfs_pass})
-        cmd_str = "hdfs dfs -mkdir -p %s/%s/" % (hdfs_path, self.table_name)
+        cmd_str = "hdfs dfs -fs hdfs://%s -mkdir -p %s/%s/" % (namenode_addr, hdfs_path, self.table_name)
```
typhoonzero: We are not using `namenode_addr` currently because, in some deployment, the HDFS client needs to be configured to connected to multiple namenodes so that this deployment is Highly Available.

typhoonzero: Maybe if namenode_addr is set (not empty string), then add `-fs` else, the command keeps `"hdfs dfs -mkdir -p %s/%s/"` to support both multi namenode setup and single namenode setup.

typhoonzero (approved):

Issue #1265, opened 2019-11-26
Enable PAI TensorFlow CI


tonyyang-svail:
https://github.com/sql-machine-learning/sqlflow/blob/b7f12bc023b88ba0fa3b24cb42b4f33a6ca53b40/scripts/test_maxcompute.sh#L37-L39

PR #1333, opened 2019-12-04
Add a python utility function to create temporaries on OSS

shendiaomo:
A preparation for #1265

### DIFF
--- a/python/sqlflow_submitter/pai/utils.py
+++ b/python/sqlflow_submitter/pai/utils.py
@@ -1,0 +1,63 @@
+# Copyright 2019 The SQLFlow Authors. All rights reserved.
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import oss2
+import os
+import uuid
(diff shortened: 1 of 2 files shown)


python/sqlflow_submitter/pai/utils.py
```diff
@@ -0,0 +1,63 @@
...
+    oss_temporary copies `filename` to a temporary object on OSS and deletes it a.s.a.p.
+    Example: 
+    with oss_temporary(YOUR_AK, YOUR_SK, ENDPOINT, 'test.py') as f:
+        do_something_with(f)
+    '''
+    bucket_name = 'sqlflow-pai-submitter'
+    auth = oss2.Auth(ak, sk)
+    bucket = oss2.Bucket(auth, endpoint, bucket_name)
+    
+    bucket.create_bucket(oss2.BUCKET_ACL_PRIVATE, oss2.models.BucketCreateConfig(oss2.BUCKET_STORAGE_CLASS_IA))
```
typhoonzero, line 55: maybe don't need to create a bucket every time.

shendiaomo: It's ok because what `create_bucket` do is indeed `create_if_not_exist`

python/sqlflow_submitter/pai/utils.py
```diff
@@ -0,0 +1,63 @@
+# Copyright 2019 The SQLFlow Authors. All rights reserved.
```
wangkuiyi, line 1: I would suggest not to name source dirs or files by 'util'; a more precise name is often preferred.  How about renaming this file to be `oss_temporary.py`?

wangkuiyi (approved):

python/sqlflow_submitter/pai/utils.py
shendiaomo: I meant to provide some other functions in this file, anyway, I'll come up with a better name.

PR #1334, opened 2019-12-04
Do not send feature derivation logs to client to be less annoying

### LINKED ISSUE #1331: The output of the training statement too complicated.
The output of the training statement is too complicated. It is recommended to remove or just keep the columns' name list. As the following shows:

image

typhoonzero:
Fix https://github.com/sql-machine-learning/sqlflow/issues/1331


Yancey0623 (approved):
LGTM

PR #1328, opened 2019-12-04
add PAI e2e test for internal env

typhoonzero:
Add one end to end test for submitting jobs in the internal environment. To run the test:

```bash
export SQLFLOW_submitter=pai
export SQLFLOW_TEST_DB=maxcompute
export MAXCOMPUTE_PROJECT="xxx"
export MAXCOMPUTE_ENDPOINT="xxx"
export MAXCOMPUTE_AK="xxx"
export MAXCOMPUTE_SK="xxx"
cd cmd/sqlflowserver && go test -v -run TestEnd2EndMaxComputePAI
```

Issue #1321, opened 2019-12-03
[PAI Submitter] Add PAI submitter


typhoonzero:
PAI is a machine learning platform on aliyun: https://help.aliyun.com/product/30347.html

We need to support submit jobs as PAI Tensorflow jobs:

- [ ] Connect to DB Proxy Alisa . @weiguoz @shendiaomo 
- [x] Use dataset to read from MaxCompute data, create temp maxcompute table in one couler step  @typhoonzero 
- [x] save model to OSS without credentials
- [x] Distributed training support @typhoonzero 
- [x] Support submiting PAI couler program and PAI single submit
- [ ] Support custom models (and model zoo) @typhoonzero 
- [ ] Enable CI @shendiaomo  @weiguoz 
- [ ] Support Analyze/GBDT, on spark or local @Yancey1989 @shendiaomo

PR #1336, opened 2019-12-04
PAI submitter support read directly from odps

### LINKED ISSUE #1335: Create a temp MaxCompute table from SELECT statement and run the training
Part of https://github.com/sql-machine-learning/sqlflow/issues/1321

For submitting training jobs to PAI and read data from MaxCompute, we need to generate two Couler steps from one train SQL:

1. run `create table tmp_training_table from (select ...)` to create a temp table.
2. use the temp table in `TableRecordDataset` when submitting a PAI training job.

typhoonzero:
Resolves https://github.com/sql-machine-learning/sqlflow/issues/1335
part of https://github.com/sql-machine-learning/sqlflow/issues/1321

Should merge https://github.com/sql-machine-learning/sqlflow/pull/1328 first


cmd/sqlflowserver/main_test.go
```diff
@@ -45,6 +45,9 @@ var caseTrainTable = "train"
 var caseTestTable = "test"
 var casePredictTable = "predict"
 
+// caseInto used by CaseTrainSQL
```
wangkuiyi: Sorry, but what is CaseTrainSQL?

pkg/sql/codegen/tensorflow/codegen.go
```diff
@@ -209,6 +211,20 @@ func Train(trainIR *ir.TrainClause) (string, error) {
 			fmt.Sprintf("\"%s\": [%s]", target, strings.Join(perTargetFeatureColumnsCode, ",\n")))
 	}
 	isKeras, estimatorStr := IsKerasModel(trainIR.Estimator)
+	isPAI := os.Getenv("SQLFLOW_submitter") == "pai"
+	paiTable := ""
+	// TODO(typhoonzero): if isPAI, create two Couler steps
+	if isPAI {
+		fromRegex, err := regexp.Compile("FROM\\s([a-z0-9_\\.]*)")
```
shendiaomo: 1. Use a raw string to get rid of escaping?
1. Use `\s*` in case there is more than one whitespace, and I think we may need a multiline setting to make the following work on this.
    ```sql
    select * from 
        iris.trainable ...
    ```
3. May the table name capitalized? If so, we'd better use `[\w\.]` instead of `[a-z0-9_\\.]`

cmd/sqlflowserver/main_test.go
typhoonzero: updated the comment to:

```
// caseInto is used by function CaseTrainSQL in this file. When
// testing with MaxCompute, the project is pre-created, we only need to
// specify the table name in that case.
```

pkg/sql/codegen/tensorflow/codegen.go
```diff
@@ -209,6 +211,20 @@ func Train(trainIR *ir.TrainClause) (string, error) {
...
+		if err != nil {
+			return "", err
+		}
+		matches := fromRegex.FindAllStringSubmatch(trainIR.Select, -1)
+		if len(matches) != 1 {
+			return "", fmt.Errorf("only support simple SQL query, but got %s", trainIR.Select)
```
shendiaomo: Can't we utilize the third party SQL parser to support complex SQL queries here?

typhoonzero: No. Queries with subqueries will not write the result to a table, the result is produced in memory.

pkg/sql/codegen/tensorflow/codegen.go
```diff
@@ -215,7 +215,7 @@ func Train(trainIR *ir.TrainClause) (string, error) {
 	paiTable := ""
 	// TODO(typhoonzero): if isPAI, create two Couler steps
 	if isPAI {
-		fromRegex, err := regexp.Compile("FROM\\s([a-z0-9_\\.]*)")
+		fromRegex, err := regexp.Compile("FROM[\\s\\n]+([\\w\\.]*)")
```
shendiaomo, line 218: I'm not sure, but if my memory serves me, the golang `regexp` requires a `(?m)` prefix to enable the multiline mode, in that case, `\s` will match `\n` automatically. Can we add a test case for the multiline case?

typhoonzero: Already tested on http://regex-golang.appspot.com/

shendiaomo (approved):
lgtm.

PR #1339, opened 2019-12-04
Don't exit interactive mode on slight error

### LINKED ISSUE #1337: Any error in a SQL statement would cause REPL exits the interactive mode
image
I believe this is caused by  fd63870

shendiaomo:
Fix #1337


Yancey0623 (approved):
LGTM

PR #1315, opened 2019-12-03
Fix gocyclo createByType

wangkuiyi:
Screen Shot 2019-12-02 at 9 12 36 PM

PR #1344, opened 2019-12-05
Make full use of reflect to simplify code

shendiaomo:
This is a follow-up of #1315 .


pkg/sql/column_type.go
```diff
@@ -26,104 +26,8 @@ import (
...
-	} // end of instantiators.
-)
-
-func createByType(rt reflect.Type) (interface{}, error) {
-	f, ok := instantiators[rt]
-	if ok {
-		return f(), nil
-	}
-	return nil, fmt.Errorf("unrecognized column scan type %v", rt)
+func createByType(t reflect.Type) interface{} {
```
weiguoz, line 29: The meaning of `createByType` is not clear.

weiguoz (approved):

pkg/sql/column_type.go
wangkuiyi: What is your suggestion?

wangkuiyi (approved):

pkg/sql/column_type.go
shendiaomo: 3x for pointing out this, I'll come up with a better name later.

PR #1347, opened 2019-12-05
Replace seconds in minute with unix seconds

ggaaooppeenngg:
Signed-off-by: Peng Gao <peng.gao.dut@gmail.com>

### DIFF
--- a/pkg/server/sqlflowserver.go
+++ b/pkg/server/sqlflowserver.go
@@ -168,7 +168,7 @@ func parse2Any(val interface{}) (proto.Message, error) {
 		return &wrappers.BytesValue{Value: v}, nil
 	case time.Time:
 		return &pyts.Timestamp{
-			Seconds: int64(v.Second()),
+			Seconds: int64(v.Unix()),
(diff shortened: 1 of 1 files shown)


pkg/server/sqlflowserver.go
```diff
@@ -168,7 +168,7 @@ func parse2Any(val interface{}) (proto.Message, error) {
 		return &wrappers.BytesValue{Value: v}, nil
 	case time.Time:
 		return &pyts.Timestamp{
-			Seconds: int64(v.Second()),
+			Seconds: int64(v.Unix()),
```
wangkuiyi, line 171: Thank you for pointing this out!

I tried to understand the change by reading https://godoc.org/github.com/golang/protobuf/ptypes/timestamp#Timestamp.GetSeconds and saw the example code:

```
Timestamp timestamp;
timestamp.set_seconds(time(NULL));
timestamp.set_nanos(0);
```

Where the POSIX call `time` returns the UNIX seconds.

wangkuiyi (approved):

PR #1359, opened 2019-12-06
Fast fail REPL if DB connection fails && Add test cases for this

### LINKED ISSUE #1338: REPL shows an unpleasant  error message if its connection is refused
image

On the opposite, the MySQL client shows a more friendly message in the same situation:
image

shendiaomo:
1. Fix #1338 and add a couple of test cases to ensure it's fixed.
2. Propose a mechanism to test the `main` function and its exit status.

### DIFF
--- a/cmd/repl/repl_test.go
+++ b/cmd/repl/repl_test.go
@@ -17,10 +17,13 @@ import (
 	"bufio"
 	"bytes"
 	"container/list"
+	"fmt"
 	"os"
+	"os/exec"
 	"regexp"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/golang/protobuf/proto"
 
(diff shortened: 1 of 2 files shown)


cmd/repl/repl_test.go
```diff
@@ -36,6 +39,44 @@ import (
 
 var space = regexp.MustCompile(`\s+`)
 
+func testMainFastFail(t *testing.T, interactive bool) {
+	a := assert.New(t)
+	// Run the crashing code when FLAG is set
+	if os.Getenv("FLAG") == "false" {
```
wangkuiyi: For less probability of name conflicting, I assume we'd change `FLAG` to a longer and more precise name like `SQLFLOW_TEST_FAST_FAIL_INTERACTIVE_OR_NOT`.

wangkuiyi (commented):
Issue #1338 is about SQLFlow doesn't present a friendly error message when DB connection fails. I don't understand why we could fix this issue by adding a unit test?

cmd/repl/repl_test.go
shendiaomo: > Issue #1338 is about SQLFlow doesn't present a friendly error message when DB connection fails. I don't understand why we could fix this issue by adding a unit test?

The bug in #1338 occurs since the original `repl.go` doesn't check the connectability of DB before entering the interactive mode. The fix is actually the new  function `assertConnectable` in `repl.go`. and the call to it in the function `main`

shendiaomo: > For less probability of name conflicting, I assume we'd change `FLAG` to a longer and more precise name like `SQLFLOW_TEST_FAST_FAIL_INTERACTIVE_OR_NOT`.

Done.

tonyyang-svail (approved):
LGTM.

PR #1376, opened 2019-12-10
Support specifying optimizers in SQLFlow

shendiaomo:
Fix https://github.com/sql-machine-learning/sqlflow/pull/1278#discussion_r352413256 , #1275 and #869

### DIFF
--- a/pkg/sql/codegen/tensorflow/codegen.go
+++ b/pkg/sql/codegen/tensorflow/codegen.go
@@ -169,8 +169,75 @@ func IsKerasModel(estimator string) (bool, string) {
 	return false, fmt.Sprintf("tf.estimator.%s", estimator)
 }
 
-func validateAttributes(trainStmt *ir.TrainStmt) error {
+// TODO(shendiaomo): Make the optimizer related code more general and exported in `attribute.go` if other frameworks
+// than TensorFlow have to support python objects as model attributes.
+
+func attrIsOptimizer(attrKey string) bool {
+	switch attrKey {
+	case "model.optimizer", "model.dnn_optimizer", "model.linear_optimizer":
+		return true
+	}
+	return false
+}
+
+func setDefaultOptimizer(trainStmt *ir.TrainStmt, optimizerParamName string) {
+	// TODO(shendiaomo): Try to get the default value from the python `inspect` module instead of hard coding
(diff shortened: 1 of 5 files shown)


pkg/sql/codegen/tensorflow/codegen.go
```diff
@@ -169,14 +169,81 @@ func IsKerasModel(estimator string) (bool, string) {
...
+	optimizerArgs := map[string]map[string]interface{}{}
+	for k, v := range trainStmt.Attributes {
+		if attrIsOptimizer(k) {
+			if optimizerArgs[k] == nil {
+				optimizerArgs[k] = map[string]interface{}{}
+			}
+
+		}
+		pieces := strings.Split(k, ".")
+		if len(pieces) == 3 {
```
typhoonzero: else return some optimizer format error?

pkg/sql/codegen/tensorflow/codegen.go
```diff
@@ -169,14 +169,81 @@ func IsKerasModel(estimator string) (bool, string) {
...
+		}
+	}
+	trainStmt.Attributes[optimizerParamName] = defaultValue
+}
+
+func constructOptimizers(trainStmt *ir.TrainStmt) {
+	optimizerArgs := map[string]map[string]interface{}{}
+	for k, v := range trainStmt.Attributes {
+		if attrIsOptimizer(k) {
+			if optimizerArgs[k] == nil {
```
typhoonzero, line 201: may be use piece[1] as the key, e.g. `model.optimizer.learning_rate` -> `optimizer`, but not `model.optimizer`?

shendiaomo: That's my original idea that's attractive. I hesitated for a long time before I choose the current naming. Let's discuss it with the others. @tonyyang-svail @Yancey1989

pkg/sql/codegen/tensorflow/codegen.go
shendiaomo: That's not necessary. If the attribute is not an optimizer, it will stay untouched thus raise an error in `initializeAttributes`.

tonyyang-svail (approved):
Generally LGTM.

PR #1399, opened 2019-12-11
add tensorflow docgen

typhoonzero:


pkg/sql/codegen/xgboost/docgen.go
```diff
@@ -34,7 +34,7 @@ const docTemplateText = `# XGBoost Parameters
 
 ` + "```SQL" + `
 SELECT * FROM boston.train
-TRAIN xgboost.gbtree
+TO TRAIN xgboost.gbtree
```
wangkuiyi, line 37: I think in the model zoo design, we plan to make `TO TRAIN` followed by a Docker image name + '/' + model name.  I don't remember that we could have `.` in the name.  Is there something we should update the format of the name? If so, let's update the model zoo design.

typhoonzero: We can have `.` already now, because https://github.com/sql-machine-learning/sqlflow/blob/develop/pkg/sql/lexer.go#L103 the `IsNumber` can accept `.`

tonyyang-svail (approved):
LGTM++.

PR #1427, opened 2019-12-15
Rename column type.go into fields

wangkuiyi:
- Rename file name `column_type.go` into `fields.go`, because this source file is about SQL fields but not TensorFlow input columns.

- Rename `parseVal` into `fieldValue`.

- Rename `universalizeColumnType` into `fieldType`.


typhoonzero (approved):
Thanks~

PR #1441, opened 2019-12-16
Enable text highlighting in magic cells

### LINKED ISSUE #1381: Need to improve SQL Syntax color scheme
Color scheme should be applied to all the standard SQL/SQLFlow syntax. Under current version, only a few preposition words like from and with are colored in green, while other syntax including select, where, to train and etc are all plain text, which is not really user-friendly.

shendiaomo:
Fix #1381 .
The result on my MacBook looks like:
image
Most of the keywords show color correctly. I'll later try to find a way to highlight the extended keyword correctly. (Probably in a different color.)


wangkuiyi (approved):
不明觉厉！

PR #1414, opened 2019-12-13
Make semicolon at the end of the SQL statement optional

tonyyang-svail:
Fix #1276.

PR #1443, opened 2019-12-16
Rename some syntax rules

### LINKED ISSUE #1419: What is the difference between select_stmt and select in sql.y
The rule defining `select` is as follows

https://github.com/sql-machine-learning/sqlflow/blob/d0db43bdb7fb74298b7816915a0f5d1a8b934145/pkg/sql/sql.y#L209-L216

It looks to me that `select` should have a more accurate name of `standard_select`, am I correct?

### LINKED ISSUE #1420: Rename opt_where and opt_limit to where_clause and limit_clause
How about we change `opt_where` into `where_clause`, and `opt_limit` to `limit_clause` to keep the naming consistent?

https://github.com/sql-machine-learning/sqlflow/blob/68ae7c62934d3086579ddd244b2648d4ffa3bbf4/pkg/sql/sql.y#L218-L226

wangkuiyi:
Fix https://github.com/sql-machine-learning/sqlflow/issues/1420
Fix https://github.com/sql-machine-learning/sqlflow/issues/1419

Also addresses the comment https://github.com/sql-machine-learning/sqlflow/pull/1414#discussion_r357934028

### DIFF
--- a/pkg/sqlfs/hive_writer_test.go
+++ b/pkg/sqlfs/hive_writer_test.go
@@ -17,6 +17,7 @@ import (
 	"fmt"
 	"io"
 	"math/rand"
+	"os"
 	"os/exec"
 	"path"
 	"strings"
@@ -35,7 +36,7 @@ func hasHDFSDir(hdfsPath string) bool {
 }
 
 func TestNewHiveWriter(t *testing.T) {
-	testDriver = getEnv("SQLFLOW_TEST_DB", "hive")
+	testDriver := os.Getenv("SQLFLOW_TEST_DB")
 	if testDriver != "hive" {
 		t.Skip("Skip as SQLFLOW_TEST_DB is not Hive")
 	}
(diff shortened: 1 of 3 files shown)


pkg/sqlfs/hive_writer_test.go
```diff
@@ -35,7 +36,7 @@ func hasHDFSDir(hdfsPath string) bool {
 }
 
 func TestNewHiveWriter(t *testing.T) {
-	testDriver = getEnv("SQLFLOW_TEST_DB", "hive")
+	testDriver := os.Getenv("SQLFLOW_TEST_DB")
```
wangkuiyi, line 39: This quick fix is off the topic of this PR. I have to add it because I need it to fix the failure of the unit test. Without this fix, when I run `go test -v ./...`, the call to `getEnv` returns "hive" as the default value, which then triggers the run of `TestNewHiveWriter` even if there is no Hive Server setup, and the unit test fails.

PR #1454, opened 2019-12-17
Split the emacs key bindings into a separate file

shendiaomo:
This is to decouple keybindings from `prompt.go`, so as to make both of them easier to maintain.


Yancey0623 (approved):
LGTM

weiguoz (approved):

PR #1470, opened 2019-12-18
Enable highlighting for extended SQL in code cells

shendiaomo:
This change slightly improves #1441 by supporting highlighting extended SQL keywords in light green.
The result seems like:
image


wangkuiyi (approved):
Amazing! We are having Javascript code in our codebase now! 💯

PR #1482, opened 2019-12-19
remove legacy expression resolver and codegen for alps and elasticdl

### LINKED ISSUE #1452: Remove legacy syntax rules
These following syntax rules parse standard SELECT statements in an ad-hoc way.

https://github.com/sql-machine-learning/sqlflow/blob/4df83c65e1559ebf3360ba7accaa5f55394a4102/pkg/parser/extended_syntax_parser.y#L162-L188

We haven't removed them because `codegen_alps.go` depends on them.  We don't have the bandwidth at the right moment to update `codegen_alps.go`. Let us remove it, and the above syntax rules. We can add `codegen_alps.go` back later after we have the bandwidth.

typhoonzero:
Fix https://github.com/sql-machine-learning/sqlflow/issues/1452

Backup branch:
 https://github.com/sql-machine-learning/sqlflow/tree/backup_codegen_alps_elasticdl


wangkuiyi:
Does this PR also remove codegen_elasticdl.go? We might need to let @terrytangyuan  and @uuleon know about the detailed plan, like when we will add these codegens back.

terrytangyuan (approved):
😞 Sadly approving. Really should've exposed stable APIs first before letting others implement new submitters.

wangkuiyi:
We will add `codegen_{alps,elasticdl}.go` back ASAP. @terrytangyuan @uuleon 

We have to remove them temporarily because it is really hard labor to gradually upgrade the SQLFlow feature column resolver while keeping the upgrading working with codegens.  This hard labor is due to that we have too much technical debt deposited during the recent months when we were rushing towards application milestones.

Without the old executor and the codegens depending on them, we could have PRs for upgrading small and easy to review.  We will try to finish the upgrade ASAP and add the codegens back, before we upgrade the codegens accordingly.

PR #1489, opened 2019-12-19
Add more default train val metrics

### LINKED ISSUE #590: How to customize the appropriate model evaluation indicators in sqlflow?
The existing model evaluation indicator is accuracy. For some classification projects, some other model evaluation indicators are needed, such as recall, F1, etc. For some regression projects, like RMSE, MAE, etc. In practical applications, how to customize the appropriate model evaluation indicators in sqlflow?

typhoonzero:
Fix https://github.com/sql-machine-learning/sqlflow/issues/590

Note, User should be aware of which metric can be used by the current model, e.g. AUC must be used in a binary classification model.


tonyyang-svail (commented):
Maybe add a test case? The test case can also serve as an example.

tonyyang-svail (approved):
LGTM.

PR #1521, opened 2019-12-24
Support showing figures as Braille ASCII Art when there's no iTerm2

shendiaomo:
This is a follow-up of #1507 .
This commit implements a matplotlib terminal backend with [plotille](https://github.com/tammoippen/plotille) and uses the backend to show figures when iTerm2 is not available. The result seems like:
image
which looks rather close to the original figure:
Sixel Image


shendiaomo:
**Note:** We strongly recommend users use iTerm2, this is an alternative if iTerm2 is not available for some users, and these users can still use the data URL to view images on a web browser.

wangkuiyi (dismissed):

wangkuiyi (approved):
LGTM++

PR #1528, opened 2019-12-24
Remove impossible conditions

shendiaomo:
We do not need to determine whether validation_select is empty because it's never empty: https://github.com/sql-machine-learning/sqlflow/blob/3a9aa22da2aaf002038edf71ed03c6429af05463/pkg/sql/ir_generator.go#L98-L100


python/sqlflow_submitter/tensorflow/train.py
```diff
@@ -122,37 +121,34 @@ def train_input_fn():
         train_dataset = train_dataset.shuffle(SHUFFLE_SIZE).batch(batch_size).cache().repeat(epochs if epochs else 1)
         return train_dataset
 
-    if validate_select == "":
-        classifier.train(input_fn=lambda:train_dataset)
```
shendiaomo, line 126: The bug here is never triggered because the control flow never passes the condition check.

typhoonzero: I think the fix should be: do not add default validation select, and if validation select is empty, do not run the validation.

shendiaomo: The aim of the commit is to remove useless(and buggy) code, which does not change any logic. What you suggest is another topic that slightly changes users' experience. How about addressing these two requirements separately? The latter can be provided in another pull request after some discussion.

typhoonzero: If we agree with "do not add default validation select", then some changes made in this PR need to be reverted once we merged this PR. I recommend close this PR and file another PR to directly enhance this feature.

typhoonzero (dismissed):
Approving. We may need to create an issue recording this, and add cases when the user is not specifying "validation select" in WITH clause.

typhoonzero (approved):

PR #1529, opened 2019-12-24
[Refactor] refactor pkg sql

typhoonzero:
1. move `PipeWriter`, `PipeReader` to a single package `pipe`, and change name to `pipe.Writer` and `pipe.Reader`.
2. move feature derivation to a single package.
3. move verifier to a single package.


pkg/featurederivation/feature_derivation.go
```diff
@@ -11,7 +11,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package sql
+package featurederivation
```
wangkuiyi, line 14: How about featurederivation => feature?

wangkuiyi (approved):
LGTM++

pkg/featurederivation/feature_derivation.go
typhoonzero: Sounds good, will do that in next PR.

PR #1533, opened 2019-12-25
Fix training can specify any label

### LINKED ISSUE #1499: Training will still start and successfully run when specified label not in the table
Training SQL:

```sql
SELECT * FROM housing.train
TO TRAIN DNNRegressor
WITH
  model.hidden_units=[128,32],
  train.epoch = 20,
  train.batch_size=4,
  validation.select="select * from housing.test"
LABEL class
INTO sqlflow_models.my_reg_model;
```

Yet the `LABEL` class is not in `housing.train` table, it should be `target`, but the above SQL can still run.

typhoonzero:
Fix https://github.com/sql-machine-learning/sqlflow/issues/1499

### DIFF
--- a/pkg/sql/executor_ir_test.go
+++ b/pkg/sql/executor_ir_test.go
@@ -128,7 +128,7 @@ WITH
 LABEL class
 INTO sqlflow_models.my_xgboost_model_by_program;
 
-SELECT sepal_length as sl, sepal_width as sw, class FROM iris.test
+SELECT sepal_length as sl, sepal_width as sw FROM iris.test
 TO PREDICT iris.predict.class
 USING sqlflow_models.my_xgboost_model_by_program;
 
@@ -138,7 +138,6 @@ USING TreeExplainer;
 `, modelDir, database.GetSessionFromTestingDB())
 		a.True(goodStream(stream.ReadAll()))
 	})
-
 }
 
 func TestExecuteXGBoostClassifier(t *testing.T) {
--- a/pkg/sql/executor_ir_test.go
+++ b/pkg/sql/executor_ir_test.go
@@ -128,7 +128,7 @@ WITH
 LABEL class
 INTO sqlflow_models.my_xgboost_model_by_program;
 
-SELECT sepal_length as sl, sepal_width as sw FROM iris.test
+SELECT sepal_length as sl, sepal_width as sw, class FROM iris.test
 TO PREDICT iris.predict.class
 USING sqlflow_models.my_xgboost_model_by_program;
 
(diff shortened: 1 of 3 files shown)


pkg/sql/ir_generator.go
```diff
@@ -127,6 +137,31 @@ func generateTrainStmtByModel(slct *parser.SQLFlowSelectStmt, connStr, cwd, mode
...
+		return e
+	}
+
+	for _, fc := range trainStmt.Features {
+		for _, field := range fc {
+			for _, fm := range field.GetFieldDesc() {
+				name := fm.Name
+				_, ok := trainFields.get(name)
+				if !ok {
+					return fmt.Errorf("Feature field not exist in database: %s", name)
```
shendiaomo: 1. Feature field **doesn't** exist
2. Error strings should not be capitalized https://github.com/golang/go/wiki/CodeReviewComments#error-strings

pkg/sql/ir_generator.go
```diff
@@ -127,6 +137,31 @@ func generateTrainStmtByModel(slct *parser.SQLFlowSelectStmt, connStr, cwd, mode
...
+				}
+			}
+		}
+	}
+	labelFieldName := trainStmt.Label.GetFieldDesc()[0].Name
+	_, ok := trainFields.get(labelFieldName)
+	if !ok {
+		return fmt.Errorf("Label field not exist in database: %s", labelFieldName)
```
shendiaomo: ditto

pkg/sql/executor_ir_test.go
```diff
@@ -128,7 +128,7 @@ WITH
 LABEL class
 INTO sqlflow_models.my_xgboost_model_by_program;
 
-SELECT sepal_length as sl, sepal_width as sw FROM iris.test
+SELECT sepal_length as sl, sepal_width as sw, class FROM iris.test
```
tonyyang-svail: Why updating this line?

pkg/sql/ir_generator.go
```diff
@@ -124,7 +136,38 @@ func generateTrainStmtByModel(slct *parser.SQLFlowSelectStmt, connStr, cwd, mode
 	if err != nil {
 		return nil, err
 	}
-	return generateTrainStmtWithInferredColumns(slctWithTrain, connStr)
+	return generateTrainStmtWithInferredColumns(slctWithTrain, connStr, false)
```
tonyyang-svail, line 169: Why we are not verifying the label field of the trainSQL using prediction?

pkg/sql/executor_ir_test.go
typhoonzero: Removing, it's for debug.

pkg/sql/ir_generator.go
typhoonzero: Predict SQL may not select the label column in the SELECT statement, we need only verify the feature columns only in that case.

shendiaomo (approved):

PR #1540, opened 2019-12-25
Fix load model twice when predict and analyze

### LINKED ISSUE #1240: Model table are loaded twice for prediction/explain job
The first time loading model table is at `generatePredictIR` to generate `PredIR.TrainIR`.
 
https://github.com/sql-machine-learning/sqlflow/blob/e907cada17489c650802884dafc233b1ed039a21/pkg/sql/ir_generator.go#L105

The second time loading model table is at `runPredIR` to recover the model directory.

https://github.com/sql-machine-learning/sqlflow/blob/e907cada17489c650802884dafc233b1ed039a21/pkg/sql/executor_ir.go#L264

It would be great to only load model table once. @typhoonzero What do you think?

The same situation also applies to explain jobs.

typhoonzero:
Fix https://github.com/sql-machine-learning/sqlflow/issues/1240, will move on to refactoring with `Request` in another PR.

### DIFF
--- a/pkg/sql/executor_ir.go
+++ b/pkg/sql/executor_ir.go
@@ -215,12 +215,13 @@ func runSQLProgram(wr *PipeWriter, sqlProgram string, db *database.DB, modelDir 
 	//
 	// The IR generation on the second statement would fail since it requires inspection the schema of some_table,
 	// which depends on the execution of create table some_table as (select ...);.
-	cwd, err := ioutil.TempDir("/tmp", "sqlflow_models")
-	if err != nil {
-		return err
-	}
-	defer os.RemoveAll(cwd)
+
 	for _, sql := range sqls {
+		cwd, err := ioutil.TempDir("/tmp", "sqlflow_models")
+		if err != nil {
+			return err
+		}
+
 		var r ir.SQLStatement
 		connStr := db.URL()
 		if parser.IsExtendedSyntax(sql) {
@@ -242,6 +243,10 @@ func runSQLProgram(wr *PipeWriter, sqlProgram string, db *database.DB, modelDir 
 		r.SetOriginalSQL(sql.Original)
 		if e := runSingleSQLIR(wr, r, db, modelDir, cwd, session); e != nil {
 			return e
+		}
+		err = os.RemoveAll(cwd)
+		if err != nil {
+			return err
 		}
 	}
 	return nil
--- a/pkg/sql/executor_ir.go
+++ b/pkg/sql/executor_ir.go
@@ -217,13 +217,18 @@ func runSQLProgram(wr *pipe.Writer, sqlProgram string, db *database.DB, modelDir
 	//
 	// The IR generation on the second statement would fail since it requires inspection the schema of some_table,
 	// which depends on the execution of create table some_table as (select ...);.
-
 	for _, sql := range sqls {
 		cwd, err := ioutil.TempDir("/tmp", "sqlflow_models")
 		if err != nil {
 			return err
 		}
-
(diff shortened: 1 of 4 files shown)


pkg/sql/executor_ir.go
```diff
@@ -244,10 +261,7 @@ func runSingleSQLIR(wr *PipeWriter, sqlIR ir.SQLStatement, db *database.DB, mode
 	}()
 	// TODO(typhoonzero): can run LogFeatureDerivationResult(wr, trainStmt) here to send
 	// feature derivation logs to client, yet we disable if for now so that it's less annoying.
-	if e := submitter().Setup(wr, db, modelDir, session); e != nil {
-		return e
-	}
-	defer submitter().Teardown()
+	submitter().Setup(wr, db, modelDir, cwd, session)
```
shendiaomo, line 281: It seems we need to define a named variable here.

pkg/sql/executor_ir.go
```diff
@@ -203,16 +215,21 @@ func runSQLProgram(wr *PipeWriter, sqlProgram string, db *database.DB, modelDir
 	//
 	// The IR generation on the second statement would fail since it requires inspection the schema of some_table,
 	// which depends on the execution of create table some_table as (select ...);.
+	cwd, err := ioutil.TempDir("/tmp", "sqlflow_models")
+	if err != nil {
+		return err
+	}
+	defer os.RemoveAll(cwd)
```
shendiaomo: Do all the statements in a SQL program share the same temporary dir?

pkg/sql/submitter.go
```diff
@@ -164,17 +160,16 @@ func (s *defaultSubmitter) ExecuteTrain(cl *ir.TrainStmt) (e error) {
...
-			var code string
-			if isXGBoostModel(cl.TrainStmt.Estimator) {
-				code, e = xgboost.Pred(cl, s.Session)
-			} else {
-				code, e = tensorflow.Pred(cl, s.Session)
-			}
-			if e == nil {
-				e = s.runCommand(code)
-			}
+	// NOTE(typhoonzero): model is already loaded under s.Cwd
```
shendiaomo, line 156: Do we need LoadModel anymore? How about deleting it?

pkg/sql/executor_ir.go
typhoonzero: Yes.

pkg/sql/submitter.go
typhoonzero: Done

pkg/sql/executor_ir.go
shendiaomo: If there are more than one `train` statements in the program, may the first one interfere with the second one? Will the checkpoints of the 1st `train` remains in the `model_save` directory?

tonyyang-svail: I agree with @shendiaomo. Each train/pred SQL should have exactly one working directory, and it is separate from other SQL's working directory.

typhoonzero: Thanks for this information, @shendiaomo @tonyyang-svail, done updating.

pkg/sql/executor_ir.go
```diff
@@ -243,6 +244,10 @@ func runSQLProgram(wr *PipeWriter, sqlProgram string, db *database.DB, modelDir
 		if e := runSingleSQLIR(wr, r, db, modelDir, cwd, session); e != nil {
 			return e
```
shendiaomo: It seems that we have to manually remove the temp dir in every return site in the loop 😓

shendiaomo (approved):
LGTM. It's great to remove the duplicated code.

PR #1541, opened 2019-12-25
Fix the 'unexpected truncation of the dataset' warning

### LINKED ISSUE #1539: We get the warning about truncated iterator in SQLFlow
image

shendiaomo:
Fix #1539 
The call sites to `cache` don't work actually.


typhoonzero (approved):
May need to add an issue we need to cache dataset properly.

PR #1547, opened 2019-12-26
Fix the 'Cannot infer num from shape <unknown>' error

### LINKED ISSUE #1546: SQLFlow can't run a BoostedTreesRegressor
SQLFlow supports xgboost like demonstrated in [xgboost tutorial](https://sql-machine-learning.github.io/sqlflow/doc/tutorial/housing-xgboost/). The following training statements works well.
```sql
SELECT * FROM boston.train
TO TRAIN xgboost.gbtree
WITH
    objective="reg:squarederror",
    train.num_boost_round = 30
COLUMN crim, zn, indus, chas, nox, rm, age, dis, rad, tax, ptratio, b, lstat
LABEL medv
INTO sqlflow_models.my_xgb_regression_model;
```
[BoostedTreesRegressor](https://www.tensorflow.org/api_docs/python/tf/estimator/BoostedTreesClassifier?version=stable) is the TensorFlow implementation of GBDT as a canned estimator. SQLFlow should support all canned estimators. 
However, if we change the `xgboost.gbtree` in the above statement  to `BoostedTreesRegressor` like:

```sql
SELECT * FROM boston.train
TO TRAIN BoostedTreesRegressor
WITH model.n_batches_per_layer=300
COLUMN crim, zn, indus, chas, nox, rm, age, dis, rad, tax, ptratio, b, lstat
LABEL medv
INTO sqlflow_models.my_xgb_regression_model;
```

The following error will occur:
image

I believe this is a bug in shapes of features because `BoostedTreesEstimator` has to infer shapes from inputs, but SQLFlow didn't set shapes in the dataset.
https://github.com/sql-machine-learning/sqlflow/blob/f10c001a49e8657932c350d5460cb1e8f3b060e2/python/sqlflow_submitter/tensorflow/input_fn.py#L49
As we've seen, the above code doesn't provide the `shape` argument to `from_generator`, the shapes of input features are all `<unknown>`, that's the reason for the error 
```
ValueError: Cannot infer num from shape <unknown>

```

shendiaomo:
Fix #1546 
The fix provides an appropriate argument to the `shape` parameter of `from_generator` in `input_fn.py`.
This change affects all that depends on the input dataset, we have to modify every place to stack/flatten the input features.

### DIFF
--- a/python/sqlflow_submitter/tensorflow/train.py
+++ b/python/sqlflow_submitter/tensorflow/train.py
@@ -74,7 +74,7 @@ def keras_train_and_save(estimator, model_params, save,
         loss=classifier_pkg.loss,
         metrics=keras_metrics)
     if hasattr(classifier, 'sqlflow_train_loop'):
-        def flatten(feature, label):  # TODO(shendiaomo): This is temporary
+        def flatten(feature, label):  # TODO(shendiaomo): Modify the cluster model to adapt the new input structure
             for k in feature:
                 feature[k] = feature[k][0]
             return feature, [label]
--- a/python/sqlflow_submitter/tensorflow/train.py
+++ b/python/sqlflow_submitter/tensorflow/train.py
@@ -74,7 +74,11 @@ def keras_train_and_save(estimator, model_params, save,
         loss=classifier_pkg.loss,
         metrics=keras_metrics)
     if hasattr(classifier, 'sqlflow_train_loop'):
-        classifier.sqlflow_train_loop(train_dataset)
+        def flatten(feature, label):  # TODO(shendiaomo): This is temporary
+            for k in feature:
+                feature[k] = feature[k][0]
+            return feature, [label]
+        classifier.sqlflow_train_loop(train_dataset.map(flatten))
     else:
         if label_meta["feature_name"] != "":
             history = classifier.fit(train_dataset,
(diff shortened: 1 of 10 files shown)


python/sqlflow_submitter/tensorflow/train.py
```diff
@@ -74,7 +74,11 @@ def keras_train_and_save(estimator, model_params, save,
         loss=classifier_pkg.loss,
         metrics=keras_metrics)
     if hasattr(classifier, 'sqlflow_train_loop'):
-        classifier.sqlflow_train_loop(train_dataset)
+        def flatten(feature, label):  # TODO(shendiaomo): This is temporary
```
typhoonzero: TODO comment needs to be followed by an explanation of what there is to do: http://google.github.io/styleguide/pyguide.html#312-todo-comments

typhoonzero (commented):
LGTM

python/sqlflow_submitter/tensorflow/train.py
shendiaomo: Done.

typhoonzero (approved):

tonyyang-svail:
@shendiaomo Looks like this PR is causing our CI failing on MaxCompute. https://travis-ci.com/sql-machine-learning/sqlflow/jobs/270622191

Issue #1435, opened 2019-12-16
[Refactor] remove Datasource in IR


typhoonzero:
`Datasource string` is currently also available in `Session`, so we can remove it now.

PR #1553, opened 2019-12-26
Use request struct to pass execution context

typhoonzero:
Related: https://github.com/sql-machine-learning/sqlflow/issues/1435


wangkuiyi:
I see this PR wrapped three parameters up into a new type `RequestContext`.

I understand that we want an execution context. So, maybe `pkg/step.Context`.

Here we are talking about the target directory structure.  I think the new structure could be derived from that SQLFlow will have two levels of execution processes.

1. The SQLFlow server takes a SQL program, parse it into IR, and generate a workflow to run by Arog/Tekton, and
1. each step of the workflow 
   1. in the near term, call `cmd/step`, which in turn calls `pkg/step`, to execute a SQL statement, and
   1. in the long run, it becomes a Couler/Fluid function.

So, what we need are

```
cmd/{sqlflowserver, step}
```

and 

```
pkg/
    parser/ -- convert SQL program into IR
    ir/
    workflow/ -- convert IR into a Couler/Fluid program and then an Argo/Tekton YAML
        argo/ -- submit and watch Argo/Tekton jobs
        tekton/ --- submit and watch Tekton jobs
    step/ -- called by cmd/step to interpret and run a SQL statement
        features/ -- feature derivation in Go. In the long run, this will be replaced by Python code
        codegen/ -- converts a SQL statement  into a Python submitter
```

typhoonzero:
Closing due to: https://github.com/sql-machine-learning/sqlflow/issues/1585

PR #1557, opened 2019-12-27
Make loops more pythonic

shendiaomo:


python/sqlflow_submitter/tensorflow/predict.py
```diff
@@ -71,22 +71,17 @@ def eval_input_fn(batch_size, cache=False):
...
     classifier.load_weights(save)
-    del pred_dataset
     pred_dataset = eval_input_fn(1, cache=True).make_one_shot_iterator()
     buff_rows = []
     column_names = feature_column_names[:]
     column_names.append(label_meta["feature_name"])
     with buffered_db_writer(conn.driver, conn, result_table, column_names, 100, hdfs_namenode_addr, hive_location, hdfs_user, hdfs_pass) as w:
-        while True:
-            try:
-                features = pred_dataset.get_next()
```
tonyyang-svail, line 87: @typhoonzero  I am curious why the original Python code is written in this way.

typhoonzero: Maybe I just kept the testing code when I trying to get the feature sample as well as predict results, anyway this is a cool fix.

typhoonzero (approved):

PR #1560, opened 2019-12-27
change the 'Analyze' to 'EXPLAIN' in repl 

### LINKED ISSUE #1559: change the 'Analyze' to 'EXPLAIN' in repl
In repl mode, the default inference keyword `Analyze` should be changed to `EXPLAIN`

Echo9573:
fix #1559

### DIFF
--- a/cmd/repl/prompt.go
+++ b/cmd/repl/prompt.go
@@ -35,12 +35,13 @@ var defaultSuggestions = []prompt.Suggest{
 	{"USING", ""},
 	{"WITH", ""},
 	{"TO", ""},
+	{"LABEL", ""},
 }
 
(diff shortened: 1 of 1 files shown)


cmd/repl/prompt.go
```diff
@@ -35,12 +35,13 @@ var defaultSuggestions = []prompt.Suggest{
 	{"USING", ""},
 	{"WITH", ""},
 	{"TO", ""},
+	{"LABEL", ""},
```
shendiaomo: We don't need `LABEL` here because it can only appear in a training clause.

shendiaomo (approved):

tonyyang-svail (approved):
LGTM++

PR #1513, opened 2019-12-22
Remove testing database connection URL literals all over the codebase

wangkuiyi:
Fix https://github.com/sql-machine-learning/sqlflow/issues/1496

Fix https://github.com/sql-machine-learning/sqlflow/issues/1519

PR #1576, opened 2019-12-30
Rename MAXCOMPUTE_AK into SQLFLOW_TEST_DB_MAXCOMPUTE_AK

### LINKED ISSUE #1575: Rename env variables with prefix MAXCOMPUTE_
If we search for `MAXCOMPUTE_ASK`, we would see [many environment variables](https://github.com/sql-machine-learning/sqlflow/search?q=MAXCOMPUTE_AK&unscoped_q=MAXCOMPUTE_AK), which are all for testing purpose.

To prevent users from the misunderstanding that these environment variables are for passing in MaxCompute production settings to SQLFlow, I think it should add them the prefix `SQLFLOW_TEST_DB_`.

wangkuiyi:
Fix https://github.com/sql-machine-learning/sqlflow/issues/1575

This is part of https://github.com/sql-machine-learning/sqlflow/pull/1513, which is too big to debug and review.

For the only [Python source code file](https://github.com/sql-machine-learning/sqlflow/pull/1576/files#diff-ac9db26862b19f3cf387d9e102208e7a) involved in, this PR did some pretty-printing work, which is a little bit off-topic due to my Emacs settings.

### DIFF
--- a/pkg/sql/template_elasticdl.go
+++ b/pkg/sql/template_elasticdl.go
@@ -107,16 +107,16 @@ class PredictionOutputsProcessor(BasePredictionOutputsProcessor):
         if all(
             k in os.environ
             for k in (
-                "MAXCOMPUTE_PROJECT",
-                "MAXCOMPUTE_AK",
-                "MAXCOMPUTE_SK",
+                "SQLFLOW_TEST_DB_MAXCOMPUTE_PROJECT",
+                "SQLFLOW_TEST_DB_MAXCOMPUTE_AK",
+                "SQLFLOW_TEST_DB_MAXCOMPUTE_SK",
             )
         ):
             self.odps_writer = ODPSWriter(
-                os.environ["MAXCOMPUTE_PROJECT"],
-                os.environ["MAXCOMPUTE_AK"],
-                os.environ["MAXCOMPUTE_SK"],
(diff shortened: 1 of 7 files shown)


terrytangyuan (dismissed):

typhoonzero (dismissed):

pkg/sql/template_elasticdl.go
```diff
@@ -107,16 +107,16 @@ class PredictionOutputsProcessor(BasePredictionOutputsProcessor):
         if all(
             k in os.environ
             for k in (
-                "MAXCOMPUTE_PROJECT",
-                "MAXCOMPUTE_AK",
-                "MAXCOMPUTE_SK",
+                "SQLFLOW_TEST_DB_MAXCOMPUTE_PROJECT",
```
shendiaomo: It seems these are not for test, is it?

wangkuiyi: Great catch!

wangkuiyi: @terrytangyuan Should I revert these changes in `template_elasticdl.go`?

terrytangyuan: Yes, revert them for now. We also need to change the namings in ElasticDL accordingly in order to make this codegen work.

terrytangyuan (approved):

PR #1577, opened 2019-12-30
TensorFlow boosting trees explain

shendiaomo:
This implementation is based on https://www.tensorflow.org/tutorials/estimator/boosted_trees_model_understanding
After we train a BoostedTreesClassifier on titanic:
```sql
SELECT * FROM titanic.train TO TRAIN BoostedTreesClassifier 
    WITH model.n_batches_per_layer=1, 
               model.n_trees=50, 
               model.max_depth=3,
               model.center_bias=True,
               train.epoch=10,
               train.batch_size=100
    LABEL survived 
    INTO sqlflow_models.my_bt_model;
```
The output looks like:
```
{'accuracy': 0.8473625, 'accuracy_baseline': 0.6161616, 'auc': 0.89060915, 'auc_precision_recall': 0.86166745, 'average_loss': 0.38625988, 'label/mean': 0.3838384, 'loss': 0.38568884, 'precision': 0.821875, 'prediction/mean': 0.3797143, 'recall': 0.76900584, 'global_step': 89}
```

This commit supports two kinds of figures
1. bar (same to the original shap bar figure)
    ```sql
    SELECT * FROM titanic.train TO EXPLAIN sqlflow_models.my_bt_model USING TreeExplainer
    ```
    This will output 
    image

2. violin (similar to the original shap dot figure) 
    ```sql
    SELECT * FROM titanic.train TO EXPLAIN sqlflow_models.my_bt_model WITH summary.plot_type=violin USING TreeExplainer;
    ```
    will output
    image

Hope this may inspire the follow-up work on PAI.

### DIFF
--- a/pkg/sql/codegen/tensorflow/template_analyze.go
+++ b/pkg/sql/codegen/tensorflow/template_analyze.go
@@ -1,0 +1,83 @@
+// Copyright 2019 The SQLFlow Authors. All rights reserved.
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package tensorflow
+
+import "sqlflow.org/sqlflow/pkg/sql/ir"
+
+type analyzeFiller struct {
+	DataSource    string
+	Select        string
+	SummaryParams string
+	// below members comes from trainStmt
+	EstimatorClass    string
+	IsKerasModel      bool
+	FieldDescs        []*ir.FieldDesc
+	FeatureColumnCode string
+	Y                 *ir.FieldDesc
+	ModelParams       map[string]interface{}
+	Save              string
+	IsPAI             bool
+}
+
+const boostedTreesAnalyzeTemplateText = `
+from sqlflow_submitter.tensorflow import analyze
+import tensorflow as tf
+try:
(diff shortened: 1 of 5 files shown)


python/sqlflow_submitter/tensorflow/analyze.py
```diff
@@ -0,0 +1,145 @@
...
+
+# Disable Tensorflow INFO and WARNING logs
+if TF_VERSION_2:
+    import logging
+    tf.get_logger().setLevel(logging.ERROR)
+else:
+    tf.logging.set_verbosity(tf.logging.ERROR)
+    from .pai_distributed import define_tf_flags
+
+def analyze(datasource, estimator_cls, select, feature_columns, feature_column_names,
```
wangkuiyi, line 46: In https://github.com/sql-machine-learning/sqlflow/pull/1587, I am renaming Analyze into Explain. How about we rename this function `explain`?

pkg/sql/codegen/tensorflow/template_analyze.go
```diff
@@ -0,0 +1,83 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package tensorflow
+
+import "sqlflow.org/sqlflow/pkg/sql/ir"
+
+type analyzeFiller struct {
```
wangkuiyi, line 18: How about renaming analyzeFiller into explainFiller?

wangkuiyi (approved):

python/sqlflow_submitter/tensorflow/analyze.py
shendiaomo: I'll fix it in the next PR.

pkg/sql/codegen/tensorflow/template_analyze.go
shendiaomo: I'll fix it in the next PR.

PR #1592, opened 2019-12-31
reduce workflow test epochs

typhoonzero:


cmd/sqlflowserver/main_test.go
```diff
@@ -1375,7 +1375,8 @@ func checkWorkflow(ctx context.Context, cli pb.SQLFlowClient, stream pb.SQLFlow_
 	req := &pb.FetchRequest{
 		Job: &pb.Job{Id: workflowID},
 	}
-	for i := 0; i < 120; i++ {
+	// wait 30min for the workflow execution since it may take time to allocate enough nodes.
```
wangkuiyi, line 1378: It seems that each step of this loop waits for 3 seconds. So, 1800 steps take 90 mins, not 30 mins. Should we further reduce the number of steps here down to 120/3 = 40?

wangkuiyi (approved):

PR #1598, opened 2020-01-01
Try upgrade base image from Ubuntu 16.04 to 18.04

wangkuiyi:
Currently, we use Miniconda to install Python 3.6 in the Docker image.  This seems redundant as containers are virtualenvs. Before I try to remove Miniconda, I want to make sure that apt-get could install Python 3.6. However, the default Python 3 version on Ubuntu 16.04 is 3.5. So, I am trying to upgrade the base image to Ubuntu 18.04, whose default Python 3 version is 3.6.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:16.04
+FROM ubuntu:18.04
 
 # The default source archive.ubuntu.com is busy and slow. We use the following source makes docker build running faster.
 RUN echo '\n\
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,6 +9,8 @@ deb http://us.archive.ubuntu.com/ubuntu/ xenial-proposed main restricted univers
 deb http://us.archive.ubuntu.com/ubuntu/ xenial-backports main restricted universe multiverse \n\
 ' > /etc/apt/sources.list
 
+RUN apt-get update
+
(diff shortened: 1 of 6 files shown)


wangkuiyi:
Installing git on Ubuntu 18.04 got an error during CI and during local docker build. After Googling, and adding `apt-get upgrade && apt-get autoclean`, still I cannot fix it.

```
Step 6/42 : RUN /install-download-tools.bash
 ---> Running in a727b74eb4de
Reading package lists...
Building dependency tree...
Reading state information...
bzip2 is already the newest version (1.0.6-8.1ubuntu0.2).
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 git : Depends: liberror-perl but it is not going to be installed
E: Unable to correct problems, you have held broken packages.
The command '/bin/sh -c /install-download-tools.bash' returned a non-zero code: 100
```

wangkuiyi:
The above error was due to that I didn't change the code name xenial to bionic in the following code snippet.

https://github.com/sql-machine-learning/sqlflow/blob/2982b8390b7329c3c73e16c13be05e4ed4a1115e/Dockerfile#L4-L10

Retry upgrading Ubuntu in #1600

PR #1600, opened 2020-01-01
Retry upgrade Ubuntu version

wangkuiyi:
This follows up https://github.com/sql-machine-learning/sqlflow/pull/1598

A direct necessity of upgrading Ubuntu is to use modern Python. Applications like SHAP requires Python >=3.6, which is not officially supported on Ubuntu 16.04.  Miniconda could build from source code with patches, but Miniconda is not a reasonable dependency when we are using Docker.

### DIFF
--- a/scripts/docker/install-download-tools.bash
+++ b/scripts/docker/install-download-tools.bash
@@ -15,5 +15,4 @@
 
 set -e
 
-apt-get update
-apt-get install -y curl wget bzip2 unzip git
+apt-get install -y curl wget unzip git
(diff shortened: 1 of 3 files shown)


scripts/docker/install-download-tools.bash
```diff
@@ -15,5 +15,4 @@
 
 set -e
 
-apt-get update
-apt-get install -y curl wget bzip2 unzip git
+apt-get install -y curl wget unzip git
```
wangkuiyi, line 18: bzip2 is in Ubuntu 18.04 by default.

weiguoz:
Shall we upgrade the version in the `scripts/tensorflow_step_image/Dockerfile`?

weiguoz (approved):
LGTM

typhoonzero:
I tried to use `python:3.7-buster` as the base image but since the "python" images use debian as it's base, it's hard to install mysql so I just gave up trying.

PR #1616, opened 2020-01-04
Move feature derivation

wangkuiyi:
- Move package `pkg/featurederivation` to `pkg/step/feature`
- Move files `pkg/featurederivation/feature_derivation*.go` to `pkg/step/feature/derivation*.go`
- Rename type `featurederivation.FeatureColumnMap` to `feature.ColumnMap`
- Rename func `featurederivation.LogFeatureDerivationResult` to `feature.LogDerivationResult`.


pkg/step/feature/derivation.go
```diff
@@ -386,7 +386,7 @@ func InferFeatureColumns(trainStmt *ir.TrainStmt, dataSource string) error {
 
 	// set back trainStmt.Features in the order of select
 	for _, target := range columnTargets {
-		targetFeatureColumnMap := fcMap[target]
+		targetColumnMap := fcMap[target]
```
typhoonzero: The original name `targetFeatureColumnMap` refers to Tensorflow feature column, maybe it's better than `ColumnMap`?

typhoonzero (dismissed):
LGTM++ Thanks

pkg/step/feature/derivation.go
wangkuiyi: I see. I will follow your suggestion and change it back

typhoonzero (approved):

PR #1624, opened 2020-01-06
Support write explain result (dfc) BoostedTreesEstimator

### LINKED ISSUE #1612: Support write model explain values into a result table
The current `SELECT * from pred_table TO EXPLAIN some_model` will generate an image showing the feature importance. For cases that in a full-text terminal or distributed platform (explanation data will be generated on different nodes), we can not generate the image directly and display it. In that case, we may need to write the explanation data to a table, and use tools to display the image:

```
SELECT * from pred_table
TO EXPLAIN some_model
WITH attribute=value
USING explainer
INTO some_result_table;
```

Note that the `INTO` clause is optional, if not specified `INTO` we will try directly display the image, if `INTO` is specified, we will write the explanation data into that table.

typhoonzero:
Fix https://github.com/sql-machine-learning/sqlflow/issues/1612

### DIFF
--- a/pkg/sql/codegen/tensorflow/template_explain.go
+++ b/pkg/sql/codegen/tensorflow/template_explain.go
@@ -28,6 +28,11 @@ type explainFiller struct {
 	ModelParams       map[string]interface{}
 	Save              string
 	IsPAI             bool
+	ResultTable       string
+	HDFSNameNodeAddr  string
+	HiveLocation      string
+	HDFSUser          string
+	HDFSPass          string
 }
 
 const boostedTreesExplainTemplateText = `
@@ -79,5 +84,11 @@ explain.explain(datasource="{{.DataSource}}",
                 feature_metas=feature_metas,
                 label_meta=label_meta,
                 model_params=model_params,
-                save="{{.Save}}", **summaryAttrs)
+                save="{{.Save}}",
+                result_table="{{.ResultTable}}",
+                hdfs_namenode_addr="{{.HDFSNameNodeAddr}}",
+                hive_location="{{.HiveLocation}}",
+                hdfs_user="{{.HDFSUser}}",
+                hdfs_pass="{{.HDFSPass}}",
+                **summaryAttrs)
 `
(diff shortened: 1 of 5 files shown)


tonyyang-svail:
@typhoonzero What does DFC stand for?

typhoonzero:
directional feature contributions. See https://www.tensorflow.org/tutorials/estimator/boosted_trees_model_understanding for details

python/sqlflow_submitter/tensorflow/explain.py
```diff
@@ -68,12 +70,40 @@ def _input_fn():
...
+        cursor.close()
+
+def write_dfc_result(dfc_mean, result_table, conn,
+                     feature_column_names,
+                     hdfs_namenode_addr, hive_location,
+                     hdfs_user, hdfs_pass):
+    with buffered_db_writer(conn.driver, conn, result_table, ["feature", "dfc"], 100, hdfs_namenode_addr, hive_location, hdfs_user, hdfs_pass) as w:
+        for row_name in feature_column_names:
+            w.write([row_name, dfc_mean.loc[row_name]])
+            print(row_name, dfc_mean.loc[row_name])
```
tonyyang-svail: clean up?

tonyyang-svail (dismissed):

pkg/sql/codegen/tensorflow/template_explain.go
```diff
@@ -79,5 +84,11 @@ explain.explain(datasource="{{.DataSource}}",
                 feature_metas=feature_metas,
                 label_meta=label_meta,
                 model_params=model_params,
-                save="{{.Save}}", **summaryAttrs)
+                save="{{.Save}}",
+                result_table="{{.ResultTable}}",
+                hdfs_namenode_addr="{{.HDFSNameNodeAddr}}",
+                hive_location="{{.HiveLocation}}",
+                hdfs_user="{{.HDFSUser}}",
+                hdfs_pass="{{.HDFSPass}}",
```
tonyyang-svail, line 92: Is there a way to avoid passing `hdfs_` in all subsequent calls? Is it a good idea to wrap `hive/hdfs*` in a single struct?

typhoonzero: Will try to refactor this in incoming PRs, also need to be done in `predict.py`

tonyyang-svail (approved):

PR #1626, opened 2020-01-06
Remove call to SplitMultipleSQL

### LINKED ISSUE #1430: Can we remove pkg/sql/sql_splitter.go now?
It seems that after the introduction of the new general parser, we are no longer relying on `splitExtendedSQL`. Can we remove it?

```
$ git grep splitExtendedSQL
pkg/sql/sql_splitter.go:// splitExtendedSQL splits an extended select statement into
pkg/sql/sql_splitter.go:func splitExtendedSQL(slct string) ([]string, error) {
pkg/sql/sql_splitter_test.go:   s, err := splitExtendedSQL(`select a train b with c;`)
pkg/sql/sql_splitter_test.go:   s, err = splitExtendedSQL(`  select a predict b using c;`)
pkg/sql/sql_splitter_test.go:   s, err = splitExtendedSQL(` select a from b;`)
pkg/sql/sql_splitter_test.go:   s, err = splitExtendedSQL(`train a with b;`)
```

tonyyang-svail:
Fix #1430.


pkg/parser/sql_counter.go
```diff
@@ -0,0 +1,23 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package parser
+
+// NumberOfStatements returns the number of SQL statement given a SQL program
+func NumberOfStatements(dialect, statements string) (int, error) {
```
wangkuiyi: I am afraid we might not want to create a new file to host this function.  Also, I am not sure if we need this function; it's used only twice.  Adding this function creates more lines of code than not to have it.

pkg/parser/sqlflow_parser.go
```diff
@@ -123,6 +123,24 @@ func thirdPartyParse(dialect, program string) ([]*SQLFlowStmt, int, error) {
 	return spr, i, nil
 }
 
+func findNextSQLStatement(statements string) (string, error) {
```
wangkuiyi: Can we don't skip over the parser and call the lexer directly?  In my previous PRs, I made the `parseSQLStatement` in extended_syntax.y able to return index as the external parsers. I think we can use the ability to get rid of the necessity to call the lexer directly.

wangkuiyi (changes requested):
Let us try not to look over the parser and calling the lexer directly.

pkg/parser/sqlflow_parser.go
tonyyang-svail: I was trying to use the return index from extended_syntax.y but failed at the first attempt.

Consider we parsing `to train with ... label l into some_table; select * from some_other_table`. The extended syntax parser will return an error indexed at the character `s` of `select`. So we should try to parse the string before the error index (i.e. `to train with ... label l into some_table; `), and this time the parsing is successful. To put this logic into code, we have

```go
pr, idx, err := parseSQLFlowStmt(statements)

if err != nil {
    pr, _, err := parseSQLFlowStmt(statements[:idx])
    if err != nil {
        return nil, err
    }
    return pr, err
}

return pr, err
```

However, this logic fails on cases like `to train a with a = b label c into d select * from e`. Our parser won't be raising error while it should be.

tonyyang-svail: Update:

One solution I can think of is to make the extended syntax parser incapable of recognizing `;`, meaning parsing `to train a with a = b label c into d;` will return an error at `;` (this behavior is consistent with Hive parser and Calcite parser).

So when parsing `to train a with a = b label c into d; select * from e;`, the extended syntax parser will return an error indexed at the character `;`. So we check if the index is pointed at `;`, if so, we should try to parse the string before the error-index (i.e. `to train a with a = b label c into d`), and this time the parsing is successful. To put this logic into code, we have

```go
pr, idx, err := parseSQLFlowStmt(statements)

if statements[idx] == ';' {
    pr, _, err := parseSQLFlowStmt(statements[:idx])
    if err != nil {
        return nil, err
    }
    return pr, err
}

return pr, err
``` 

And this logic will also capture incorrect statements like `to train a with a = b label c into d select * from e`.

wangkuiyi: If we expect our parser errors with `to train a with a = b label c into d select * from e` as we want a `;` between `into d` and `select`, we can simply change the rule `end_of_stmt` to must have a `;` from that `;` is optional.

wangkuiyi (approved):
LGTM++

wangkuiyi:
Just want to make sure that you will fix this TODO https://github.com/sql-machine-learning/sqlflow/pull/1626/files#diff-b4455f2ddb298f6a19eb306415ca19c1R176-R177 soon after merging this PR, right?

typhoonzero (approved):
LGTM++

tonyyang-svail:
> Just want to make sure that you will fix this TODO https://github.com/sql-machine-learning/sqlflow/pull/1626/files#diff-b4455f2ddb298f6a19eb306415ca19c1R176-R177 soon after merging this PR, right?

Yes, I will do it today.

pkg/parser/sqlflow_parser.go
tonyyang-svail: @wangkuiyi originally mentioned the following, I am the one writing it down.

"
The design flaw of https://github.com/sql-machine-learning/sqlflow/pull/1626#discussion_r363575099 is it assumes upper-level logic knows the implementation details of the lower-level.

For a parser, there are four levels of logic:
1. syntax - parser
1. lexical - lexer
1. rune - Unicode
1. bytes/characters

While https://github.com/sql-machine-learning/sqlflow/pull/1626#discussion_r363575099 is suggesting code on level 1 access the logic of level 4.
"

PR #1627, opened 2020-01-06
Update models

typhoonzero:
Update models to include latest models.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -100,7 +100,7 @@ ARG WITH_SQLFLOW_MODELS="ON"
 RUN if [ "${WITH_SQLFLOW_MODELS:-ON}" = "ON" ]; then \
   git clone https://github.com/sql-machine-learning/models.git && \
   cd models && \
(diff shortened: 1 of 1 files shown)


Yancey0623 (approved):
LGTM

Dockerfile
```diff
@@ -100,7 +100,7 @@ ARG WITH_SQLFLOW_MODELS="ON"
 RUN if [ "${WITH_SQLFLOW_MODELS:-ON}" = "ON" ]; then \
   git clone https://github.com/sql-machine-learning/models.git && \
   cd models && \
-  git checkout 4af6f567ba2dfda57a99d7a5985bfe11314582db && \
+  git checkout 333e04da34aa518b0d798f1898be236694d4d42d && \
```
wangkuiyi, line 103: Should we tag this commit and refer to it by the tag name?

wangkuiyi (approved):

Dockerfile
typhoonzero: We'll move on to use the model Docker image instead of installing it, so just keep it when model zoo is done.

PR #1638, opened 2020-01-07
Remove legacy parse

tonyyang-svail:


pkg/database/db.go
```diff
@@ -33,6 +33,11 @@ type DB struct {
 	*sql.DB
 }
 
+// ConnectionString returns the connection string in driver_type://data_source_name format
```
wangkuiyi: This function already exists, named URL()

pkg/verifier/verifier_test.go
```diff
@@ -22,40 +23,53 @@ import (
 	"sqlflow.org/sqlflow/pkg/parser"
 )
 
+func parseOneStatement(statement string) (*parser.SQLFlowSelectStmt, error) {
```
wangkuiyi: Duplicated definition as in ir_generator_test.go

pkg/sql/ir_generator_test.go
```diff
@@ -25,11 +26,21 @@ import (
 	"sqlflow.org/sqlflow/pkg/sql/ir"
 )
 
+func parseOneStatement(statement string) (*parser.SQLFlowSelectStmt, error) {
```
wangkuiyi: This function duplicates with function `ParseOneStatement` in `parser.go`.

wangkuiyi (commented):
Thanks to verifying that we could remove LegacyParse and replace it by parseOneStatement.

Two suggestions:

- `ParseOneStatement` is already there in `parser.go`; there is no need to add them, duplicated in `ir_generator_test.go` and `verifier_test.go`.
- `ParseOneStatement` doesn't parse a statement, but parse a program with one and only one statement. Given that a set with only one element is known as a **unit set**. How about we rename it into `ParseUnitProgram`?

pkg/sql/ir_generator_test.go
tonyyang-svail: Thanks for reminding me.

pkg/database/db.go
tonyyang-svail: Thanks for pointing it out!

tonyyang-svail:
> ParseOneStatement doesn't parse a statement, but parse a program with one and only one statement. Given that a set with only one element is known as a unit set. How about we rename it into ParseUnitProgram?

Sounds good. I will do that in the next PR.

pkg/verifier/verifier_test.go
tonyyang-svail: Removed.

typhoonzero (approved):

PR #1645, opened 2020-01-07
Do not use init to register submitters

typhoonzero:
Since `submitterRegistry` is a statically initialized global, init the values in one line is much easier to know which "submitters" we are supporting.


tonyyang-svail:
Hi @typhoonzero, could you please mention why it is bad to register submitter in the `init()`?

typhoonzero:
> Hi @typhoonzero, could you please mention why it is bad to register submitter in the init()?

Done

shendiaomo (approved):

PR #1657, opened 2020-01-08
Add set -e in CI script

tonyyang-svail:
Add set -e in the CI script so that the script will stop at the exact failed step.


Yancey0623 (approved):
LGTM

typhoonzero (approved):
Should have been done a long time ago.

PR #1665, opened 2020-01-09
[Parser] More accurate error message

tonyyang-svail:
The parser error idx is derived from `lexer.previous`, so should the error message.

For example, when parsing `to train a select b from c`, the error message should include the `select` keyword.

Before: `syntax error: at (17 ~ 17)-th runes near " b from c;"`

After: `syntax error: at (11 ~ 17)-th runes near "select b f"`


typhoonzero (approved):
LGTM

PR #1666, opened 2020-01-09
Replace the IsKeras function with python issubclass to make code clean

shendiaomo:
1. Remove the `IsKeras` function in golang and the `is_keras_model` parameter in python
1. Tuning indentation to be Google-style conformable
1. Several mini changes to be pythonic
1. Please check the `Hide whitespace changes` box to review
image


python/sqlflow_submitter/tensorflow/train.py
```diff
@@ -182,16 +181,12 @@ def train(is_keras_model,
...
-        if verbose == 1:
-            tf.get_logger().setLevel((4-verbose) * 10)  # logging.INFO levels range from 10~40
-    else:
-        if TF_VERSION_2:
-                tf.get_logger().setLevel((4-verbose) * 10)
-        else:
-            if verbose >= 2:
-                tf.logging.set_verbosity(tf.logging.INFO)
+    assert 0 <= verbose <= 3
+    is_keras_model = not issubclass(estimator, (tf.estimator.Estimator, tf.estimator.BoostedTreesEstimator))
```
typhoonzero: Should call it `is_premade_estimator`.  A model is Keras model if it's not an `Estimator` does not make sense, even though we only support Keras and Estimators

shendiaomo: `is_estimator` is better, there may be some models in `sqlflow_models` that implement the estimator interface

typhoonzero (approved):
Cool

Issue #1389, opened 2019-12-10
The third-party parser starts a Java process every time it parses a SQL program


wangkuiyi:
https://github.com/sql-machine-learning/sqlflow/blob/5bf232b52f02f8bfc72862044dc699a51ad6a94c/pkg/sql/tpp/parser.go#L120-L125

It is expensive to start a Java process.  This current approach would slow down the QPS of the SQLFlow server.  We will need to change the third-party server into gRPC servers.

wangkuiyi:
The following three unit tests that run the same set of cases show extremely slow parsing speed of calling Java programs.

```
=== RUN   TestParseWithMySQL
--- PASS: TestParseWithMySQL (0.01s)
=== RUN   TestParseWithHive
--- PASS: TestParseWithHive (44.52s)
=== RUN   TestParseWithMaxCompute
--- FAIL: TestParseWithMaxCompute (35.77s)
```

The parsing time cost of calling the HiveQL Java command-line parser is **445 times** of that of calling the TiDB parser.

tonyyang-svail:
TODO list:

Java Server
- [x] Add Java gRPC skeleton code for the parser. #1673
- [x] Update Java parser building doc. #1788
- [x] Integrate Java gRPC server with Calcite/HiveQL parser. #1790

Go Client
- [x] Integrate Go gRPC client with Java gRPC parser.

Profiling
- [x] Profile gRPC vs cmd

PR #1673, opened 2020-01-09
Add Java gRPC skeleton code for the parser

tonyyang-svail:
Part of solving https://github.com/sql-machine-learning/sqlflow/issues/1389.


java/parser/src/main/proto/Parser.proto
```diff
@@ -0,0 +1,42 @@
...
+  // The sql_program represents a SQL program that contains possible more than
+  // one SQL statements. For example, `select * from my_table; select * from another_table to train ...`.
+  string sql_program = 2;
+}
+
+message ParserResponse {
+  // The sql_statements represent part of the sql_program accepted by the parser.
+  // For example, if the ParserRequest is `select 1; select 2 to train ...`.
+  // The ParserResponse contains two statements, `select 1;` and `select 2 `
+  repeated string sql_statements = 1;
```
typhoonzero, line 32: Do we need to return a struct like:

https://github.com/sql-machine-learning/sqlflow/blob/129be40d6e2627a30e7f06aaa8604d872542b6c4/pkg/parser/sqlflow_parser.go#L28-32

tonyyang-svail: Yes, the API should be consistent with 
https://github.com/sql-machine-learning/sqlflow/blob/129be40d6e2627a30e7f06aaa8604d872542b6c4/pkg/parser/sqlflow_parser.go#L113

Dockerfile
```diff
@@ -87,6 +87,7 @@ mv $GOPATH/bin/sqlflowserver /usr/local/bin && \
 mv $GOPATH/bin/repl /usr/local/bin && \
 (cd python/couler && python setup.py -q install) && \
 cd java/parser && \
+protoc --java_out=src/main/java --grpc-java_out=src/main/java/ --proto_path=src/main/proto/ src/main/proto/Parser.proto && \
```
wangkuiyi, line 90: Is this line duplicated with the newly added line to test/java.sh?

wangkuiyi: If so, it seems that we only need the line here, but not test/java.sh?

wangkuiyi (approved):

Dockerfile
tonyyang-svail: > Is this line duplicated with the newly added line to test/java.sh?

They are not duplicated. The line here is used for building the .jar file in the release image. The line in test/java.sh is used for testing.

This logic is consistent with having two `go generate ./...`s. One in Dockerfile, one in `scripts/test/units.sh`.

PR #1680, opened 2020-01-10
Add a unit test for  NAS Classifier

shendiaomo:


typhoonzero:
@shendiaomo can you please also add a test case under models repo?

shendiaomo:
> @shendiaomo can you please also add a test case under models repo?

It will be done later because the test framework in models repo only works for Keras models at the moment.

shendiaomo:
> > @shendiaomo can you please also add a test case under models repo?
> 
> It will be done later because the test framework in models repo only works for Keras models at the moment.

The unit test is in sql-machine-learning/models/pull/51

typhoonzero (approved):
Cool

PR #1681, opened 2020-01-10
Comment out unstable test cases

tonyyang-svail:
Related  #1682.


shendiaomo:
This is closed because #1677 seems to have fixed #1682 .

PR #1699, opened 2020-01-14
Update Java formatter in pre-commit

wangkuiyi:
Currently, we configure pre-commit to use this https://github.com/wangkuiyi/google-style-precommit-hook Java format hook script, which downloads the formatter, if not yet, when users run the git commit command, into `./cahce/`.

This PR changes the behavior to download the formatter when building the Docker image.  Also, it assumes that the formatter is in /usr/local/bin, which is the common case as users try to install the tool before using it.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -31,3 +31,10 @@
     rev: v4.3.21
     hooks:
     -   id: isort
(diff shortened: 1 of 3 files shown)


tonyyang-svail (approved):
LGTM.

PR #1701, opened 2020-01-14
Make USING clause optional in EXPLAIN statment

### LINKED ISSUE #1695: Support omitting USING clause in explain statements
Generally, we can deduce the explainer type from the models, that is, the USING clause is not necessary in most cases.

shendiaomo:
Fix #1695

### DIFF
--- a/pkg/parser/extended_syntax_parser.y
+++ b/pkg/parser/extended_syntax_parser.y
@@ -218,6 +218,10 @@ explain_clause
 | TO EXPLAIN IDENT USING IDENT INTO IDENT { $$.TrainedModel = $3; $$.Explainer = $5; $$.ExplainInto = $7 }
 | TO EXPLAIN IDENT WITH attrs USING IDENT { $$.TrainedModel = $3; $$.ExplainAttrs = $5; $$.Explainer = $7 }
 | TO EXPLAIN IDENT WITH attrs USING IDENT INTO IDENT { $$.TrainedModel = $3; $$.ExplainAttrs = $5; $$.Explainer = $7; $$.ExplainInto = $9 }
+| TO EXPLAIN IDENT { $$.TrainedModel = $3; }
+| TO EXPLAIN IDENT INTO IDENT { $$.TrainedModel = $3; $$.ExplainInto = $5 }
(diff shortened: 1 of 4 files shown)


pkg/parser/extended_syntax_parser.y
```diff
@@ -218,6 +218,10 @@ explain_clause
 | TO EXPLAIN IDENT USING IDENT INTO IDENT { $$.TrainedModel = $3; $$.Explainer = $5; $$.ExplainInto = $7 }
 | TO EXPLAIN IDENT WITH attrs USING IDENT { $$.TrainedModel = $3; $$.ExplainAttrs = $5; $$.Explainer = $7 }
 | TO EXPLAIN IDENT WITH attrs USING IDENT INTO IDENT { $$.TrainedModel = $3; $$.ExplainAttrs = $5; $$.Explainer = $7; $$.ExplainInto = $9 }
+| TO EXPLAIN IDENT { $$.TrainedModel = $3; }
+| TO EXPLAIN IDENT INTO IDENT { $$.TrainedModel = $3; $$.ExplainInto = $5 }
+| TO EXPLAIN IDENT WITH attrs { $$.TrainedModel = $3; $$.ExplainAttrs = $5; }
+| TO EXPLAIN IDENT WITH attrs INTO IDENT { $$.TrainedModel = $3; $$.ExplainAttrs = $5; $$.ExplainInto = $7 }
```
tonyyang-svail: Do you think it is a good idea to define an optional rule 
```
opt_using
: /* empty */  { $$ = ""}
| USING IDENT  { $$ = $2 }
;
```
So that the explain clause is cleaner?

shendiaomo: Yes and done.

tonyyang-svail (approved):
LGTM++

pkg/sql/codegen/xgboost/codegen_explain.go
```diff
@@ -29,9 +28,6 @@ const (
 
 // Explain generates a Python program to explain a trained model.
 func Explain(explainStmt *ir.ExplainStmt, session *pb.Session) (string, error) {
-	if explainStmt.Explainer != "TreeExplainer" {
```
typhoonzero, line 32: `explainStmt.Explainer` is actually not used now, maybe add an issue about use the actual value.

PR #1694, opened 2020-01-13
Support explaining DNNs using SHAP

shendiaomo:
This is based on the discussion in https://github.com/slundberg/shap/issues/280
The screenshot looks like:
image

PR #1714, opened 2020-01-15
Add unit test in python for dnn explanation

shendiaomo:
Fix https://github.com/sql-machine-learning/sqlflow/pull/1694#issuecomment-574219483

### DIFF
--- a/python/sqlflow_submitter/tensorflow/explain_example.py
+++ b/python/sqlflow_submitter/tensorflow/explain_example.py
@@ -21,6 +21,7 @@ from sqlflow_submitter.tensorflow.predict import pred
 from sqlflow_submitter.tensorflow.train import train
 
 if __name__ == "__main__":
+    # Train and explain BoostedTreesClassifier
     train(datasource=datasource,
           estimator=tf.estimator.BoostedTreesClassifier,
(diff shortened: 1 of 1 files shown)


python/sqlflow_submitter/tensorflow/explain_example.py
```diff
@@ -56,3 +58,37 @@
...
+            feature_metas=feature_metas,
+            label_meta=label_meta,
+            model_params={
+                "n_classes": 3,
+                "hidden_units": [100, 100],
+            },
+            save="dnnmodel",
+            is_pai=False,
+            plot_type='bar',
+            result_table="iris.explain_result")
```
typhoonzero, line 94: Will this overwrite the above test's result?

typhoonzero (approved):

python/sqlflow_submitter/tensorflow/explain_example.py
shendiaomo: No. They are saved to different files and DNN doesn't support explaining to table now.

typhoonzero: Maybe set `result_table=""` to avoid confusing?

PR #1717, opened 2020-01-15
Add doc string for models in sqlflow_models

### LINKED ISSUE #1558: Extract docstrings to model parameters dynamically
As #1556 and #1555 demonstrated, the current implementation of parameter checking for custom models is limited, we should generate model parameter documents dynamically to make autocomplete and semantic check works.

shendiaomo:
Partially fix #1558 
A pitfall is that it takes longer for repl/sqlflowserver to start-up because we have to `import tensorflow`

### DIFF
--- a/pkg/sql/codegen/attribute/attribute.go
+++ b/pkg/sql/codegen/attribute/attribute.go
@@ -16,6 +16,8 @@ package attribute
 import (
 	"encoding/json"
 	"fmt"
+	"log"
+	"os/exec"
 	"reflect"
 	"sort"
 	"strings"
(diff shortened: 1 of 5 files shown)


pkg/sql/codegen/attribute/attribute.go
```diff
@@ -158,6 +160,18 @@ func NewDictionaryFromModelDefinition(estimator, prefix string) Dictionary {
 // PremadeModelParamsDocs stores parameters and documents of all known models
 var PremadeModelParamsDocs map[string]map[string]string
 
+// ExtractDocString extracts parameter documents from python doc strings
+func ExtractDocString(module ...string) {
+	cmd := exec.Command("python", "-uc", fmt.Sprintf("__import__('extract_docstring').print_param_doc('%s')", strings.Join(module, "', '")))
```
wangkuiyi, line 165: Should we make a local call here to print the docstring, other than start a process?

shendiaomo: That's somewhat hard to implement because we have to import the classes to get the docstrings. I have investigated several python bindings of golang but they are either too primitive ([go-python](https://github.com/sbinet/go-python)) or not as reliable.

wangkuiyi (approved):

pkg/sql/codegen/attribute/attribute.go
```diff
@@ -158,6 +160,18 @@ func NewDictionaryFromModelDefinition(estimator, prefix string) Dictionary {
...
+	output, e := cmd.CombinedOutput()
+	if e != nil {
+		log.Println("ExtractDocString failed: ", e, string(output))
+	}
+	if e := json.Unmarshal(output, &PremadeModelParamsDocs); e != nil {
```
typhoonzero, line 170: Maybe add a comment that unmarshal can append entries rather than overwrite it: https://golang.org/pkg/encoding/json/#Unmarshal

scripts/test/ipython.sh
```diff
@@ -45,6 +45,7 @@ DATASOURCE="mysql://root:root@tcp(127.0.0.1:3306)/?maxAllowedPacket=0"
 export PYTHONPATH=$GOPATH/src/sqlflow.org/sqlflow/python
 
 sqlflowserver &
+sleep 10
```
tonyyang-svail, line 48: Is it possible to change this sleep to a while loop that pings the SQLFlow server?

PR #1722, opened 2020-01-16
Enable auto type conversion from int to float

### LINKED ISSUE #1356: Automatic conversion in attribute type checking
Type checking of attributes is too strict, providing an `int` to an attribute with type `float` results in error.
For example, `subsample` of xgboost accepts `float`s, if we pass a `0` or `1` to the attribute, it' should work rather than fail:
image

Support automatic conversion from `int` to `float`.

shendiaomo:
Fix #1356. It seems we only need to support converting from int to float.
Now the following statement can run correctly:
```sql
SELECT * FROM iris.train TO TRAIN xgboost.gbtree WITH subsample=1 LABEL class INTO sqlflow_models.xxx
```

### DIFF
--- a/pkg/sql/codegen/attribute/attribute.go
+++ b/pkg/sql/codegen/attribute/attribute.go
@@ -97,7 +97,10 @@ func (d Dictionary) Validate(attrs map[string]interface{}) error {
 		}
 
 		if desc.Type != Unknown && desc.Type != reflect.TypeOf(v) {
-			return fmt.Errorf(errUnexpectedType, k, desc.Type, v)
+			// Allow implicit converstion from int to float to ease typing
+			if !(desc.Type == Float && reflect.TypeOf(v) == Int) {
+				return fmt.Errorf(errUnexpectedType, k, desc.Type, v)
+			}
 		}
 
 		if desc.Checker != nil {


tonyyang-svail:
Hi @shendiaomo, thanks for submitting this feature. I am not sure if it is necessary though. Why shouldn't we require a user to write `subsample = 1.0`?

Subsample only accepts (0, 1], and 1 is the default value, so the use case is likely being `0.*`.

shendiaomo:
> Hi @shendiaomo, thanks for submitting this feature. I am not sure if it is necessary though. Why shouldn't we require a user to write `subsample = 1.0`?
> 
> Subsample only accepts (0, 1], and 1 is the default value, so the use case is likely being `0.*`.

This is only an example that shows how the auto type conversion works. In fact, there're many parameters other than `subsample` that accepts a `float` greater than 1. For example, `gamma`, `scale_pos_weight `, `min_child_weight`, `lambda`, `alpha` in XGBoost, and `tree_complexity `in BoostedTreesEstimator.
Enabling the auto conversion from int to float in SQLFlow is conformable to the principle of least surprise because the users already have the implicit conversion in python for years.

wangkuiyi (approved):

typhoonzero (commented):
Should Add test to ensure this works.

shendiaomo:
> Should Add test to ensure this works.

Roger that. I will fix it in another PR.

PR #1730, opened 2020-01-16
Support explaining feature combination via shap interaction value and decision plot

shendiaomo:
Use the following statement to explain the example in the [explain tutorial](https://github.com/sql-machine-learning/sqlflow/blob/develop/doc/tutorial/housing-explain.md)
```sql
SELECT * FROM boston.train 
TO EXPLAIN sqlflow_models.my_xgb_regression_model 
WITH shap_summary.plot_type=decision;
```
The result is decision thats show the importance of crossed features as well as other features.

If we show only several samples as 
```sql
SELECT * FROM boston.train LIMIT 1
TO EXPLAIN sqlflow_models.my_xgb_regression_model
WITH shap_summary.plot_type=decision;
```
The result decision2 shows the cumulative effect of interactions.

p. s. Maybe we have to refactor the explanation module later to reorganize the explanation methods.

### DIFF
--- a/python/sqlflow_submitter/xgboost/explain.py
+++ b/python/sqlflow_submitter/xgboost/explain.py
@@ -50,7 +50,7 @@ def explain(datasource, select, feature_field_meta, label_name,
 
     shap_values, shap_interaction_values, expected_value = xgb_shap_values(x)
 
-    if summary_params["plot_type"] == "decision":
+    if summary_params.get("plot_type") == "decision":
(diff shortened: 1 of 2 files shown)


shendiaomo:
See https://slundberg.github.io/shap/notebooks/plots/decision_plot.html for how to use `decision_plot` and `shap_interaction_values`.

llxxxll (approved):
LGTM!!

PR #1748, opened 2020-01-18
Move the definition of FeatureColumn

wangkuiyi:
from ir.go to feature_column.go.

The interface type FeatureColumn is in a file different from the one containing its implementations.  It took me some time to understand that feature_column.go contains the implementations of a type defined in ir.go.


terrytangyuan (approved):

typhoonzero:
LGTM++

PR #1749, opened 2020-01-18
Move sql/model.go into pkg/model/

wangkuiyi:
After the merge of this PR, we should move the function [`loadModelMeta`](https://github.com/sql-machine-learning/sqlflow/pull/1749/files#diff-32ac49e7de42a0850ab196a24645814cR128) from `ir_generator.go` to package model.


typhoonzero (approved):
LGTM

PR #1752, opened 2020-01-19
Support csv format label column

### LINKED ISSUE #1715: [Need Discuss] Support simple functions on LABEL clause
Background: https://github.com/sql-machine-learning/sqlflow/issues/1643

We may have multiple target columns when training:

The task is to use `X1,X2,X3` to predict `Y1,Y2`.

The column `Y1` and `Y2` will be concatenated to a Tensor of shape [batch_size, 2] (note that typical classification jobs, the label is of shape [batchsize, 1] ). And the input features is of shape [batch_size, time_window_size, feature_size].

We can use the below approaches to achieve this:

1. write a concatenated CSV value in a single `Y` column using pure SQL before training.
1. use `LABEL CONCAT(Y1,Y2)` to do simple label transformations.

Note we don't use `LABEL Y1, Y2` since we might need to support cases using multiple training targets: https://towardsdatascience.com/regression-models-with-multiple-target-variables-8baa75aacd

typhoonzero:
Fix https://github.com/sql-machine-learning/sqlflow/issues/1715

### DIFF
--- a/cmd/sqlflowserver/main_test.go
+++ b/cmd/sqlflowserver/main_test.go
@@ -276,9 +276,11 @@ func TestEnd2EndMySQL(t *testing.T) {
 		t.Fatalf("prepare test dataset failed: %v", err)
 	}
 
-	t.Run("TestShowDatabases", CaseShowDatabases)
-	t.Run("TestSelect", CaseSelect)
-	t.Run("TestTrainSQL", CaseTrainSQL)
+	t.Run("CaseShowDatabases", CaseShowDatabases)
+	t.Run("CaseSelect", CaseSelect)
+	t.Run("CaseTrainSQL", CaseTrainSQL)
+	t.Run("CaseTrainWithCSVLabel", CaseTrainWithCSVLabel)
+
 	t.Run("CaseTrainBoostedTreesEstimatorAndExplain", CaseTrainBoostedTreesEstimatorAndExplain)
 	t.Run("CaseTrainSQLWithMetrics", CaseTrainSQLWithMetrics)
 	t.Run("TestTextClassification", CaseTrainTextClassification)
@@ -870,6 +872,19 @@ COLUMN sepal_length, sepal_width, petal_length, petal_width
 LABEL class
 INTO sqlflow_models.my_dnn_model_custom_functional;`
 	_, _, err = connectAndRunSQL(trainSQL)
+	if err != nil {
+		a.Fail("run trainSQL error: %v", err)
+	}
+}
+
+func CaseTrainWithCSVLabel(t *testing.T) {
+	a := assert.New(t)
+	trainSQL := `select f1,f2,f3,CONCAT(f13,",", target) as class from housing.train
+TO TRAIN sqlflow_models.DNNRegressor
+WITH model.hidden_units = [10, 20]
+LABEL class
+INTO sqlflow_models.my_dnn_model_csvlabel;`
+	_, _, err := connectAndRunSQL(trainSQL)
 	if err != nil {
 		a.Fail("run trainSQL error: %v", err)
 	}
--- a/cmd/sqlflowserver/main_test.go
+++ b/cmd/sqlflowserver/main_test.go
@@ -294,6 +294,8 @@ func TestEnd2EndMySQL(t *testing.T) {
 	t.Run("CaseTrainRegression", CaseTrainRegression)
 	t.Run("CaseTrainXGBoostRegression", CaseTrainXGBoostRegression)
 	t.Run("CasePredictXGBoostRegression", CasePredictXGBoostRegression)
+	t.Run("CaseTrainAndExplainXGBoostModel", CaseTrainAndExplainXGBoostModel)
+
 	t.Run("CaseTrainDeepWideModel", CaseTrainDeepWideModel)
 	t.Run("CaseTrainDeepWideModelOptimizer", CaseTrainDeepWideModelOptimizer)
 	t.Run("CaseTrainAdaNetAndExplain", CaseTrainAdaNetAndExplain)
--- a/pkg/sql/codegen/xgboost/template_explain.go
+++ b/pkg/sql/codegen/xgboost/template_explain.go
@@ -22,7 +22,7 @@ type explainFiller struct {
 	DatasetSQL           string
 	ShapSummaryParams    string
 	FeatureFieldMetaJSON string
-	LabelName            string
+	LabelJSON            string
 }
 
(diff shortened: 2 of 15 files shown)


pkg/sql/ir_generator.go
```diff
@@ -125,30 +125,30 @@ func generateTrainStmt(slct *parser.SQLFlowSelectStmt) (*ir.TrainStmt, error) {
...
 
 	if e := verifier.VerifyColumnNameAndType(tr.SQLFlowSelectStmt, pr, db); e != nil {
-		return nil, fmt.Errorf("VerifyColumnNameAndType: %v", e)
+		return nil, nil, fmt.Errorf("VerifyColumnNameAndType: %v", e)
 	}
 
 	pr.TrainClause = tr.TrainClause
 
-	return pr, nil
+	return pr, tr.SQLFlowSelectStmt, nil
```
wangkuiyi, line 151: Does this change fix https://github.com/sql-machine-learning/sqlflow/issues/1759 ?

cmd/sqlflowserver/main_test.go
```diff
@@ -875,6 +879,19 @@ INTO sqlflow_models.my_dnn_model_custom_functional;`
 	}
 }
 
+func CaseTrainWithCSVLabel(t *testing.T) {
```
wangkuiyi, line 882: I am a little confused -- why is the label called CSV label? CSV is a file format, right? Should we say `concat_label`?

pkg/step/feature/derivation.go
```diff
@@ -486,6 +487,23 @@ func setBackToIR(trainStmt *ir.TrainStmt, fcMap ColumnMap, columnTargets []strin
 	}
 }
 
+// setBackToLabel update label field meta
+func setBackToLabel(trainStmt *ir.TrainStmt, fmMap FieldDescMap) {
+	// NOTE: clustering model may not specify Label
+	labelName := trainStmt.Label.GetFieldDesc()[0].Name
+	if labelName == "" {
+		return
```
wangkuiyi: I see. How about moving the comment `// NOTE: clustering model doesn't have label` to the end of this line?

pkg/step/feature/derivation.go
```diff
@@ -486,6 +487,23 @@ func setBackToIR(trainStmt *ir.TrainStmt, fcMap ColumnMap, columnTargets []strin
...
+	}
+	trainStmt.Label = &ir.NumericColumn{
+		FieldDesc: fmMap[labelName],
+	}
+	// use shape [] if label shape is [1] for Tensorflow scalar label shape should be [].
+	shape := trainStmt.Label.GetFieldDesc()[0].Shape
```
wangkuiyi, line 501: Here I don't understand -- how comes shape? And why do we need to set the field Shape?

pkg/step/feature/derivation.go
```diff
@@ -486,6 +487,23 @@ func setBackToIR(trainStmt *ir.TrainStmt, fcMap ColumnMap, columnTargets []strin
 	}
 }
 
+// setBackToLabel update label field meta
+func setBackToLabel(trainStmt *ir.TrainStmt, fmMap FieldDescMap) {
```
wangkuiyi, line 492: I am not sure what does the phrase *set back* mean?

pkg/sql/codegen/xgboost/template_explain.go
```diff
@@ -22,22 +22,22 @@ type explainFiller struct {
 	DatasetSQL           string
 	ShapSummaryParams    string
 	FeatureFieldMetaJSON string
-	LabelName            string
+	LabelJSON            string
```
wangkuiyi, line 25: What is the relationship between label and JSON? Is that XGBoost requires JSON-encoded inputs?

wangkuiyi (commented):
I am not familiar with the related source files. Just added some comments trying to better understand the code.

pkg/sql/ir_generator.go
typhoonzero: Not really, this is only a feature requested from @Echo9573 to support some time series model training, please refer to: #1715

cmd/sqlflowserver/main_test.go
typhoonzero: CSV stands for "Comma Separated Values", since it's some values, I think it's OK to represent the data in the database cell.

pkg/sql/codegen/xgboost/template_explain.go
typhoonzero: It's just used to pass `ir.FieldDesc` to a Python dict object using JSON as the serialization method.

pkg/step/feature/derivation.go
typhoonzero: Indeed "set back" is not easy to understand. I was thinking, this function is setting derived column information back to the original IR structure, so it's a "set back", maybe we need a better name. I am updating this to the comment though.

pkg/step/feature/derivation.go
typhoonzero: I don't know why Tensorflow only accept shape `[]` for the label, pass `[1]` will cause an error. I'm wondering if we should pass `[None, 1]` because Tensorflow uses `None` to indicate the batch size dimension. I still need to do more tests to confirm this.

pkg/step/feature/derivation.go
typhoonzero: Done.

cmd/sqlflowserver/main_test.go
wangkuiyi: I see. Then `CSVLabel` means `comma-separated value label`, which is ambiguous -- is it a value or a label.  Should we use the name `CaseTrainWithCommaSeparatedLabel`?

If there could be column-separated label and semi-column-separated labels in the future, we might want `ConcatenatedLabel` or `ConcatLabel` instead.

pkg/step/feature/derivation.go
wangkuiyi: Do you mean `updateLabel` or `deriveLabel` like `deriveColumn`?

wangkuiyi (approved):

cmd/sqlflowserver/main_test.go
typhoonzero: Sounds good, will update the naming in next PR.

pkg/step/feature/derivation.go
typhoonzero: I prefer `deriveLabel` will update this in the next PR.

PR #1758, opened 2020-01-19
Support one-line comment in REPL

### LINKED ISSUE #1757: SQLFlow doesn't support comments without semicolon;
image
If we delete the `;`, REPL will concatenate the two lines.

shendiaomo:
Partially fix #1757

### DIFF
--- a/cmd/repl/prompt.go
+++ b/cmd/repl/prompt.go
@@ -84,7 +84,7 @@ func (p *promptState) changeLivePrefix() (string, bool) {
 
 func (p *promptState) execute(in string, cb func(string)) {
 	in = strings.TrimRight(in, " \t")
-	if in != "" {
+	if in != "" && !(p.statement == "" && strings.HasPrefix(strings.TrimSpace(in), "-- ")) {
 		p.statement += in
 		if strings.HasSuffix(in, ";") {
 			p.updateHistory()
--- a/cmd/repl/prompt.go
+++ b/cmd/repl/prompt.go
@@ -84,7 +84,7 @@ func (p *promptState) changeLivePrefix() (string, bool) {
 
 func (p *promptState) execute(in string, cb func(string)) {
 	in = strings.TrimRight(in, " \t")
-	if in != "" && !(p.statement == "" && strings.HasPrefix(strings.TrimSpace(in), "-- ")) {
+	if in != "" && !(p.statement == "" && lineIsComment(in)) {
 		p.statement += in
 		if strings.HasSuffix(in, ";") {
 			p.updateHistory()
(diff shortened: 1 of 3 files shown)


cmd/repl/prompt.go
```diff
@@ -84,7 +84,7 @@ func (p *promptState) changeLivePrefix() (string, bool) {
 
 func (p *promptState) execute(in string, cb func(string)) {
 	in = strings.TrimRight(in, " \t")
-	if in != "" {
+	if in != "" && !(p.statement == "" && strings.HasPrefix(strings.TrimSpace(in), "-- ")) {
```
wangkuiyi: I think we need unit test for function execute.

cmd/repl/repl.go
```diff
@@ -45,6 +45,9 @@ const tablePageSize = 1000
 func readStmt(scn *bufio.Scanner) (string, error) {
 	stmt := ""
 	for scn.Scan() {
+		if stmt == "" && strings.HasPrefix(strings.TrimSpace(scn.Text()), "-- ") {
```
wangkuiyi: We need unit test for readStmt.

wangkuiyi (commented):
Thanks for this PR. I think unt test would help us understand the change.

tonyyang-svail (commented):
Hi @shendiaomo, where would you prefer handling comments, in the parser or in REPL?

One advantage of handling comments in the parser is it can be shared across REPL/SQLFlow server/SQLFlow client.

cmd/repl/prompt.go
shendiaomo: Done.

shendiaomo:
> Thanks for this PR. I think unt test would help us understand the change.

Done.

shendiaomo:
> Hi @shendiaomo, where would you prefer handling comments, in the parser or in REPL?
> 
> One advantage of handling comments in the parser is it can be shared across REPL/SQLFlow server/SQLFlow client.

I agree that support comments in the parser is great. However, I think the one-line comment will be executed in third party parsers. So this change seems to be the only way to support the feature.

wangkuiyi (approved):

cmd/repl/repl.go
```diff
@@ -40,11 +40,22 @@ import (
 
 const tablePageSize = 1000
 
+func lineIsComment(line string) bool {
+	line = strings.TrimSpace(line)
+	if line == "--" || strings.HasPrefix(line, "-- ") {
```
wangkuiyi, line 45: Does SQL syntax state that `--` plus whitespace indicate the beginning of a comment?

shendiaomo: The standard syntax only requires double hyphen as the beginning of a comment.

It's a MySQL extension that requires a whitespace after the `--`. Because it also works on standard syntax as well as most SQL dialects, it may be reasonable for SQLFlow to accept this extension, thus making SQLFlow programs with comments portable to all other SQL dialects.

Issue #1434, opened 2019-12-16
Re-structure the source code


wangkuiyi:
We are in a stage that if we want to go further, we need to restructure our code tree.

The code directory structure depends on logical structure of SQLFlow.  We are making SQLFlow working in the following way:

1. The SQLFlow server translates a request, which is a SQL program of zero, one, or more statements, into a workflow, where each step calls the command-line tool `sqlflow` to run a statement.  As we can see that this workflow doesn't include control flows like if-else or loop, because SQL doesn't have control flows.

1. The translation consists of two steps
   1. parsing, which converts the SQL program into the program IR, and
   1. codegen, which converts the program IR, which is a list of statement IRs, into the workflow, whose each step calls `sqlflow` to execute a statement IR.
      1. The command-line program `sqlflow` converts a statement IR into a Python submitter program and executes it.

In general, we are going to have the following packages in `pkg/`:

1. ir
1. parser
1. codegen
1. executor

wangkuiyi:
1. pkg/parser.Parse(program) -> []parser.SQLFlowStmt
1. pkg/resolver.Resolve([]parser.SQLFlowStmt, *database.DB, modelDir string, cwd string, getTrainStmtFromModel bool) -> []ir.SQLStatement
1. pkg/verifier.Verify([]ir.SQLStatement, db *database.DB) -> []ir.SQLStatement
1. pkg.couler.GenerateCode([]ir.SQLStatement, *session.Session) -> string
1. pkg/runner.run(*pipe.Writer, string) -> workflowid

The call to run a SQL program should like:

```go
runner.run(
    *pipe.Writer,
    couler.GenerateCode(
        resolver.Resolve(parser.Parse(program), *database.DB, modelDir string, cwd string, getTrainStmtFromModel bool)
    )
)

```


1. pkg/resolver.ResolveAttribute(parser.SQLFlowStmt) -> map[string]interface{}
1. pkg/resolver.ResolveColumns(parser.SQLFlowStmt) -> map[string][]ir.FeatureColumn

wangkuiyi:
@typhoonzero Could we have something like

```go
func Parse(program string) []*parser.SQLFlowStmt {
}

func Resolve([]*parser.SQLFlowStmt, *database.DB, cwd string) []*ir.SQLFlowStmt {
}

func Verify([]*ir.SQLFlowStmt) []*ir.SQLFlowStmt {
}

func GenCode([]*ir.SQLFlowStmt) string {
}

[...]

PR #1773, opened 2020-01-21
Cleanup pkg/parser API

wangkuiyi:
Minimize the API of pkg/parser and make it like the one in https://github.com/sql-machine-learning/sqlflow/issues/1434#issuecomment-575966175

After the merge of this PR, the exported types include:

1. `type SQLFlowStmt struct`, and
1. AST node-like types defined in goyacc file.

The exported functions include:

1. `func ParseStatement(dialect, program string) (*SQLFlowStmt, error)`
1. `func Parse(dialect, program string) ([]*SQLFlowStmt, error) `

The following commands help me understand the API.

```bash
git grep 'type *[A-Z]'
git grep 'func *[A-Z]'
```


shendiaomo (approved):
lgtm++

PR #1779, opened 2020-01-22
Add end to end test for repl

shendiaomo:
On my local testing, these commits have coverage of around 67%,  which is about 20% higher than before.

### DIFF
--- a/cmd/repl/repl_test.go
+++ b/cmd/repl/repl_test.go
@@ -28,7 +28,9 @@ import (
 	"github.com/stretchr/testify/assert"
 
 	"github.com/c-bata/go-prompt"
+	"sqlflow.org/sqlflow/pkg/database"
 	"sqlflow.org/sqlflow/pkg/sql"
+	"sqlflow.org/sqlflow/pkg/sql/testdata"
 )
 
 var space = regexp.MustCompile(`\s+`)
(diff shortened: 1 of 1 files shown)


cmd/repl/repl_test.go
```diff
@@ -26,12 +27,164 @@ import (
...
+	assertConnectable(dbConnStr)
+	testDB, _ := database.OpenAndConnectDB(dbConnStr)
+	if testDBDriver == "mysql" {
+		_, e := testDB.Exec("CREATE DATABASE IF NOT EXISTS sqlflow_models;")
+		if e != nil {
+			return e
+		}
+		return testdata.Popularize(testDB.DB, testdata.IrisSQL)
+	}
+	t.Skip("Skipping mysql tests")
```
typhoonzero, line 50: Skipping tests that not using mysql

cmd/repl/repl_test.go
```diff
@@ -26,12 +27,164 @@ import (
...
 )
 
-// TODO(shendiaomo): end to end tests like sqlflowserver/main_test.go
-
 var space = regexp.MustCompile(`\s+`)
+var dbConnStr = "mysql://root:root@tcp(127.0.0.1:3306)/?maxAllowedPacket=0"
+var testDBDriver = os.Getenv("SQLFLOW_TEST_DB")
+var session = makeSessionFromEnv()
+
+func prepareTestData(t *testing.T) error {
```
typhoonzero: rename to `prepareTestDataOrSkip` ?

cmd/repl/repl_test.go
```diff
@@ -26,12 +27,164 @@ import (
...
+	out = <-outC
+	return
+}
+
+func TestSwitchDatabase(t *testing.T) {
+	a := assert.New(t)
+	prepareTestData(t)
+	session.DbConnStr = dbConnStr
+	results := []interface{}{}
+	for r := range sql.RunSQLProgram("show tables", "", session).ReadAll() {
```
typhoonzero: `sql.RunSQLProgram` is testing the `sql` package, since below case is already testing `runStmt` maybe we can just remove this test case.

cmd/repl/repl_test.go
shendiaomo: I think it means ''We are skipping mysql tests", the same as https://github.com/sql-machine-learning/sqlflow/blob/b4a186551ff0877c534e9a84d3a51673389e0860/cmd/sqlflowserver/main_test.go#L260-L262

cmd/repl/repl_test.go
shendiaomo: Done.

cmd/repl/repl_test.go
shendiaomo: Done.

wangkuiyi (approved):

typhoonzero (approved):

PR #1791, opened 2020-01-31
Remove unused ALPS Python code

tonyyang-svail:


typhoonzero:
Can we keep this until we are so sure that it could be removed?

tonyyang-svail:
Has the ALPS related Go code been removed? Shall we keep Go&Python consistent?

typhoonzero:
> Has the ALPS related Go code been removed?

Yes, it's already removed, I think it's better to keep the python code so that we can add the go codegens when we need it.

tonyyang-svail:
Sure.

PR #1792, opened 2020-02-01
Integrate Go gRPC client with Java gRPC parser

tonyyang-svail:
Part of #1389.


java/parser/src/main/java/org/sqlflow/parser/ParserGrpcServer.java
```diff
@@ -6,6 +6,7 @@
 import java.io.IOException;
 import java.util.concurrent.TimeUnit;
 import java.util.logging.Logger;
+import org.apache.commons.cli.*;
```
wangkuiyi: https://google.github.io/styleguide/javaguide.html#s3.3.1-wildcard-imports

pkg/parser/external/java_parser.go
```diff
@@ -34,56 +32,69 @@ func newJavaParser(typ string) *javaParser {
...
+func startServerIfNotUp() error {
+	address := getServerAddress()
+	if isServerUp(address) {
+		return nil
 	}
 
-	// TODO(yi): It is very expensive to start a Java process.  It
-	// slows down SQLFlow server's QPS if for every parsing
-	// operation, we'd have to start a Java process.
+	port := getEnv("SQLFLOW_EXTERNAL_PARSER_PORT", "12345")
```
wangkuiyi: I am afraid that 12345 is a magic number. Consider that in the future we want to add another gRPC server -- if we allow default port, we might accidentally set its default port to 12345 too and make a conflict.

A safe approach might be that the program `log.Fatal` if there is no valid value of `SQLFLOW_PARSER_SERVER_PORT`.

pkg/parser/external/java_parser.go
```diff
@@ -34,56 +32,69 @@ func newJavaParser(typ string) *javaParser {
 	return &javaParser{typ: typ}
 }
 
-type parseResult struct {
-	Statements []string `json:"statements"`
-	Position   int      `json:"position"`
-	Error      string   `json:"error"`
+func getServerAddress() string {
+	ip := getEnv("SQLFLOW_EXTERNAL_PARSER_IP", "localhost")
+	port := getEnv("SQLFLOW_EXTERNAL_PARSER_PORT", "12345")
```
wangkuiyi: 12345 is a magic number.  Another use of this number is at https://github.com/sql-machine-learning/sqlflow/pull/1792/files#diff-f612dacf52e3192f6479782d99b95e1cR55. Let us either remove it or define a global constant. I think we can remove it.

pkg/parser/external/java_parser.go
```diff
@@ -34,56 +32,69 @@ func newJavaParser(typ string) *javaParser {
 	return &javaParser{typ: typ}
 }
 
-type parseResult struct {
-	Statements []string `json:"statements"`
-	Position   int      `json:"position"`
-	Error      string   `json:"error"`
+func getServerAddress() string {
+	ip := getEnv("SQLFLOW_EXTERNAL_PARSER_IP", "localhost")
```
wangkuiyi: I suggest we keep only the port environment variable and remove the IP one.

Referring to https://golang.org/pkg/net/?m=all#Dial, we see that this function could return `fmt.Sprintf(":%d", port)` without the IP.  In this case, the Go net package tries to connect to the port bound to any of the network interfaces.

pkg/parser/external/java_parser.go
```diff
@@ -34,56 +32,69 @@ func newJavaParser(typ string) *javaParser {
...
+		return nil, -1, err
+	}
+	if r.Error != "" {
+		return nil, -1, fmt.Errorf(r.Error)
 	}
+	return r.SqlStatements, int(r.Index), nil
+}
 
-	return pr.Statements, pr.Position, nil
+func getEnv(name, fallback string) string {
```
wangkuiyi: Following my suggestions above, we can remove this function definition.

pkg/parser/external/java_parser.go
```diff
@@ -34,56 +32,69 @@ func newJavaParser(typ string) *javaParser {
...
+}
+
+func (p *javaParser) Parse(program string) ([]string, int, error) {
+	if err := startServerIfNotUp(); err != nil {
 		return nil, -1, err
 	}
 
-	var pr parseResult
-	if err = json.Unmarshal(output, &pr); err != nil {
+	c, err := grpc.Dial(getServerAddress(), grpc.WithInsecure())
```
wangkuiyi: I am not sure, but I vaguely remember that we usually reuse a gRPC connection for many request/responses, other than establishing the connecting for every remote call.

wangkuiyi:
I noticed that at the server-side, for each remote call, we create an "adaptor":

https://github.com/sql-machine-learning/sqlflow/blob/10723efe508db547af9a03a6b53a02b2fbf5b60a/java/parser/src/main/java/org/sqlflow/parser/ParserGrpcServer.java#L73

It looks to me that we can keep these adaptors as class data members to save recreation time.

java/parser/src/main/java/org/sqlflow/parser/ParserGrpcServer.java
tonyyang-svail: Sure, I will remove the wildcard imports.

pkg/parser/external/java_parser.go
tonyyang-svail: Sure, I will remove all occurrences of "12345" and define `SQLFLOW_PARSER_SERVER_PORT` in the Dockerfile.

> A safe approach might be that the program log.Fatal if there is no valid value of SQLFLOW_PARSER_SERVER_PORT.

Agree.

pkg/parser/external/java_parser.go
tonyyang-svail: I will remove it.

pkg/parser/external/java_parser.go
tonyyang-svail: Sounds good.

pkg/parser/external/java_parser.go
tonyyang-svail: Sure, I will remove this function.

pkg/parser/external/java_parser.go
tonyyang-svail: You are right. It is encouraged to reuse a gRPC connection. I will update the code.

ref: https://github.com/grpc/grpc-go/pull/2034/files

tonyyang-svail:
> It looks to me that we can keep these adaptors as class data members to save recreation time.

I wasn't able to figure out whether the parser instance inside the adaptor is thread-safe, so I decided to `new` one on every request.

pkg/parser/external/java_parser_server.go
```diff
@@ -0,0 +1,85 @@
...
+func getServerPort() string {
+	port := os.Getenv("SQLFLOW_PARSER_SERVER_PORT")
+	if port == "" {
+		log.Fatal("undefined environment variable SQLFLOW_PARSER_SERVER_PORT")
+	}
+	return port
+}
+
+func isServerUp() bool {
+	cmd := exec.Command("curl", "-v", fmt.Sprintf("localhost:%s", getServerPort()))
```
wangkuiyi, line 40: Can we do not assume the existence of curl as a dependency?

```go
// Below init function
func TestNETServer_Run(t *testing.T) {
    // Simply check that the server is up and can
    // accept connections.
    conn, err := net.Dial("tcp", ":1123")
    if err != nil {
        t.Error("could not connect to server: ", err)
    }
    defer conn.Close()
}
```

pkg/parser/external/java_parser_server.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	for i := 0; i < 3; i++ {
+		time.Sleep(time.Second)
+		if isServerUp() {
+			return nil
+		}
+	}
+
+	return fmt.Errorf("unable to start external parser service")
```
wangkuiyi, line 67: unable => Unable

wangkuiyi (approved):
Great work!

Approving with the hope that comments can be addressed in future PRs.

pkg/parser/external/java_parser_server.go
tonyyang-svail: Maybe I am wrong, but I am following https://github.com/golang/go/wiki/CodeReviewComments#error-strings where error strings should not be capitalized.

pkg/parser/external/java_parser_server.go
tonyyang-svail: Thanks for the suggestion, I will try out `net.Dial`. I previous attempt on `grpc.Dial` gives nil error while the server doesn't exist.

PR #1795, opened 2020-02-03
fix pai deep model explain

typhoonzero:


Yancey0623 (approved):
LGTM, thanks.
I will complete the unit test of `PAICmd`.

PR #1800, opened 2020-02-03
Use net.Dial to determine if gRPC exists

### LINKED ISSUE #1389: The third-party parser starts a Java process every time it parses a SQL program

tonyyang-svail:
This PR is a follow up of suggestions at https://github.com/sql-machine-learning/sqlflow/pull/1792#discussion_r374339368.

This is the last PR related to https://github.com/sql-machine-learning/sqlflow/issues/1389. Therefore it should fix https://github.com/sql-machine-learning/sqlflow/issues/1389.


pkg/parser/external/java_parser_server.go
```diff
@@ -37,8 +38,13 @@ func getServerPort() string {
 }
 
 func isServerUp() bool {
-	cmd := exec.Command("curl", "-v", fmt.Sprintf("localhost:%s", getServerPort()))
-	if err := cmd.Run(); err != nil {
+	conn, err := net.Dial("tcp", fmt.Sprintf(":%s", getServerPort()))
```
shendiaomo: Shall we set a `timeout` value? How about add a unit test for this function.

tonyyang-svail: Good point. I will add a unit test.

pkg/parser/external/java_parser_server_test.go
```diff
@@ -0,0 +1,64 @@
...
+	a := assert.New(t)
+
+	port, err := getFreePort()
+	a.NoError(err)
+
+	address := fmt.Sprintf(":%d", port)
+	a.False(isServerUp(address))
+
+	// start Java parser server
+	cmd := exec.Command("java",
```
wangkuiyi, line 45: I might be a little bit overkilling to start a Java server in order to test isServerUp. Instead, a goroutine running a Go server might be sufficient.

wangkuiyi (approved):

PR #1831, opened 2020-02-08
Speed up KernelExplainer via kmeans

shendiaomo:
1. Enlarge batch size
2. Call `shap.kmeans` to reduce sample amount as recommended by SHAP official, see https://github.com/slundberg/shap/blob/master/notebooks/kernel_explainer/Diabetes%20regression.ipynb
The result looks like:
titanic_dot
It seems the binary features are clustered together，which seems ok without loss of generality


tonyyang-svail (approved):
LGTM!

PR #1835, opened 2020-02-10
Fix random sorted suggestions

shendiaomo:
In Go 1, the order in which elements are visited when iterating over a map using a for range statement is random, which is strange for SQLFlow REPL users, this change fixed the problem.
See https://stackoverflow.com/questions/55925822/in-golang-why-are-iterations-over-maps-random and https://nathanleclaire.com/blog/2014/04/27/a-surprising-feature-of-golang-that-colored-me-impressed/

### DIFF
--- a/cmd/repl/prompt.go
+++ b/cmd/repl/prompt.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


Yancey0623 (approved):
LGTM

PR #1821, opened 2020-02-07
Fix estimator predict performance

typhoonzero:
Fix https://github.com/sql-machine-learning/sqlflow/issues/1813

PR #1842, opened 2020-02-12
Fix PAI export model, add category hash col for ctr case

typhoonzero:
1. Fix running on PAI after merging https://github.com/sql-machine-learning/sqlflow/pull/1821 (upload exported model to OSS and load it when predicting)
1. Add `CATEGORY_HASH` as category_column_with_hash_buckets to support CTR test case
1. Fix writing prediction feature columns when features are CSV format.

TODO: remove `pai_maxcompute_input_fn` completely


Yancey0623 (approved):
LGTM

PR #1849, opened 2020-02-13
Enable LIMIT clause for calcite

shendiaomo:
A `SELECT` statement with `LIMIT` is `SqlKind.ORDER_BY` rather than `SqlKind.SELECT`.


Yancey0623 (approved):
LGTM

PR #1850, opened 2020-02-13
Support history searching in REPL

shendiaomo:
This change is different from traditional `reverse-i-search` in that:
1. Use an auto-completion UI to display the command history.
1. Support back-and-forth search while `reverse-i-search` only search back

There's one pitfall:
1. `go-prompt` 's completion will stop at whitespace or `=`, which means a user had better avoid input the special characters

TODO:
1. Find a mechanism to fix the pitfall
1. Add unit test for this feature

### DIFF
--- a/cmd/repl/prompt.go
+++ b/cmd/repl/prompt.go
@@ -71,7 +71,8 @@ type promptState struct {
 	livePrefix                     string
 	enableLivePrefix               bool
 	keywords                       []string
-	history                        []string
+	history                        []prompt.Suggest
+	inSearching                    bool
 	historyFileName                string
 	estimatorCls                   string
 	modelParamDocs                 map[string][]prompt.Suggest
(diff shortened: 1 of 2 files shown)


cmd/repl/prompt.go
```diff
@@ -287,12 +312,16 @@ var consoleWriter = prompt.NewStdoutWriter()
 
 func runPrompt(cb func(string)) {
 	state := newPromptState()
+	history := []string{}
+	for _, h := range state.history {
+		history = append([]string{h.Text}, history...)
+	}
```
weiguoz, line 318: Revert `state.history`?

shendiaomo: Yes，because `state.history` is loaded reversely to make the history completion show recent commands first.

weiguoz (approved):

PR #1851, opened 2020-02-13
Make ExplainStmt.Into optional

weiguoz:


tmp_estimator_name
```diff
@@ -0,0 +1 @@
+DNNClassifier
```
tonyyang-svail: Is this file a leftover for debugging?

typhoonzero: This is a file fetched from OSS when predicting a model, should add this in `.gitignore`

wangkuiyi (dismissed):

wangkuiyi (approved):

PR #1855, opened 2020-02-14
Rename shap_summary to summary to be consistent in all explainers

shendiaomo:


weiguoz:
Several months ago, I was not sure if the SQLFlow uses any other explainer, and the `shap_summary` implied the attribute came for `SHAP`.
It looks like no other explainers planned for the short term, so I agree to rename `shap_summary` to `summary`.

weiguoz (approved):
Several months ago, I was not sure if the SQLFlow uses any other explainer, and the shap_summary implies the attribute comes for SHAP.
It looks like no other explainers planned for the short term, so I agree to rename shap_summary to summary.

PR #1856, opened 2020-02-14
Fix category hash data type

typhoonzero:
- Fix `category_id_with_hash_buckets` support string and int data types.
- Add more CTR test cases


tonyyang-svail (approved):
LGTM.

PR #1859, opened 2020-02-14
Enable UNION clause for calcite and mysql

shendiaomo:
This is similar to #1849 .
The following statement is supported now
```sql
(SELECT *  FROM fraud.train WHERE class=0 LIMIT 1024) UNION (SELECT * FROM fraud.train WHERE class=1) TO TRAIN ...;
```


tonyyang-svail (approved):
LGTM++

tonyyang-svail:
Note: the added `(SELECT *  FROM fraud.train WHERE class=0 LIMIT 1024) UNION (SELECT * FROM fraud.train WHERE class=1) TO TRAIN ...;` test cases also passed on HiveQL parser.

PR #1864, opened 2020-02-17
Improve history completer

shendiaomo:
This is an improvement for #1850 
history_search


cmd/repl/emacs_key_bindings.go
```diff
@@ -191,18 +219,34 @@ var emacsCtrlKeyBindings = []prompt.KeyBind{
...
 	{
 		Key: prompt.ControlR,
 		Fn: func(buf *prompt.Buffer) {
-			startSearching()
+			startSearch(buf)
+		},
+	},
+	// Navigate the older command
+	{
+		Key: prompt.ControlP,
```
weiguoz, line 235: Do we have instructions about the prompt key?

weiguoz (approved):

PR #1884, opened 2020-02-20
Add a tutorial for lstmbasedtimeseries

Echo9573:


doc/tutorial/energe_lstmbasedtimeseries.md
```diff
@@ -0,0 +1,283 @@
+# LSTMBasedTimeSeriesModel on SQLFlow Tutorial
+
+This is a tutorial on how to apply a Time Series Model on [energy dataset]( https://www.dropbox.com/s/pqenrr2mcvl0hk9/GEFCom2014.zip?dl=0). 
+
+The dataset is taken from the GEFCom2014 energy forecasting [competition](Tao Hong, Pierre Pinson, Shu Fan, Hamidreza Zareipour, Alberto Troccoli and Rob J. Hyndman, "Probabilistic energy forecasting: Global Energy Forecasting Competition 2014 and beyond", International Journal of Forecasting, vol.32, no.3, pp 896-913, July-September, 2016.). It consists of years of hourly electricity load data from the New England ISO and also includes hourly temperature data. We just choose part of electricity load data which date range is from 2014-10-01 to 2014-12-31 to complete our tutorial. And, we load the dataset into MySQL manually.
```
shendiaomo: > We just choose part of electricity load data **which** date range

Should `which` be `whose`?

doc/tutorial/energe_lstmbasedtimeseries.md
```diff
@@ -0,0 +1,283 @@
...
+As the following shows, because we want to train a model that `n_in = 10 `and `n_out=4` in the next stage, we construct a time series data that length is 14. The reconstruct MySQL Statement is:
+
+```sql
+%%sqlflow
+drop table if exists energy.con_all;
+create table energy.con_all as
+select * from(
+select
+    t.dt,
+    @lagfield0:= @lagfield1 as col_1,
```
shendiaomo, line 72: Add a simple explanation for the syntax?

doc/tutorial/energe_lstmbasedtimeseries.md
```diff
@@ -0,0 +1,283 @@
...
+```sql
+%%sqlflow
+select * from energy.normalized limit 5;
+```
+
+### PART 1.2  Reconstruct Data
+
+Then, you need to choose an appropriate timestep of the series (n_in) and you need to specify how much time-steps target data you want to predict (n_out) by the trained model. By the way, the `n_in` parameter can be tuned manually during the model training stage.
+
+As the following shows, because we want to train a model that `n_in = 10 `and `n_out=4` in the next stage, we construct a time series data that length is 14. The reconstruct MySQL Statement is:
```
shendiaomo: reconstruct -> reconstructed?

doc/tutorial/energe_lstmbasedtimeseries.md
```diff
@@ -0,0 +1,283 @@
...
+```
+
+```sql
+%%sqlflow
+select count(*) from energy.raw;
+```
+
+### PART 1.1  Scale Data
+
+Because the raw data can't be used directly to build the model, we need to reconstruct the raw data into the series data instead. Before that, we need to scale the data first into the range of (0, 1). It should be noted that data scaling must be done before the data reconstruct shown in the next step.
```
shendiaomo: > we need to scale the data first into the range of (0, 1

This is min-max normalization, add a link such as https://en.wikipedia.org/wiki/Feature_scaling#Rescaling_(min-max_normalization) as an explanation?

doc/tutorial/energe_lstmbasedtimeseries.md
Echo9573: OK~ Thanks~~

doc/tutorial/energe_lstmbasedtimeseries.md
Echo9573: Yes，Thanks!

doc/tutorial/energe_lstmbasedtimeseries.md
Echo9573: OK ,Thanks~

typhoonzero (approved):
LGTM

PR #1886, opened 2020-02-21
Increase test GRPC deadline (timeout)

typhoonzero:
1. Increase end2end test deadline for PAI jobs.
2. Do not reprepare dataset when running PAI maxcompute tests.

### DIFF
--- a/cmd/sqlflowserver/main_test.go
+++ b/cmd/sqlflowserver/main_test.go
@@ -99,7 +99,8 @@ func connectAndRunSQL(sql string) ([]string, [][]*any.Any, error) {
 	}
 	defer conn.Close()
 	cli := pb.NewSQLFlowClient(conn)
-	ctx, cancel := context.WithTimeout(context.Background(), 1800*time.Second)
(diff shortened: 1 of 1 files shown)


tonyyang-svail (approved):
LGTM.

cmd/sqlflowserver/main_test.go
```diff
@@ -1724,10 +1725,6 @@ func TestEnd2EndMaxComputePAI(t *testing.T) {
 
 	go start(modelDir, caCrt, caKey, unitTestPort, false)
 	waitPortReady(fmt.Sprintf("localhost:%d", unitTestPort), 0)
-	err = prepareTestData(dbConnStr)
```
tonyyang-svail, line 1727: So are we assuming the pre-existence of the test data?

typhoonzero: Yes. Recreating the dataset tables will clean the current permission settings, which cause other users may not have access to the dataset tables.

PR #1897, opened 2020-02-24
Support indicator column

shendiaomo:
IndicatorColumn represents `tf.feature_column.indicator_column`.
ref: https://www.tensorflow.org/api_docs/python/tf/feature_column/indicator_column

### DIFF
--- a/pkg/sql/ir_generator.go
+++ b/pkg/sql/ir_generator.go
@@ -34,6 +34,7 @@ const (
 	seqCategoryID = "SEQ_CATEGORY_ID"
 	categoryHash  = "CATEGORY_HASH"
 	embedding     = "EMBEDDING"
+	indicator     = "INDICATOR"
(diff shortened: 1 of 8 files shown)


pkg/sql/ir_generator.go
```diff
@@ -708,6 +711,59 @@ func parseEmbeddingColumn(el *parser.ExprList) (*ir.EmbeddingColumn, error) {
 		Name:           embColName}, nil
 }
 
+func parseIndicatorColumn(el *parser.ExprList) (*ir.IndicatorColumn, error) {
+	help := "INDICATOR(CATEGORY_ID(...)/CATEGORY_HASH(...)|col_name])"
```
typhoonzero, line 715: change slash `/` to `|`

typhoonzero (approved):

Issue #1904, opened 2020-02-26
Proposal: Managing external & internal parsers


tonyyang-svail:
## Problem

SQLFlow calls Java HiveQL/ODPS parser to help to parse the SQL program. As shown in the following dependency graph, each parser is wrapped as a gRPC server which takes a string and returns the parsed result.

![test3](https://user-images.githubusercontent.com/29932814/75299910-294c0380-57eb-11ea-88f1-f6162b869c1e.png)

However, some parsers like ODPS parser can't be open-sourced. This leads to the circular dependencies between the internal code and open-sourced code, i.e. the open-sourced Java gRPC server needs to call ODPS parser while the close-sourced ODPS parser needs to return `ParseResult`.

![test](https://user-images.githubusercontent.com/29932814/75299952-513b6700-57eb-11ea-9742-b1369f88198d.png)

## Solution

We remove the dependencies from gRPC server to ODPS parser via [dynamic loading](https://stackoverflow.com/questions/5571466/dynamically-loading-a-class-in-java). To be specific, the open-sourced code creates ODPS parser instance by the following.

```java
Object parser = Class.forName("org.sqlflow.parser.internal.odpsParser....").newInstance();  
((ParserInterface)parser).parse(sql);  
```

By doing so, we only have a one-way dependency from internal repo to the GitHub repo.

![test2](https://user-images.githubusercontent.com/29932814/75300660-74ffac80-57ed-11ea-8b6e-39641ed53869.png)

## Implementation Details

GitHub repo:
1. Each parser should be an extension of a `ParserInterface`.
```java
interface ParseInterface {
ParseResult parse(String sql);
}
```

Internal repo:
1. The building process starts from building the GitHub repo into a `.jar` file then [add it to the Maven project](https://stackoverflow.com/questions/4955635/how-to-add-local-jar-files-to-a-maven-project).
1. The deployment of the internal `.jar` file should be the same as the open-sourced `.jar` file, since they share the same entry point.

cc @typhoonzero @weiguoz

typhoonzero:
This is definitely needed. Each parser would be one independent jar file to the gRPC service.

weiguoz:
Totally agree.
Also, we need to deploy the `interface` to a maven repository, such as: [mvnrepository.com](mvnrepository.com) or some private repository.

tonyyang-svail:
Following the comments, I updated the design as the following.

[...]

PR #1911, opened 2020-02-27
Refactor Java parser to Interface

tonyyang-svail:
Part of #1904.


java/parser/src/main/java/org/sqlflow/parser/parse/ParseResult.java
```diff
@@ -0,0 +1,14 @@
+package org.sqlflow.parser.parse;
+
+import java.util.ArrayList;
+
+// ParseResult contains the parsing result of ParseAndSplit
+public class ParseResult {
+  // SQL statements accepted by the parser
+  public ArrayList<String> Statements;
```
weiguoz: Change `Statements` to `statements` and remove the `Public`.
Access the class member by `getStatements()/setStatements` methods.

java/parser/src/main/java/org/sqlflow/parser/ParserGrpcServer.java
```diff
@@ -75,10 +79,10 @@ public void parse(ParserRequest request, StreamObserver<ParserResponse> response
 
       ParseResult parse_result;
       if (request.getDialect().equals("calcite")) {
```
weiguoz, line 81: May `request.getDialect ` be `null`?

java/parser/src/main/java/org/sqlflow/parser/ParserGrpcServer.java
```diff
@@ -75,10 +79,10 @@ public void parse(ParserRequest request, StreamObserver<ParserResponse> response
...
-        CalciteParserAdaptor parser = new CalciteParserAdaptor();
+        ParseInterface parser = new CalciteParserAdaptor();
         parse_result = parser.ParseAndSplit(request.getSqlProgram());
       } else {
-        HiveQLParserAdaptor parser = new HiveQLParserAdaptor();
+        ParseInterface parser = new HiveQLParserAdaptor();
         parse_result = parser.ParseAndSplit(request.getSqlProgram());
```
weiguoz, line 86: parseResult

java/parser/src/main/java/org/sqlflow/parser/parse/ParseInterface.java
```diff
@@ -0,0 +1,15 @@
...
+  //
+  // It returns <statements, -1, ""> if Calcite parser accepts the SQL program.
+  //     input:  "select 1; select 1;"
+  //     output: {"select 1;", "select 1;"}, -1 , nil
+  // It returns <statements, idx, ""> if Calcite parser accepts part of the SQL program,
+  // indicated by idx.
+  //     input:  "select 1; select 1 to train; select 1"
+  //     output: {"select 1;", "select 1"}, 19, nil
+  // It returns <nil, -1, error> if an error is occurred.
+  ParseResult ParseAndSplit(String sql);
```
weiguoz: Change `ParseAndSplit` to`parseAndSplit` or `parse`

java/parser/src/main/java/org/sqlflow/parser/parse/ParseResult.java
tonyyang-svail: > Change Statements to statements

You are right. [Fields should be named to camelcase](https://google.github.io/styleguide/javaguide.html#s5.2.5-non-constant-field-names). I will fix it in the next PR.

> ... remove the Public. Access the class member by getStatements()/setStatements methods.

[I searched around](https://stackoverflow.com/questions/11071407/advantage-of-set-and-get-methods-vs-public-variable), looks like getters/setters are recommended when developing more complicated classes. Here `ParseResult` is a simple struct, it doesn't have any method and it won't be inherited by any other class.

java/parser/src/main/java/org/sqlflow/parser/ParserGrpcServer.java
tonyyang-svail: The Java style check in our pre-commit hook doesn't complain about the variable naming though. Instead of spotting the naming manually, a better approach is configuring the style checker.

Also, if we decide to go with the camel case, there are multiple variable names needs to be changed (e.g. `sub_sql`, `accumulated_position`) in addition to `parser_result`, so it is better to file the change in an separate PR.

java/parser/src/main/java/org/sqlflow/parser/ParserGrpcServer.java
tonyyang-svail: No. If the client set the dialect value, the dialect won't be `null`; If the client didn't set the dialect, the optional dialect field will be set to empty string.

ref: [the Protocol Buffer Language Guide](https://developers.google.com/protocol-buffers/docs/proto3#default)
> When a message is parsed, if the encoded message does not contain a particular singular element, the corresponding field in the parsed object is set to the default value for that field. These defaults are type-specific: For strings, the default value is the empty string.

java/parser/src/main/java/org/sqlflow/parser/parse/ParseInterface.java
tonyyang-svail: You are right. Function naming should be camelCase. I will fix it in the following up PR.

weiguoz (approved):
LGTM.

PR #1913, opened 2020-02-27
Delete OSS model path before training

typhoonzero:
Delete OSS model path before training with the same model name.


Yancey0623 (approved):
LGTM

PR #1930, opened 2020-03-02
Default do not run validation if no validation.select

typhoonzero:
Default do not run validation if not specified `validation.select`


Yancey0623 (approved):
LGTM

PR #1939, opened 2020-03-04
Distributed train must specify validation.select

typhoonzero:
Estimator only support distributed training by calling `train_and_evaluate`, so fail fast if distributed training with no `validation.select` specified.


python/sqlflow_submitter/tensorflow/train.py
```diff
@@ -238,6 +238,9 @@ def validate_input_fn():
         # NOTE(typhoonzero): if only do training, no validation result will be printed.
         classifier.train(lambda: train_input_fn(), max_steps=train_max_steps)
 
+    if is_pai and FLAGS.task_index != 0:
+        print("skip exporting model on woker != 0")
```
shendiaomo: woker -> worker

shendiaomo (approved):

PR #1952, opened 2020-03-05
Save typing in COLUMN clause with regex

shendiaomo:
Sometimes typing the `COLUMN` clause is very tedious because there're many fields that have to apply the same logic. For example, the titanic dataset has many fields that are best used as indicators:
```sql
use titanic;
desc train;
```
```
+--------------+---------+------+-----+---------+-------+
|    FIELD     |  TYPE   | NULL | KEY | DEFAULT | EXTRA |
+--------------+---------+------+-----+---------+-------+
| pclass_1     | int(11) | YES  |     | <nil>   |       |
| pclass_2     | int(11) | YES  |     | <nil>   |       |
| pclass_3     | int(11) | YES  |     | <nil>   |       |
| sex_female   | int(11) | YES  |     | <nil>   |       |
| sex_male     | int(11) | YES  |     | <nil>   |       |
| embarked_c   | int(11) | YES  |     | <nil>   |       |
| embarked_q   | int(11) | YES  |     | <nil>   |       |
| embarked_s   | int(11) | YES  |     | <nil>   |       |
| title_master | int(11) | YES  |     | <nil>   |       |
| title_misc   | int(11) | YES  |     | <nil>   |       |
| title_miss   | int(11) | YES  |     | <nil>   |       |
| title_mr     | int(11) | YES  |     | <nil>   |       |
| title_mrs    | int(11) | YES  |     | <nil>   |       |
| nosibsp      | int(11) | YES  |     | <nil>   |       |
| noparch      | int(11) | YES  |     | <nil>   |       |
| nullcabin    | int(11) | YES  |     | <nil>   |       |
| cabinalpha   | int(11) | YES  |     | <nil>   |       |
| family       | int(11) | YES  |     | <nil>   |       |
| isalone      | int(11) | YES  |     | <nil>   |       |
| ismother     | int(11) | YES  |     | <nil>   |       |
| age          | float   | YES  |     | <nil>   |       |
| realfare     | float   | YES  |     | <nil>   |       |
| survived     | int(11) | YES  |     | <nil>   |       |
+--------------+---------+------+-----+---------+-------+
```
If we try to train a `BoostedTreesClassifier`, we have to write
```sql
SELECT * FROM train TO TRAIN BoostedTreesClassifier
WITH model.n_batches_per_layer=4, 
     train.batch_size=100, 
     model.center_bias=True 
COLUMN INDICATOR(CATEGORY_ID(sex_male, 2)), 
       INDICATOR(CATEGORY_ID(sex_female, 2)),
       INDICATOR(CATEGORY_ID(pclass_1, 2)), 
       INDICATOR(CATEGORY_ID(pclass_2, 2)), 
       INDICATOR(CATEGORY_ID(pclass_3, 2)),
       -- many similar lines
LABEL survived INTO tfbt_model;
```

This PR try to leverage regular expressions to simplify typing, the above training statement can now be written as
```sql
SELECT * FROM train TO TRAIN BoostedTreesClassifier
WITH model.n_batches_per_layer=4, 
     train.batch_size=100, 
     model.center_bias=True 
-- age and  realfare are floats, family has a max value of 11, others are binary
COLUMN INDICATOR(CATEGORY_ID("[^arf].*, 2)),
       INDICATOR(CATEGORY_ID(family, 12)),
LABEL survived INTO tfbt_model;
```

**Known limitation**: the golang `regexp` module doesn't support lookahead syntax. This makes it hard to write regex to exclude some fields.

### DIFF
--- a/pkg/step/feature/derivation.go
+++ b/pkg/step/feature/derivation.go
@@ -322,7 +322,7 @@ func getFeatureColumnTargets(trainStmt *ir.TrainStmt) []string {
 }
 
 // deriveFeatureColumn will fill in "fcMap" with derivated FeatureColumns.
-func deriveFeatureColumn(fcMap ColumnMap, columnTargets []string, fmMap FieldDescMap, selectFieldTypeMap fieldTypes, trainStmt *ir.TrainStmt) error {
+func deriveFeatureColumn(fcMap ColumnMap, columnTargets []string, fdMap FieldDescMap, selectFieldTypeMap fieldTypes, trainStmt *ir.TrainStmt) error {
 	// 1. Infer omitted category_id_column for embedding_columns
 	// 2. Add derivated feature column.
 	//
@@ -352,12 +352,9 @@ func deriveFeatureColumn(fcMap ColumnMap, columnTargets []string, fmMap FieldDes
 				hasMatch := false
 				for sf := range selectFieldTypeMap {
 					if r.MatchString(sf) {
-						applied, err := fcTargetMap[f][0].ApplyTo(sf)
+						applied, err := fcTargetMap[f][0].ApplyTo(fdMap[sf])
 						if err != nil {
 							return err
-						}
-						if len(applied.GetFieldDesc()) != 0 {
-							applied.GetFieldDesc()[0].Shape = fmMap[sf].Shape
 						}
 						fcMap[target][sf] = []ir.FeatureColumn{applied}
 						hasMatch = true
@@ -366,7 +363,7 @@ func deriveFeatureColumn(fcMap ColumnMap, columnTargets []string, fmMap FieldDes
 				if !hasMatch {
 					return fmt.Errorf("'%s' in 'column clause' does not match any selected fields", f)
 				}
-				delete(fmMap, f)
+				delete(fdMap, f)
 			} else {
 				fcMap[target][f] = fcTargetMap[f]
 			}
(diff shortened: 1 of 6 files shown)


pkg/step/feature/derivation.go
```diff
@@ -338,11 +338,37 @@ func deriveFeatureColumn(fcMap ColumnMap, columnTargets []string, fmMap FieldDes
...
+							return err
+						}
+						fcMap[target][sf] = []ir.FeatureColumn{applied}
+						hasMatch = true
+					}
+				}
+				if !hasMatch {
+					return fmt.Errorf("'%s' in 'column clause' does not match any selected fields", f)
+				}
+				delete(fdMap, f)
```
typhoonzero, line 366: Why need to delete the field here?

pkg/verifier/verifier.go
```diff
@@ -92,43 +92,27 @@ func Verify(q string, db *database.DB) (FieldTypes, error) {
...
-	for _, c := range trainParsed.Columns["feature_columns"] {
-		name, err := getExpressionFieldName(c)
-		if err != nil {
-			return err
+	for n, t := range trainFields {
+		if n == trainParsed.Label {
+			continue
 		}
-		it, ok := predFields.Get(name)
+		pt, ok := predFields.Get(n)
```
typhoonzero, line 110: That's a cool fix.

pkg/step/feature/derivation.go
shendiaomo: In the `if` statement that starts at L343 and ends at L367,  we assume that `f` is a regular expression because it is not in the selected fields `selectFieldTypeMap`.
After the `if !hasMatch` condition in L363-365, we are sure `f` is a regex that matches some fields in `selectFieldTypeMap`, as a result, the original `f` entry is deleted from `fdMap` because it's not a real field.

typhoonzero: I see.

typhoonzero (approved):

PR #1955, opened 2020-03-06
Run pai tests in parallel

typhoonzero:
1. Increase gRPC client timeout in test cases.
2. Make PAI test cases run in parallel.


Yancey0623 (approved):
LGTM

PR #1969, opened 2020-03-10
Disable wildcard for tensorflow model attributes to detect input errorr earlier

shendiaomo:
```sql
SELECT * FROM iris.train TO TRAIN DNNClassifier WITH model.n_class=3, model.hidden_units=[20, 20] LABEL class INTO iris.dnn_model;
```
The above statement will throw a very long error message.
After this change, the error message is
```
2020/03/10 11:45:05 runSQLProgram error: unsupported attribute model.n_class
```

### DIFF
--- a/pkg/sql/codegen/tensorflow/codegen.go
+++ b/pkg/sql/codegen/tensorflow/codegen.go
@@ -306,6 +306,17 @@ func InitializeAttributes(trainStmt *ir.TrainStmt) error {
 	constructOptimizers(trainStmt)
 	constructLosses(trainStmt)
 	attrValidator := modelAttr.Update(commonAttributes)
+	if len(modelAttr) == 0 {
+		// TODO(shendiaomo): Use the same mechanism as `sqlflow_models` to extract parameters automatically
+		// Unknown custom models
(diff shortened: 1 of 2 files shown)


typhoonzero:
Keras models accept self-defined model parameters which may not know in advance, you need to check whether it's a Keras model and do the model parameter check.

shendiaomo:
> Keras models accept self-defined model parameters which may not know in advance, you need to check whether it's a Keras model and do the model parameter check.

Roger that.

pkg/sql/codegen/tensorflow/codegen.go
```diff
@@ -307,6 +306,17 @@ func InitializeAttributes(trainStmt *ir.TrainStmt) error {
 	constructOptimizers(trainStmt)
 	constructLosses(trainStmt)
 	attrValidator := modelAttr.Update(commonAttributes)
+	if len(modelAttr) == 0 {
```
typhoonzero, line 309: The variable name `modelAttr` makes this function hard to read, maybe can call it `expectedModelAttrs`?

typhoonzero (approved):
LGTM with a naming comment.

pkg/sql/codegen/tensorflow/codegen.go
shendiaomo: Fix later.

PR #1983, opened 2020-03-12
Avoid many occurrence of cleanCwd

### LINKED ISSUE #1982: Avoid many occurrence of cleanCwd
We can avoid many occurrence `cleanCwd` by wrapping the logic in the for loop in a separate function, then use defer.

https://github.com/sql-machine-learning/sqlflow/blob/f7c232619917577b48609159c413e55d72ec9a73/pkg/sql/executor_ir.go#L177-L183

tonyyang-svail:
Fix #1982

### DIFF
--- a/pkg/sql/executor_ir.go
+++ b/pkg/sql/executor_ir.go
@@ -45,7 +45,7 @@ type WorkflowJob struct {
 	JobID string
 }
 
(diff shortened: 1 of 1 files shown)


typhoonzero (approved):
LGTM++

PR #1991, opened 2020-03-12
Add XGBoost object attribute checker

### LINKED ISSUE #1380: An bug reported when the objective="binary:logistic" in xgboost
When using xgboost for binary classification, set in the with parameter: when objective is "binary: logistic", an error will be reported as following:
image

tonyyang-svail:
Fix #1380.


Yancey0623 (approved):
LGTM

PR #2012, opened 2020-03-16
Replace standard log by logrus in Submit&Fetch

weiguoz:
Part of #1988 

This PR:
1. Crop some functions from `logrus`
2. Add logs(with context) for [Submit](https://github.com/sql-machine-learning/sqlflow/pull/2012/files#diff-45b52a537a003bab57099e84c65c0985R80-R85)/[Fetch](https://github.com/sql-machine-learning/sqlflow/pull/2012/files#diff-b03e31960a96575596cdfc09b491fefaR75-R81) processes.


cmd/sqlflow/main_test.go
```diff
@@ -41,7 +42,9 @@ USING my_model;
 func TestWorkflow(t *testing.T) {
 	a := assert.New(t)
 	datasource := "mysql://root:root@tcp(127.0.0.1:3306)/?maxAllowedPacket=0"
-	code, e := compile("couler", testSQLProgram, datasource)
+	codegen := "couler"
+	log := log.WithFields(log.Fields{"codegen": codegen})
```
typhoonzero: 1. Better to use the same scheme when calling the `log` package, use package function or use the object. Mixing two usages is a little bit confusing.
1. The object name should be `logger` not `log`.

suggest:

1. logging with no fields:
```go
logger := log.GetDefaultLogger()
logger.Info(...)
// or construct a default global logger variable
```
1. logging with fields

```sql
logger := log.WithFields(...)
logger.Info(...)
```

weiguoz: Thanks for the suggestion which makes the log easier.

shendiaomo:
Do we need to add a unit test to check the output of the log function? Just like what `repl_test.go` does.

weiguoz:
> eck the output of the log function? Just like what `repl_test.go` do

Yes, we need.
I'm working on another feature about the log, which requires the unit tests very much. 
It seems to need some tricky. I will throw out the unit tests in that PR.

---
updated: [UT for log](https://github.com/sql-machine-learning/sqlflow/pull/2027/files#diff-ebcc52a256f6cb69fb37d20e42464476)

typhoonzero (approved):
LGTM++

PR #2019, opened 2020-03-17
Support using external memory to train XGBoost models

shendiaomo:
This is for training data that is bigger than memory. 
See https://xgboost.readthedocs.io/en/latest/tutorials/external_memory.html as a reference.

### DIFF
--- a/doc/model_parameter.md
+++ b/doc/model_parameter.md
@@ -159,6 +159,11 @@ INTO sqlflow_models.my_xgb_regression_model;
 	<td>Subsample ratio of the training instance.</td>
 </tr>
 <tr>
+	<td>train.cache</td>
+	<td>bool</td>
+	<td>whether use external memory to cache train data</td>
+</tr>
+<tr>
 	<td>train.num_boost_round</td>
 	<td>int</td>
 	<td>[default=10]<br>The number of rounds for boosting.<br>range: [1, Infinity]</td>
--- a/doc/model_parameter.md
+++ b/doc/model_parameter.md
@@ -159,7 +159,7 @@ INTO sqlflow_models.my_xgb_regression_model;
 	<td>Subsample ratio of the training instance.</td>
 </tr>
 <tr>
-	<td>train.cache</td>
+	<td>train.disk_cache</td>
 	<td>bool</td>
 	<td>whether use external memory to cache train data</td>
 </tr>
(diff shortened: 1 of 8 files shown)


doc/model_parameter.md
```diff
@@ -158,6 +158,11 @@ INTO sqlflow_models.my_xgb_regression_model;
 	<td>float32</td>
 	<td>Subsample ratio of the training instance.</td>
 </tr>
+<tr>
+	<td>train.cache</td>
```
typhoonzero: Attributes start with `train.` is not used for model definitions, can this be defined only in codegen?

pkg/sql/codegen/xgboost/codegen.go
```diff
@@ -37,6 +37,7 @@ range: [0,1]`, attribute.Float32RangeChecker(0, 1, true, true)},
 range: [2, Infinity]`, attribute.IntLowerBoundChecker(2, true)},
 	"objective":   {attribute.String, nil, `Learning objective`, objectiveChecker},
 	"eval_metric": {attribute.String, nil, `eval metric`, nil},
+	"train.cache": {attribute.Bool, false, `whether use external memory to cache train data`, nil},
```
typhoonzero: Seems the word cache stands for caching things in memory generally. Maybe the option name should be like `disk_cache` etc.

pkg/sql/codegen/xgboost/template_train.go
```diff
@@ -50,6 +51,7 @@ train(datasource='''{{.DataSource}}''',
         feature_column_names=feature_column_names,
         label_meta=label_meta,
         validation_select='''{{.ValidationSelect}}''',
+		cache="{{.Cache}}" == "true",
```
weiguoz: indentation

python/sqlflow_submitter/xgboost/predict.py
```diff
@@ -36,7 +36,8 @@ def pred(datasource,
         conn = db.connect_with_data_source(datasource)
     label_name = label_meta["feature_name"]
     dpred = xgb_dataset(datasource, 'predict.txt', select, feature_metas,
-                        feature_column_names, None, is_pai, pai_table, True)
+                        feature_column_names, None, is_pai, pai_table, True,
+                        True)  # NOTE: default to use external memory
```
weiguoz, line 40: I'm not sure the reason for setting `True` as the default value.

pkg/sql/codegen/xgboost/codegen.go
shendiaomo: Done.

python/sqlflow_submitter/xgboost/predict.py
shendiaomo: Because we cannot specify the attributed in the predict statement, using `True` as the default is to make the predict statement being capable of processing large datasets automatically.

pkg/sql/codegen/xgboost/template_train.go
shendiaomo: Done.

weiguoz (approved):
LGTM ++

Issue #2016, opened 2020-03-17
Analyze SQL program dependency on table level to enable run SQL program as a DAG workflow


typhoonzero:
Design doc: #2000

implementation works:

- [x] update external parser gRPC protobuf to return input/output tables of a statement
- third-party parser
    - [ ] return input/output tables result in Hive parser
    - [ ] return input/output tables result in calcite parser
    - [x] return input/output tables result in antlr parser
    - [ ] return input/output tables result in MySQL parser
- [ ] construct graph in Go
- [ ] Couler support define DAG steps
- [ ] generate DAG Argo workflow using the graph and Couler

PR #2029, opened 2020-03-19
Get iotables from java parsers

typhoonzero:
Part work of https://github.com/sql-machine-learning/sqlflow/issues/2016

### DIFF
--- a/pkg/parser/external/parser.go
+++ b/pkg/parser/external/parser.go
@@ -18,7 +18,13 @@ import "fmt"
 // Parser abstract a parser of a SQL engine, for example, Hive, MySQL,
 // TiDB, MaxCompute.
 type Parser interface {
-	Parse(program string) ([]string, int, error)
+	Parse(program string) ([]string, []*InputOutputTables, int, error)
+}
+
+// InputOutputTables record tables that a SQL statement reads and writes.
+type InputOutputTables struct {
+	InputTables  []string
+	OutputTables []string
 }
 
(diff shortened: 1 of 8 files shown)


pkg/parser/external/parser.go
```diff
@@ -18,7 +18,13 @@ import "fmt"
...
 // TiDB, MaxCompute.
 type Parser interface {
-	Parse(program string) ([]string, int, error)
+	Parse(program string) ([]string, []*InputOutputTables, int, error)
+}
+
+// InputOutputTables record tables that a SQL statement reads and writes.
+type InputOutputTables struct {
+	InputTables  []string
+	OutputTables []string
```
wangkuiyi: Instead of adding type `InputOutputTables`, I prefer to add a type `Statement` as in https://play.golang.org/p/R-bI_iy7blg.

In particular,

```go
type Statement struct {
	string // the literal statement
	inputs  []string // input tables
	outputs []string // output tables
}
```

My preference comes from the fact that when I read the signature 

```
Parse(program string) ([]string, []*InputOutputTables, int, error)
```

I cannot perceive the rule that each element in `[]string` has a correspondence in `[]*InputOutputTables`.  Whereas the type `parser/external.Statement` corresponds the statement content with its inputs/outputs, and we can have

```go
Parse(program string) ([]Statement, int, error)
```

wangkuiyi (approved):

PR #2026, opened 2020-03-18
Refactor tensorflow code

typhoonzero:
Refactor TensorFlow code to make it easy to read:

1. Move the training code for Estimator and Keras to separated Python files.
1. Put log level setting function to a single Python file.
1. Put get Tensorflow version function to a single Python file.
1. Add general `get_dataset_fn` to generate dataset for every training usage.

PR #2034, opened 2020-03-19
Add Python linter

### LINKED ISSUE #2033: Add Python linter and beautifier
So to check Python source code in this repo.

Fixing this issue doesn't check the Python generated by SQLFlow, which is tracked in https://github.com/sql-machine-learning/sqlflow/issues/2032

As a reference:

https://github.com/sql-machine-learning/elasticdl/blob/95c74b28c867dd5986724ad77d7802386624528d/.pre-commit-config.yaml#L1-L23

wangkuiyi:
Fix https://github.com/sql-machine-learning/sqlflow/issues/2033 by ading pylint check to pre-commit.

This change introduce our own bash script scripts/pre-commit/pylint.sh that calls pylint only with changed files. Without this script, we would have to fix too many Python source files to make them pylint-compatible before we can merge this PR.

### DIFF
--- a/scripts/docker/install-python.bash
+++ b/scripts/docker/install-python.bash
@@ -47,7 +47,6 @@ pip install \
     plotille==3.7 \
     seaborn==0.9.0 \
     jsbeautifier \
+    yapf \
     isort \
-    black \
-    seed-isort-config \
-    mypy
+    pylint
--- a/scripts/docker/install-python.bash
+++ b/scripts/docker/install-python.bash
@@ -30,8 +30,6 @@ apt-get install -y build-essential  libssl-dev # for building mysqlclient pip
 # NOTE: shap == 0.30.1 depends on dill but not include dill as it's dependency, need to install manually.
 # NOTE: mysqlclient depends on apt-get install mysqlclient in install-mysql.bash.
 pip install \
-    yapf==0.29.0 \
-    isort==4.3.21 \
     numpy==1.16.2 \
     tensorflow==2.0.1 \
     mysqlclient==1.4.4 \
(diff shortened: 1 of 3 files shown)


tonyyang-svail:
@wangkuiyi, the linter check fails on some Python files, mostly code formatting and comment requirements. Since there is an ongoing code refactor by @typhoonzero at https://github.com/sql-machine-learning/sqlflow/pull/2026, how about we waiting for the merge of https://github.com/sql-machine-learning/sqlflow/pull/2026 then I will help to make sure all Python code satisfies the linter?

wangkuiyi:
@tonyyang-svail Without the linter, I am afraid that refactorization work would introduce new incompatibilities. Let us find a way to decompose the work of adding lintering into smaller pieces.

tonyyang-svail:
> Let us find a way to decompose the work of adding lintering into smaller pieces.

Sure.

tonyyang-svail (dismissed):

wangkuiyi:
@tonyyang-svail I reduced this change to add only pylint pre-commit hook, and only check changed files.

Since we already have yapf and isort, I don't think black and flake are must-to-have.

pylint doesn't accept `from some_pacakge import *`. So, if we change some Python file in the future, we will need to fix such imperfections.

scripts/docker/install-python.bash
```diff
@@ -48,4 +46,7 @@ pip install \
     oss2==2.9.0 \
     plotille==3.7 \
     seaborn==0.9.0 \
-    jsbeautifier
+    jsbeautifier \
+    yapf \
+    isort \
```
tonyyang-svail, line 51: Do we need to follow the version number of the original code?

tonyyang-svail (approved):

scripts/docker/install-python.bash
wangkuiyi: I had the same question. I feel that we can try sticking with the most recent version and see if their upgrading would make incompatible reformat suggestions.

PR #2043, opened 2020-03-20
Add model market design and update model zoo design with it

typhoonzero:
Update model zoo design and add model market design


doc/design/model_market.md
```diff
@@ -0,0 +1,119 @@
...
+**NOTE: Docker image with different [tags](https://www.freecodecamp.org/news/an-introduction-to-docker-tags-9b5395636c2a/) will be recognized as different images.
+
+## Steps to Share Model Definition Docker Images to Other Users
+
+1. Login to model market.
+1. Go to the "Model Definitions" tab.
+1. Find a Docker image in the list to be shared and click on the button "share".
+1. Input the user name or user ID to share to and click "OK".
+
+The system will call Docker registry API to grant access for the user shared to after these operations. The model market will save the grant information in a database table `image_shares` that have below columns:
```
skydoorkai, line 99: Is it allowed for other users to download this Docker image, or this image can only be used in the current SQLFlow deployment?

typhoonzero: User's Docker image pull access is only granted to the SQLFlow account by adding this image into model market. Other users do not have access unless the current user share this image with him.

skydoorkai: If other users are granted access, they can download the image to their local machine?

typhoonzero: Yes, they can.

doc/design/model_market.md
```diff
@@ -0,0 +1,119 @@
...
+1. Manage user login and logout.
+1. View published images available for the current user.
+1. View published trained models available for the current user.
+1. Publish (or remove) model definition images.
+1. Publish (or remove) trained models.
+1. Share (or remove share) model definition images to other users.
+1. Share (or remove share) trained models to other users.
+
+<p align="center">
+<img src="figures/model_market_overview.png">
```
wangkuiyi: @tonyyang-svail Could we embed GraphViz plots in Markdown design docs on GitHub? It would be easier to do code review if the figure is in text format.

tonyyang-svail: You can find examples [here](https://github.com/TLmaK0/gravizo) to embed GraphViz plots.

doc/design/model_market.md
```diff
@@ -0,0 +1,119 @@
...
+</p>
+
+In order to support publishing and sharing models securely, the model market is able to communicate with some [SSO](https://en.wikipedia.org/wiki/Single_sign-on) service to authenticate users. Only users that are logged in can do operations on the model market.
+
+Then model definition images and trained model can have below accessibility settings:
+
+1. Private: Only visible to the current user.
```
wangkuiyi: Do we need "private"? I feel that we only need public or private (but shared to a specific group), like Github
provide public and private repos?

doc/design/model_market.md
```diff
@@ -0,0 +1,119 @@
...
+2. Public: Readable by every user.
+3. Private but shared to some users: visible to the current user and users that shared to.
+
+## Steps to View Model Definitions and Trained Models
+
+1. Login to model market.
```
wangkuiyi: Could users browse models without logging in?

doc/design/model_market.md
```diff
@@ -0,0 +1,119 @@
...
+1. Share (or remove share) model definition images to other users.
+1. Share (or remove share) trained models to other users.
+
+<p align="center">
+<img src="figures/model_market_overview.png">
+</p>
+
+In order to support publishing and sharing models securely, the model market is able to communicate with some [SSO](https://en.wikipedia.org/wiki/Single_sign-on) service to authenticate users. Only users that are logged in can do operations on the model market.
+
+Then model definition images and trained model can have below accessibility settings:
```
wangkuiyi: I am afraid that model definitions and trained models need to have different access control lists.  A model definition from a published paper is likely public, but a model trained with private data should be private.

typhoonzero: Yes, in this part I was trying to explain the types of visibility settings. Model definitions and trained models' accessibility settings indeed need to have different access control lists. I'll make it clear.

doc/design/model_market.md
```diff
@@ -0,0 +1,142 @@
...
+</p>
+
+1. Login to model market.
+1. Go to the "Model Definitions" tab.
+1. Click "New" to add a new image.
+1. Input the full Docker image address in the dialog and click "OK".
+1. ***Optional***: The system will call the Docker registry API to grant access for SQLFlow to pull the image. When using a public image, this step will be skipped. If access can be granted, a message should be shown on the web page.
+1. The system will start to run several checks and tests using the Docker image. If all the checks have passed, the image is added.
+
+**NOTE: Docker image with different [tags](https://www.freecodecamp.org/news/an-introduction-to-docker-tags-9b5395636c2a/) will be recognized as different images.
```
tonyyang-svail: Markdown formatting: this line should terminate with `**`.

doc/design/model_market.md
```diff
@@ -0,0 +1,142 @@
...
+
+1. `OwnerID`: Docker image owner user ID.
+1. `SharedUserID`: User ID that the image is shared to.
+
+## Steps to Publish a Trained Model
+
+The following SQL statement will publish a trained model named `my_first_model` to the model market.
+
+```sql
+SQLFLOW PUBLISH my_first_model
```
tonyyang-svail, line 109: Publishing a trained model may also need SSO. Consider a user types in this statement in the Jupyter Notebook or REPL, how does the SSO verification happen?

doc/design/model_market.md
```diff
@@ -0,0 +1,142 @@
...
+In the [model zoo design](model_zoo.md), we described how model developers develop, publish and share custom models on SQLFlow, and how analysts can make use of the shared model with SQLFlow.
+
+Here, we'll describe how to build the model market model developers and analysts.
+
+## Overview
+
+The model market is designed to:
+
+1. Manage user login and logout.
+1. View published images available for the current user.
```
skydoorkai: What is this image?  model definition image?
Also add some explanation about the model definition image.

doc/design/model_market.md
```diff
@@ -0,0 +1,142 @@
...
+## Steps to Share Trained Models to Other Users
+
+1. Login to model market.
+1. Go to the "Trained Models" tab.
+1. Find a trained model in the list to be shared and click on the button "share".
+1. Input the user name or user ID to share to and click "OK".
+
+The model market will save the grant information in a database table `traind_model_shares` that have below columns:
+
+1. `OwnerID`: trained model owner user ID.
```
brightcoder01: How about using a table to describe the DB schema？

doc/design/model_market.md
```diff
@@ -0,0 +1,142 @@
...
+
+In the [model zoo design](model_zoo.md), we described how model developers develop, publish and share custom models on SQLFlow, and how analysts can make use of the shared model with SQLFlow.
+
+Here, we'll describe how to build the model market model developers and analysts.
+
+## Overview
+
+The model market is designed to:
+
+1. Manage user login and logout.
```
brightcoder01: How do we implement the use authentication?

doc/design/model_market.md
```diff
@@ -0,0 +1,142 @@
...
+<p align="center">
+<img src="figures/list_model_defs.png">
+</p>
+
+<p align="center">
+<img src="figures/list_trained_model.png">
+</p>
+
+1. **Optional**: Login to model market. If the user is not logged in, he can only view public model definitions and trained models.
+1. Click at the "Model Definitions" tab to see the list of model definition Docker images and the model class names in each Docker image.
```
brightcoder01: Besides model class, should we also show the parameters of the model definition?
Such as the parameter `wide_embeddings` and `deep_embeddings` for the model `def wide_and_deep_classifier(wide_embeddings, deep_embeddings)`

doc/design/model_market.md
```diff
@@ -0,0 +1,142 @@
...
+```sql
+SQLFLOW PUBLISH my_first_model
+    [TO https://models.sqlflow.org/user_name]
+```
+
+By publishing a trained model, the model market will save the trained model weights together with ownership information into two database tables (can use a MySQL service in general): the **trained models table** and the **evaluation result table**.
+
+### Trained Models Table
+
+Once a training job completes, the submitter program adds/updates a row of the trained models' table, which contains (at least) the following fields.
```
workingloong, line 123: The table size may increase greatly if the submitter adds a row for each training job. Can we set a lifecycle for each trained model record? The table will drop the trained model record after the lifecycle if the owner dese not public or share the trained model.

doc/design/model_market.md
```diff
@@ -0,0 +1,142 @@
...
+
+1. The model ID (or model name), specified by the INTO clause, or `my_first_model` in the example at [model_zoo_design](model_zoo.md).
+1. The creator, the current user ID.
+1. The model zoo release, which is a Docker image commit ID, or `a_data_scientist/regressors` in the above example.
+1. The model definition, which is a Python class name, or `DNNRegressor` in the above example.
+1. The model weights file path, the path to the trained model parameters on the distributed filesystem of the cluster.
```
workingloong: The table may also need the training hyperparameters such as iteration steps, dropout rate, and hashing bucket size to preprocess string values.

brightcoder01: While executing the training pipeline, some vocabulary files will be generated during the data analysis step. The trained model also need contain the vocabulary file paths in the distributed filesystem.

doc/design/model_market.md
```diff
@@ -0,0 +1,142 @@
+# Model Market
+
+The model market is a web site where model developers can publish and share their models and analysts can find some useful models to finish the analysis work. The model market can be deployed anywhere like on the cloud or on-premise with some configurations.
+
+In the [model zoo design](model_zoo.md), we described how model developers develop, publish and share custom models on SQLFlow, and how analysts can make use of the shared model with SQLFlow.
+
+Here, we'll describe how to build the model market model developers and analysts.
```
QiJune: build the model market for model developers and analysts.

doc/design/model_zoo.md
```diff
@@ -60,97 +60,99 @@ In SQLFlow SQL grammar, the identifiers after `TRAIN`, `USING` and `INTO` have d
...
+INTO my_first_model;
 ```
 
-The following statement queries all model zoos used by `an_analyst` to train his/her models.
+Then the generated workflow will use the Docker image `a_data_scientist/regressors:v0.2` to run the statement. In this step, SQLFlow will generate the training Python program inside the Docker container and execute it. Once the job completes, **the trained model together with hyperparameters, Docker image, evaluation result and the SQL statement used will be saved.** So in one trained model, we can have:
 
-```sql
-SELECT DISTINCT model_zoo FROM sqlflow.trained_models WHERE creator="an_analyst"
-```
+1. Trained model weights.
```
brightcoder01, line 125: I think we need also store the generated transform code built using `feature column` or `keras preprocessing layers` in the trained model.
The generated code not only contains the data transform logic but also the data analysis result such as `bucketize boundaries`, `hash_bucket_size`, `min/max` and so on.

doc/design/model_zoo.md
```diff
@@ -60,97 +60,99 @@ In SQLFlow SQL grammar, the identifiers after `TRAIN`, `USING` and `INTO` have d
...
-SELECT * FROM sqlflow.trained_models WHERE creator="an_analyst"
+SELECT * FROM employee WHERE onboard_year < 2019
+TO TRAIN a_data_scientist/regressors:v0.2/MyDNNRegressor
+WITH model.hidden_units=[1024,256],
+LABEL salary
+INTO my_first_model;
 ```
 
-The following statement queries all model zoos used by `an_analyst` to train his/her models.
+Then the generated workflow will use the Docker image `a_data_scientist/regressors:v0.2` to run the statement. In this step, SQLFlow will generate the training Python program inside the Docker container and execute it. Once the job completes, **the trained model together with hyperparameters, Docker image, evaluation result and the SQL statement used will be saved.** So in one trained model, we can have:
```
brightcoder01, line 123: Then the generated workflow will use the Docker image `a_data_scientist/regressors:v0.2` to run the statement. In this step, SQLFlow will generate the training Python program inside the Docker container and execute it.
=> 
I think it's training on a single node. How about distributed training?

doc/design/model_market.md
typhoonzero: Updated to "model definition Docker images". The details are explained in `model_zoo.md`

doc/design/model_market.md
typhoonzero: The trained model record will only be added when a user executes `SQLFLOW PUBLISH ...` to publish a model. Publishing a model is done only when the trained model can be reused on other datasets or and the trained model is already getting pretty good performance (accuracy). So trained models in the list will not be too much and users can delete the model if the model is no longer needed.

doc/design/model_market.md
```diff
@@ -0,0 +1,177 @@
...
+
+The below figure shows the relationships between entities of the model market:
+
+1. Each user can create many organizations, each organization can be accessed by many users.
+1. Each organization can manage several model definitions and trained models.
+1. Each model definition have many versions.
+1. Every trained model is trained using a specific version of a model definition, yet one model definition with a specific version can be used to train many trained models.
+
+<p align="center">
+<img src="figures/model_market_er.png">
```
wangkuiyi, line 39: I cannot see this figure. Is that because you forgot to commit it?

wangkuiyi (approved):

PR #2066, opened 2020-03-26
Add model_market_er.md

wangkuiyi:


wangkuiyi:
gravizo cannot render the GraphViz plot correctly with `shape=record`.

PR #2069, opened 2020-03-27
Update the repl.md a little bit

linkerzhang:
Per my fresh experiences, 1) added a step to populate example data for running examples; 2) remove the message that is not shown now when training a classifier.

### DIFF
--- a/doc/run/repl.md
+++ b/doc/run/repl.md
@@ -9,8 +9,8 @@ In addition to building SQLFlow into a gRPC server, accessed via Jupyter Noteboo
 The SQLFlow Docker image contains the REPL command-line program.  We can run an example session by typing the following command on MacOS.  If you run Docker on Linux, please change `host.docker.internal:3306` to `localhost:3306`.
 
 ```
-docker run -it --rm --net=host sqlflow/sqlflow repl \
---datasource="mysql://root:root@tcp(host.docker.internal:3306)/?maxAllowedPacket=0"
+docker run -it --rm --net=host sqlflow/sqlflow bash -c 'chmod +x start.sh; ./start.sh populate-example-dataset-mysql-local; \
+repl --datasource="mysql://root:root@tcp(host.docker.internal:3306)/?maxAllowedPacket=0"'
 ```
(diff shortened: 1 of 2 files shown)


doc/run/repl.md
```diff
@@ -9,8 +9,8 @@ In addition to building SQLFlow into a gRPC server, accessed via Jupyter Noteboo
 The SQLFlow Docker image contains the REPL command-line program.  We can run an example session by typing the following command on MacOS.  If you run Docker on Linux, please change `host.docker.internal:3306` to `localhost:3306`.
 
 ```
-docker run -it --rm --net=host sqlflow/sqlflow repl \
---datasource="mysql://root:root@tcp(host.docker.internal:3306)/?maxAllowedPacket=0"
+docker run -it --rm --net=host sqlflow/sqlflow bash -c 'chmod +x start.sh; ./start.sh populate-example-dataset-mysql-local; \
```
wangkuiyi: I think we can do chmod + x to the start.sh in the Git repo, so we don't need to chmod +x at the runtime; am I correct?

Yancey0623 (dismissed):
LGTM

doc/run/repl.md
```diff
@@ -48,7 +48,6 @@ LABEL class
 INTO sqlflow_models.my_dnn_model;
```
typhoonzero, line 47: Maybe this SQL statement is kind of outdated. We need to write:

```sql
SELECT * FROM iris.train
TO TRAIN DNNClassifier
WITH model.n_classes=3, model.hidden_units=[128,64],
    validation.select="SELECT * FROM iris.test"
LABEL class
```

now.

wangkuiyi: Thanks for the reminder. I updated the document.

linkerzhang: thanks @typhoonzero for the comments and @wangkuiyi for the help of resolving the comment!

typhoonzero (approved):

Yancey0623 (approved):
LGTM++

PR #2076, opened 2020-03-30
Support .env file to collect environment variables into a config file

### LINKED ISSUE #2075: Move environment variables into a configure file
We have lots of environment variables that are hard for users to memorize and maintain.

1. Support a config file based on JSON/YAML. Or
2. At least support a `.env` file.

shendiaomo:
Fix #2075 . This is a quick and simple fix. More advanced ways for management of the config files are welcomed.

### DIFF
--- a/cmd/repl/repl.go
+++ b/cmd/repl/repl.go
@@ -29,6 +29,7 @@ import (
 	"strings"
 	"syscall"
 
+	"github.com/joho/godotenv"
 	"golang.org/x/crypto/ssh/terminal"
 	"sqlflow.org/sqlflow/pkg/database"
 	"sqlflow.org/sqlflow/pkg/step"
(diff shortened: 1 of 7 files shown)


cmd/repl/repl.go
```diff
@@ -240,6 +241,7 @@ func getDataSource(dataSource, db string) string {
 var currentDB string
 
 func main() {
+	step.InitEnvFromFile(filepath.Join(os.Getenv("HOME"), step.DotEnvFilename))
```
typhoonzero: Seems this file needs to be created in every step container but it seems Couler can not do that now.

typhoonzero (approved):

shendiaomo:
Thanks @typhoonzero 
This fix is now only used for REPL in an all-in-one script that is in progress.

PR #2084, opened 2020-04-01
Update contribute models doc add notes about not implemented feature

typhoonzero:


doc/contribute_models.md
```diff
@@ -83,6 +83,8 @@ In the final step, you need to publish your model so that other SQLFlow users ca
     docker push your-registry.com/model_image
     ```
     1. Then use the model image in SQLFlow by adding the Docker image name before the model name:
+    **NOTE: Below statement will not work in non-workflow mode (e.g. local run), this feature will be supported in the future.**
```
wangkuiyi, line 86: If they do not work (yet), how about removing them?

wangkuiyi (approved):

doc/contribute_models.md
typhoonzero: I think we do need a document describing the desired steps to publish a model. I'll file an issue and add this feature ASAP.

PR #2088, opened 2020-04-02
Estimator regressor explain

shendiaomo:
1. The original code explains multiclass Classifier model like:
image
After this fix, the `bar` figure will be properly rendered like:
image

2. Use the `predictions` fields to explain regressors

### DIFF
--- a/python/sqlflow_submitter/tensorflow/explain_example.py
+++ b/python/sqlflow_submitter/tensorflow/explain_example.py
@@ -96,3 +96,36 @@ if __name__ == "__main__":
             plot_type='bar',
             result_table="")
     shutil.rmtree("dnnmodel")
+
+    # Train and explain DNNRegressor
+    train(datasource=datasource,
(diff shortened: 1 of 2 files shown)


python/sqlflow_submitter/tensorflow/explain_example.py
```diff
@@ -96,3 +96,36 @@
...
+          model_params={
+              "hidden_units": [100, 100],
+          },
+          save="dnnmodel",
+          batch_size=100,
+          epoch=20,
+          verbose=0)
+
+    explain(datasource=datasource,
+            estimator_cls=tf.estimator.DNNRegressor,
```
typhoonzero, line 118: Maybe also add test case for `LinearRegressor`?

shendiaomo: Done.

typhoonzero (approved):

PR #2091, opened 2020-04-03
Support more show statements

shendiaomo:


pkg/sql/executor_normal_stmt.go
```diff
@@ -38,7 +47,7 @@ func isQuery(slct string) bool {
 	if strings.HasPrefix(s, "SELECT") && !has(s, "INTO") {
 		return true
 	}
-	if strings.HasPrefix(s, "SHOW") && (has(s, "CREATE") || has(s, "DATABASES") || has(s, "TABLES")) {
+	if strings.HasPrefix(s, "SHOW") && containsAnyString(s, "CREATE", "DATABASES", "TABLES", "COLUMNS", "TABLE", "PROCESSLIST") {
```
tonyyang-svail: For discussion, maybe we can remove this `isQuery` check and forward all standard SQL to `db.Query`? I am not sure if `db.Exec` is useful...

shendiaomo: `db.Exec` returns a https://golang.org/pkg/database/sql/#Result object, which seems cannot be acquired from a call to `db.Query`. This is the only difference I can found.
The [MySQL C++ client](https://dba.stackexchange.com/questions/24542/execute-vs-executequery-in-mysql-connector-c) recommends we call `execute` to check whether a statement is an INSERT, UPDATE, or DELETE statement. Unfortunately, the go-database module doesn't provide a similar mechanism.

pkg/sql/executor_normal_stmt.go
```diff
@@ -31,17 +31,16 @@ func runNormalStmt(wr *pipe.Writer, slct string, db *database.DB) error {
...
 // https://golang.org/pkg/database/sql/#DB.Exec .
-// We will need to extend our parser to be a full SQL parser in the future.
-func isQuery(slct string) bool {
-	s := strings.ToUpper(strings.TrimSpace(slct))
-	has := strings.Contains
-	if strings.HasPrefix(s, "SELECT") && !has(s, "INTO") {
+// The implementation is based on
+// https://github.com/xo/usql/blob/490ef67b5accd38d53cbe0a333bef85480a0a716/drivers/qtype.go#L8-L22
+// We may need to call SQLFlow parser to check the statement type in the future.
+func isQuery(stmt string) bool {
```
wangkuiyi, line 37: @shendiaomo `isQuery` is a hack according to the function comment. It means that we need to get rid of it ASAP. It is not a good idea to implement any new idea by extending it.

Suppose that the user writes a SQL statement like

```sql
SELECT * FROM sqlflow_history WHERE stmt="SELECT * FROM a TO TRAIN b INTO c";
```

the following condition would easily get confused:

```go
if strings.HasPrefix(s, "SELECT") && !strings.Contains(s, "INTO") {
```

wangkuiyi (commented):
@shendiaomo Could you please think about how to get rid of the hack of `isQuery` from the codebase? Thanks!

pkg/sql/executor_normal_stmt.go
shendiaomo: I'll check it out.

PR #2096, opened 2020-04-07
Update contribute models doc

typhoonzero:


doc/contribute_models.md
```diff
@@ -41,9 +41,65 @@ Open the browser and go to http://localhost:8888, it's a Jupyter notebook enviro
...
+
+If you need to control the details of the training process or define custom models rather than a Keras model, you can define a function `sqlflow_train_loop` to implement custom model training processes:
+
+```python
+class MyAwesomeClassifier(tf.keras.Model):
+    def __init__(self, feature_columns=None):
+        ...
+    def call(self, inputs, training=True):
+        ...
+    def sqlflow_train_loop(self, dataset, epochs=1, verbose=0):
```
wangkuiyi, line 98: Need to justify why do we need this method? I am not sure if ElasticDL can call this function.

typhoonzero: Model developers may need to develop models other than a Keras model, e.g. models using sklearn.

wangkuiyi: What is the relationship between a method in a Keras class with sklearn?

typhoonzero: We use Keras subclass model API as SQLFlow custom model API. Yet, if we need to support custom models like SVM, kmeans or combining several models and train the model with custom training procedures like in https://github.com/sql-machine-learning/models/blob/develop/sqlflow_models/deep_embedding_cluster.py, we need a function that defines this procedure.

wangkuiyi (approved):

PR #2099, opened 2020-04-08
Support basic keywords highlighting

shendiaomo:
The result looks like:
image
Later I'll try supporting better syntax highlighting (literals, comments, function calls, etc.), but I can't promise because this actually exceeds the original ability of `go-prompt`.

### DIFF
--- a/cmd/repl/output.go
+++ b/cmd/repl/output.go
@@ -1,0 +1,103 @@
+// Copyright 2020 The SQLFlow Authors. All rights reserved.
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"github.com/c-bata/go-prompt"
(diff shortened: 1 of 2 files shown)


cmd/repl/output.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+// newStdoutWriter returns a ConsoleWriter object to write to stdout.
+func newStdoutWriter() *stdoutWriter {
+	return &stdoutWriter{ConsoleWriter: prompt.NewStdoutWriter()}
+}
+
+// WriteStr writes safety string by removing control sequences and highlighting keywords.
+func (w *stdoutWriter) WriteStr(data string) {
+	words := strings.Split(data, " ")
```
lhw362950217, line 67: maybe we should support '\t' as well

cmd/repl/output.go
```diff
@@ -0,0 +1,103 @@
...
+
+// SetColor sets text and background colors. and specify whether text is bold.
+func (w *stdoutWriter) SetColor(fg, bg prompt.Color, bold bool) {
+	w.ConsoleWriter.SetColor(fg, bg, bold)
+	w.currentFgColor = fg
+	w.currentBgColor = bg
+	return
+}
+
+// isInCompletion checks whether we're rendering the input line
```
lhw362950217: isInCompletion -> isInputLine

cmd/repl/output.go
shendiaomo: `\t` is now a special character that activates auto-complete, users cannot input `\t` directly.

cmd/repl/output.go
shendiaomo: Done.

lhw362950217 (approved):
LGTM

PR #2102, opened 2020-04-09
Fix xgboost dtype

typhoonzero:
Fix XGBoost writes to predict result table when predict data have int typed values.


Yancey0623 (approved):
LGTM

PR #2103, opened 2020-04-09
Advanced syntax highlighting

shendiaomo:
This is a follow-up to #2099 
The implementation uses https://github.com/alecthomas/chroma to highlight the inputs and convert it to a colored ASCII sequence that is written to stdout using `go-prompt`.
There're several pitfalls:
- strings across multiple lines are not highlighted correctly, because we can only get the current line from `go-prompt`
- the syntax is hard-coded to `MySQL`. I'll fix this later in some way
- because the highlighting is implemented in the bottom level, prompts in history searching mode are highlighted too

Everything else worked well. The result looks like:
image


cmd/repl/output.go
```diff
@@ -25,35 +27,11 @@ type stdoutWriter struct {
...
-	"TABLES":    true,
-	"TO":        true,
-	"UPDATE":    true,
-	"USE":       true,
-	"USING":     true,
-	"VALUES":    true,
-	"WHERE":     true,
-	"WITH":      true,
-}
 var extendedKeywords = map[string]bool{
```
weiguoz, line 30: Shall we highlight the `column`, `with` as the extended keywords?

shendiaomo: I'm not sure哈. Unlike `TO`, `COLUMN` and `WITH` are used very often in standard SQL. Let's discuss this tomorrow.

brightcoder01: I think that the highlight feature is for SQLFlow, more than MySQL. As a result, highlighting the keyword `TO`, `COLUMN` and `WITH` for SQLFlow would be more user-friendly.

shendiaomo: Done.

weiguoz (approved):
LGTM

lhw362950217 (approved):
LGTM++

PR #2114, opened 2020-04-15
Add a wrapper bash script for terminal users

shendiaomo:


scripts/terminal/sh/sqlflow
```diff
@@ -0,0 +1,84 @@
...
+        FN=$(realpath -- $FN)
+        VARG="$VARG -v $FN:/root/input.sql"
+        PARAMS="$PARAMS -f /root/input.sql"
+        shift 1
+      else
+        echo "sqlflow: $2: No such file or directory"
+        exit 1
+      fi
+      ;;
+    --datasource)
```
wangkuiyi, line 57: Need a Markdown tutorial explaining how to use this command-line tool.

scripts/start.sh
```diff
@@ -98,6 +115,22 @@ function main() {
...
+      shift
+      . ~/.sqlflow_env
+      if [[ $SQLFLOW_DATASOURCE == "" ]]; then
+        echo "No data source specified. Starting builtin MySQL"
+        start_mysql
+        sleep_until_mysql_is_ready
+        SQLFLOW_DATASOURCE="mysql://root:root@tcp(${SQLFLOW_MYSQL_HOST}:${SQLFLOW_MYSQL_PORT})/?maxAllowedPacket=0"
+	  fi
+	  export SQLFLOW_DATASOURCE
+      if [[ ! -d /var/lib/mysql/iris && $SQLFLOW_DATASOURCE == mysql://* ]]; then
```
wangkuiyi, line 128: Does this mean that the command-line tool works only with MySQL?  Could it work with MaxCompute?

wangkuiyi (commented):
This approach differs from what was in my mind about the client tool.  I'd thought that we can run it as a statically-linked Go binary, instead of a Docker container.

scripts/terminal/sh/sqlflow
shendiaomo: We already have one, see https://github.com/sql-machine-learning/sqlflow/blob/develop/doc/run/repl.md. We just need to rename the document name and its title. Besides, there'll be a pull request about how to config the sqlflow client.

scripts/start.sh
shendiaomo: > Does this mean that the command-line tool works only with MySQL? Could it work with MaxCompute?

If the user doesn't create a config file after downloads. (Type `./sqlflow` and enter). The client will connect to the `mysqld` nested in the docker container. If the user has specified a connection string, the client will connect to the specified database server. The databases supported at the moment are MySQL, Hive, and MaxCompute.

wangkuiyi:
Giving that we are going to create a statically-linked client binary, as described in this [proposal](https://github.com/sql-machine-learning/sqlflow/pull/2136), should we close this PR?

shendiaomo:
> Giving that we are going to create a statically-linked client binary, as described in this [proposal](https://github.com/sql-machine-learning/sqlflow/pull/2136), should we close this PR?

Yes, it's closed.

PR #2118, opened 2020-04-15
Support to evaluate in the parser

typhoonzero:
Support evaluation statements that output evaluation result in a table:

```sql
SELECT * FROM validation_table
TO EVALUATE my_model
WITH validation.metrics="AUC,BinaryAccuracy"
LABEL class
INTO evaluation_result_table;
```


shendiaomo (commented):
https://github.com/sql-machine-learning/sqlflow/blob/2bc51c9684474158d9a293d316f59278fcd82dfd/cmd/repl/output.go#L30-L38
Please update this new keyword to the highlighter list.

shendiaomo (approved):

PR #2122, opened 2020-04-17
show train stmt syntax part

### LINKED ISSUE #2049: Allow users to query model meta information

lhw362950217:
Fixes #2049
This is the first part of our `SHOW TRAIN` stmt, it aims to support the syntax. The parse process modification will be in another review.

### DIFF
--- a/pkg/ir/ir.go
+++ b/pkg/ir/ir.go
@@ -20,6 +20,7 @@ type Executor interface {
 	ExecuteTrain(*TrainStmt) error
 	ExecutePredict(*PredictStmt) error
 	ExecuteExplain(*ExplainStmt) error
+	ExecuteShowTrain(*ShowTrainStmt) error
 }
 
 // SQLFlowStmt has multiple implementations: TrainStmt, PredictStmt, ExplainStmt and standard SQL.
@@ -169,3 +170,23 @@ func (sql *NormalStmt) IsExtended() bool { return false }
 
 // GetOriginalSQL returns the original SQL statement used to get current IR result
 func (sql *NormalStmt) GetOriginalSQL() string { return string(*sql) }
+
+// ShowTrainStmt get and output the original train sql for ModelName
+type ShowTrainStmt struct {
+	// OriginalSQL is the SHOW TRAIN stmt itself
+	OriginalSQL string
+	// The model to show the train sql
+	ModelName string
+}
+
+// Execute generates and executes code for ShowTrainStmt
+func (sql *ShowTrainStmt) Execute(s Executor) error { return s.ExecuteShowTrain(sql) }
+
+// SetOriginalSQL sets the original sql string
+func (sql *ShowTrainStmt) SetOriginalSQL(s string) { sql.OriginalSQL = s }
+
+// IsExtended returns whether a SQLFlowStmt is an extended SQL statement
+func (sql *ShowTrainStmt) IsExtended() bool { return true }
+
+// GetOriginalSQL returns the original SQL statement used to get current IR result
+func (sql *ShowTrainStmt) GetOriginalSQL() string { return sql.OriginalSQL }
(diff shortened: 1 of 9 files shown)


shendiaomo (dismissed):

pkg/parser/extended_syntax_parser.y
```diff
@@ -158,6 +166,12 @@ sqlflow_select_stmt
 		Train: true,
 		TrainClause: $1}
   }
+| show_train_clause end_of_stmt {
+	parseResult = &SQLFlowSelectStmt{
+		Extended: true,
+		ShowTrain: true,
+		ShowTrainClause: $1}
+}
```
wangkuiyi: Adding the SHOW TRAIN clause here would enable users to write 

```sql
SELECT ... SHOW TRAIN trained_model_name;
```

However, according to the design doc https://github.com/sql-machine-learning/sqlflow/pull/2073/files#diff-49e684e27ee27bc9f191167dcd303674R28, we don't need the SELECT ... prefix.

```sql
SHOW TRAIN trained_model_name;
```

How are we going to have it? Is that the realization of the idea depends on the improvement of the collaborative parsing algorithm as the pseudo-code shows: https://github.com/sql-machine-learning/sqlflow/pull/2073/files#diff-49e684e27ee27bc9f191167dcd303674R46

wangkuiyi (changes requested):

pkg/ir/ir.go
```diff
@@ -169,3 +170,23 @@ func (sql *NormalStmt) IsExtended() bool { return false }
...
+// ShowTrainStmt get and output the original train sql for ModelName
+type ShowTrainStmt struct {
+	// OriginalSQL is the SHOW TRAIN stmt itself
+	OriginalSQL string
+	// The model to show the train sql
+	ModelName string
+}
+
+// Execute generates and executes code for ShowTrainStmt
+func (sql *ShowTrainStmt) Execute(s Executor) error { return s.ExecuteShowTrain(sql) }
```
wangkuiyi, line 183: I vaguely remember that I had a discussion with @typhoonzero a few months ago about do we need the Go interface in the design of IR. I don't member the conclusion and I think I might have preferred not to use interface.  Do you @typhoonzero remember the conclusion?

typhoonzero: Is the discussion details in https://github.com/sql-machine-learning/sqlflow/issues/1434? The current `Executor` interface still needed because we  have different execution types:

1. local: run codegen and execute the generated Python code locally.
1. pai: generate code and tar the code into a tarball then call PAI command to submit a job
1. alisa: generate code and tar the code into a tarball and upload this tarball to OSS then call alias HTTP API call to submit a job.

So we should have one interface to derive these submitting operations and we should move these code out of `ir.go`. In current implementation, should move `ShowTrainStmt` 's member functions to `submitter.go`.

By the way, the current naming is quite confusing, e.g. "submitter" should be a Python API, it should not appear in Go. `Executor` is also confusing, we should go on work with https://github.com/sql-machine-learning/sqlflow/issues/1434

pkg/parser/extended_syntax_parser.y
lhw362950217: This is the first part of our modification, we just add syntax rule here. In the second part, we will change the behavior of collaborative parsing to exlude this case.

wangkuiyi (dismissed):

shendiaomo (approved):
Great! lgtm++

PR #2136, opened 2020-04-21
Add a design doc for decoupling the client code out of SQLFlow Core

### LINKED ISSUE #1744: Proposal: Make repl call SQLFlow server
Currently, the binary repl call [`func RunSQLProgram`](https://github.com/sql-machine-learning/sqlflow/blob/5ab490f740decbb8c3f293461c7920d5ef2f8979/pkg/sql/executor_ir.go#L52) locally.

I propose to 

1. Rename repl into sqlflow.
1. Make repl call SQLFlow server, just like mysql calls MySQL server.
1. Move part of `RunSQLProgram` in `executor_ir.go` to `pkg/server` and the rest to `pkg/step/executor/`.  The former part calls the later part.

BTW, we need `cmd/step`, which calls `pkg/step/executor/` and is called by the generated workflow to run each step, as explained in https://github.com/sql-machine-learning/sqlflow/issues/1746.

A side-effect of this proposal is to reduce and finally remove the file `execute_ir.go`, as the IR is not executable -- it is the submitter program generated from the IR which is executable.

shendiaomo:
Fix https://github.com/sql-machine-learning/sqlflow/issues/1744

### DIFF
--- a/doc/design/cli.md
+++ b/doc/design/cli.md
@@ -2,43 +2,34 @@
 
 ## Overview
 
-At the moment, users can build SQLFlow into a server or a [command line binary](../run/repl.md). Both of the executables have all the functionalities of SQLFlow, and both of them can only run in a Docker container.
+At the moment, users can build SQLFlow into a gRPC server or a [command line binary](../run/repl.md).  Both forms statically link to the SQLFlow core code and inherit all dependencies, including TensorFlow, various database client libraries, and various SQL dialect parsers in the form of gRPC backend servers. The many dependencies require us to release and run both forms in Docker containers.
 
-For a deployment of SQLFlow in a production environment, the SQLFlow server runs in a Kubernetes cluster. Users access SQLFlow from a Web IDE or the Jupyter Notebook.
+For convenient deployment, we want to run the SQLFlow server in containers, and end users access the server via the command-line client. In particular, we want it easy to install the client tool, so we need to prune the dependencies. As a solution, we make the client tool a statically linked Go binary that remotely calls the SQLFlow server, so users can just download and run it.
 
-For an individual developer, one may deploy SQLFlow on her personal computer as described in [Run SQLFlow Using Docker](../run/docker.md)
-
-For the above two typical scenarios, SQLFlow lacks an easy-to-use user interface for quick start, quick verification, or quick deployment. Jupyter Notebooks and web IDEs still have a certain learning curve that reduces the user experience of SQL programmers. 
-
-From another point of view, a CLI could overtake a GUI in user experience for a simple language as SQLFlow: it's easier to design, easier to implement, easier to obtain. For example, repetitive tasks can be simplified by line editing and history mechanisms. With a CLI program, a user can easily access all the power of SQLFlow from a terminal as:
-
-```bash
-$ sqlflow
-sqlflow> _
-```
-
-To achieve this goal, we also want to hide implementation details about Docker because users don't have to know this. As a result, there are two possible solutions:
-1. Wrap the current SQLFlow REPL in a script, we already have such an [attempt](https://github.com/sql-machine-learning/sqlflow/pull/2114).
-1. Decouple the REPL implementation from the core code of SQLFlow and provide a native binary that only functions as a pure client.
-
-We decide to choose the second solution because it's more lightweight, if a user already has a SQLFlow deployment in her company, all she has to do is to download the native binary.
-
+The command-line client tool, named `sqlflow`, is complementary to other forms of clients, like Jupyter Notebook. Users can write shell or Python scripts calling the client tool to realize complex applications.
 ## Design
 
 ### Naming
 
-`REPL` is not a good name, We'll rename the binary as simply `sqlflow`.
+Currently, the command-line binary form of SQLFlow has the name `repl`, which is the name of a user interface design philosophy. Many command-line tools, including mysql and python, implement the UI of REPL. Let's follow the convention of `mysql` and `mysqld` to name the new client `sqlflow`.
+
+
 
 ### Dependencies and Installation
 
-The original `repl`'s only dependency is the `it2check` bash script that determines whether to call `sixel` to render images.
+The original `repl`'s only dependency is the `it2check` bash script that determines whether to call `sixel` to render images. We'll find a way to show images later. At the moment, we simply remove this dependency.
 
 We'll provide installation scripts on Mac (Bash) and Windows (PowerShell) to install `sqlflow` and its dependencies swiftly. The four scripts are:
 
-1. install.sh (Install Docker, `sqlflow`, and the image `sqlflow/sqlflow:latest` on a Mac)
-1. install-client.sh (Only `sqlflow` on a Mac)
-1. install.ps1(Install Docker, `sqlflow`, and the image `sqlflow/sqlflow:latest` on a PC, coming later)
-1. install-client.ps1 (Only `sqlflow` on a PC, coming later)
+1. For Mac/Linux users:
+    1. `install.sh` (Install `sqlflow` and the image `sqlflow/sqlflow:latest`)
+    1. `install-client.sh` (Only `sqlflow`)
+	1. Provide a brief user guide and a bash command on the SQLFlow homepage like [Homebrew](http://brew.sh). Users can paste the command to a terminal to install SQLFlow. For example: `/bin/bash -c "$(curl -fsSL https://github.com/sql-machien-learning/sqlflow/release/latest/install.sh)"`
+
+1. For PC users (We'll support PCs later):
+    1. install.ps1(Install `sqlflow` and the image `sqlflow/sqlflow:latest`, coming later)
+    1. install-client.ps1 (Only `sqlflow` on a PC, coming later)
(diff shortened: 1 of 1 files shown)


doc/design/cli.md
```diff
@@ -0,0 +1,105 @@
+# SQLFlow Command Line Interface
```
wangkuiyi, line 1: The title and the first section could be simplified somehow.  I tried to do it in the following

----

# SQLFlow Command-line Client

## Motivations

At the moment, users can build SQLFlow into a gRPC server or an [command-line binary](../run/repl.md).  Both forms statically link to the SQLFlow core code and inherit all dependencies, including TensorFlow, various database client libraries, and various SQL dialect parsers in the form of gRPC backend servers.  The many dependencies require us to release and run both forms in Docker containers.

For convenient deployment, we want to run the SQLFlow server in containers, and end-users access the server via the command-line client.  In particular, we want it easy to install the client tool, so we need to prune the dependencies.  As a solution, we make the client tool a statically linked Go binary that remotely calls the SQLFlow server, so users just download and run it.

The command-line client tool, named `sqlflow`, is complementary to other forms of clients, like Jupyter Notebook.  Users can write shell or Python scripts calling the client tool to realize complex applications.

doc/design/cli.md
```diff
@@ -0,0 +1,105 @@
...
+1. We don't need the flag `--model_dir` any more because the training process is running in `sqlflowserver`.
+
+1. We'll define a new flag `--host` that holds the `sqlflowserver` address.
+
+#### Configuration
+We still use `godotenv` as well as environment variables to hold flags that change infrequently. The two necessary config entries are:
+
+1. `SQLFLOW_DATASOURCE` the alternative of `--datasource`. We already have this.
+
+1. `SQLFLOW_SERVER_HOST` the alternative of `--host`. This is to be added.
```
wangkuiyi: SQLFLOW_SERVER_HOST => SQLFLOW_SERVER

doc/design/cli.md
```diff
@@ -0,0 +1,105 @@
...
+1. Wrap the current SQLFlow REPL in a script, we already have such an [attempt](https://github.com/sql-machine-learning/sqlflow/pull/2114).
+1. Decouple the REPL implementation from the core code of SQLFlow and provide a native binary that only functions as a pure client.
+
+We decide to choose the second solution because it's more lightweight, if a user already has a SQLFlow deployment in her company, all she has to do is to download the native binary.
+
+## Design
+
+### Naming
+
+`REPL` is not a good name, We'll rename the binary as simply `sqlflow`.
```
wangkuiyi: We need more clear explanation here:

Currently, the command-line binary form of SQLFlow has the name `repl`, which is the name of a user interface design philosophy.  Many command-line tools, including `mysql` and `python`, implement the UI of REPL.  Let us follow the convention of `mysql` and `mysqld` to name the new client `sqlflow`.

doc/design/cli.md
```diff
@@ -0,0 +1,105 @@
...
+
+### Dependencies and Installation
+
+The original `repl`'s only dependency is the `it2check` bash script that determines whether to call `sixel` to render images.
+
+We'll provide installation scripts on Mac (Bash) and Windows (PowerShell) to install `sqlflow` and its dependencies swiftly. The four scripts are:
```
wangkuiyi: I am afraid that this is too much for users. Let us make sure that users only need to download the client tool before running it. If necessary, let us abandon it2check and other cool stuff at the right moment.

doc/design/cli.md
```diff
@@ -0,0 +1,105 @@
...
+
+1. [Table in protobuf](https://github.com/sql-machine-learning/sqlflow/blob/be7c5728f47e8d3b81893c1d712974a6ddcd5f1c/cmd/repl/repl.go#L161-L163)
+
+1. [The step package](https://github.com/sql-machine-learning/sqlflow/blob/be7c5728f47e8d3b81893c1d712974a6ddcd5f1c/cmd/repl/repl.go#L35)
+
+#### [Command Line Flags](https://github.com/sql-machine-learning/sqlflow/blob/be7c5728f47e8d3b81893c1d712974a6ddcd5f1c/cmd/repl/repl.go#L261-L268)
+
+1. We don't need the flag `--model_dir` any more because the training process is running in `sqlflowserver`.
+
+1. We'll define a new flag `--host` that holds the `sqlflowserver` address.
```
wangkuiyi: Not `--host`, but `--sqlflow_server`.  Let us make sure all names are accurate and precise.

doc/design/cli.md
```diff
@@ -0,0 +1,105 @@
...
+
+1. Use the stub to send the SQL statement get from `ReadStmt` to the server.
+
+1. Parse the result returned from the SQLFlow server.
+    - For non-sixel terminal simulators, the server side should return `Plotille` ASCII figures. See https://github.com/sql-machine-learning/sqlflow/blob/be7c5728f47e8d3b81893c1d712974a6ddcd5f1c/pkg/step/step.go#L69-L85
+
+1. Both the client construction code and result parsing code can be implemented by referring to the existing code in `sqlflowserver/main_test.go`.
+
+#### Auto-complete
+The original implementation of auto-complete is based on the `pkg/sql/attribute` package, in a pure client as `sqlflow`, we can only depend on such packages partly because the client may have a different version with the server-side.
```
wangkuiyi, line 73: Let us remove this auto-completion feature at the first stage. We need a minimum viable product at the moment. Let us define it clearly in this design doc.

doc/design/cli.md
```diff
@@ -0,0 +1,105 @@
...
+#### [RunSQLProgramAndPrintResult](https://github.com/sql-machine-learning/sqlflow/blob/be7c5728f47e8d3b81893c1d712974a6ddcd5f1c/cmd/repl/repl.go#L175)
+ 
+We have to replace the call to `step.RunSQLProgramAndPrintResult` with a new function `RunSQLProgram` that would be defined in the `sqlflow` codebase. The new function should:
+
+1. Construct a client stub from `--host`. 
+
+1. Use the stub to send the SQL statement get from `ReadStmt` to the server.
+
+1. Parse the result returned from the SQLFlow server.
+    - For non-sixel terminal simulators, the server side should return `Plotille` ASCII figures. See https://github.com/sql-machine-learning/sqlflow/blob/be7c5728f47e8d3b81893c1d712974a6ddcd5f1c/pkg/step/step.go#L69-L85
```
wangkuiyi: Plotille cannot draw images with high-precision.  Let us save the image to local files and optionally start a viewer program.  We can start Preview.app on macOS, or the `display` command-line tool from ImageMagick on Linux with X-window.

wangkuiyi (changes requested):

doc/design/cli.md
shendiaomo: Done.

doc/design/cli.md
shendiaomo: Done.

doc/design/cli.md
shendiaomo: Done.

doc/design/cli.md
shendiaomo: Done.

doc/design/cli.md
shendiaomo: The client won't rely on Plotille, it's a dependency of SQLFlow server. We only need to get the Plotille file from the server.

Anyway, we'll rethink how to show figures before integrating any related new features.

doc/design/cli.md
shendiaomo: Done. 
The client will only provide auto-complete features for built-in models (canned estimators and XGBoost) at the 1st stage. 
The auto-complete features for `sqlflow_models` and Model Zoo models will be designed at the next state when necessary.

`it2check` has nothing to do with auto-completion, we'll avoid this dependency at the first stage

doc/design/cli.md
shendiaomo: Done.

typhoonzero (changes requested):
Can we just reuse the repo `pysqlflow` so that:

1. users can use `pip install sqlflow` to install the command line together with necessary dependencies.
1. `pysqlflow` is already a gRPC client that support sending SQL statements to the SQLFlow server, just add an interactive command line to it.

shendiaomo:
> Can we just reuse the repo `pysqlflow` so that:
> 
> 1. users can use `pip install sqlflow` to install the command line together with necessary dependencies.
> 2. `pysqlflow` is already a gRPC client that support sending SQL statements to the SQLFlow server, just add an interactive command line to it.

At the moment, releasing `sqlflow` basing on `repl` is much more practical than basing on `pysqlflow` in three aspects:
1. The binary doesn't depend on any other software including python. 
2. Adding a gRPC client to `repl` is much less of work than adding a qualified interactive interface to `pysqlflow`.
3. We should provide our best to the users.

wangkuiyi (approved):
I agree that the client tool is in Go other than in Python.  A Go program can be statically-linked and has no dependencies at all. This makes the installation easier and to support more platforms.

wangkuiyi (approved):

typhoonzero (approved):

PR #2197, opened 2020-05-07
Expose ports

wangkuiyi:
By listing the ports we want to expose in the Dockerfile, we can use `-P` with `docker run` to expose all the listed ports.  In this case, `docker port` lists all the host ports mapping to the exposed ports in the container.


lhw362950217 (approved):

doc/run/cli.md
```diff
@@ -15,7 +15,7 @@ In the REPL mode, `sqlflow` supports automatic code completion and other feature
 We can run an example session by typing the following command on macOS.  We assume you have installed [Docker](https://docs.docker.com/get-docker/) on your computer.
 
 ```bash
-docker run -d --rm -P -p 50051 --name sqlflowserver \
+docker run -d --rm -P --name sqlflowserver \
```
typhoonzero, line 18: If we are using `-P`, need to add descriptions below of how to get the exposed ports on the host, so that the cli can access those ports.

wangkuiyi: I see that @lhw362950217 had an example in https://github.com/sql-machine-learning/sqlflow/pull/2197/files#diff-ed67adab77578fc4974eec456e2e126eR21 that calls the `docker port` command. I will try to emphasize the command in a successive PR.

Issue #2196, opened 2020-05-07
docker build command often rebuild the image


wangkuiyi:
Currently, we have only one Dockerfile, which defines the development environment and the released image.

This approach has a problem -- while I am developing in the local Git repo, I often change some files.  When I run `docker build`, the changed file would trigger the rebuilding process, which takes a long time.

Also, the generated Docker image is huge.

It looks like a good idea to separate the development environment image and the released image.

So, `docker build -t sqlflow:dev .` should not depend on the SQLFlow source code.  And `docker run --rm $PWD:/work sqlflow:dev` should generate a directory of built binaries and a Dockerfile, like

```
build/
 |- *.jar - Java parsers
 |- sqlflowserver - Go binary
 |- sqlflow - Go binary
 |- step - Go binary
 |- python/sqlflow - Python code
 |- notebook - Jupyter notebooks converted from Markdown files
```

Then, by running `cd build && docker build -t sqlflow .`, we get the released image.

wangkuiyi:
A related question is -- should we move `/scripts/tensorflow_step_image/Dockerfile` to `/docker/step/Dockerfile` so we have the following directory structure:

```
/docker/
 |- dev/Dockefile
 |- step/Dockerfile
 |- ci/Dockerfile
```

And we use the image `sqlflow/sqlfow:ci` desktop playground setup?

typhoonzero:
I did some same tests earlier: https://github.com/sql-machine-learning/sqlflow/tree/develop/scripts/tensorflow_step_image, yet it seems that even if we have a runtime Docker image, it is still huge, like 2.1G ( the image `sqlflow/sqlflow` size is 5.44G currently )

typhoonzero:
We designed the Dockerfile to be two parts: 1. prepare the dependencies and environment, 2. build and install SQLFlow. Part 2 should change often and we expect part 1 rarely changes.

https://github.com/sql-machine-learning/sqlflow/blob/4404ff7e0dfdf9e576f8d58e4b629bd5e607f081/Dockerfile#L75-L79

wangkuiyi:
@typhoonzero I know all the facts. The problem is that when I am developing on my desktop, even if I change a document file, `docker build` would run a very long compiling process.

One of the reason is that `mvn build` needs to download dependent Java packages to `$HOME/.cache` in the container.  If we explicitly distinguish the dev image and the release image, I hope that by running the building process explicitly via `docker run -v $PWD:/work 
[...]

PR #2198, opened 2020-05-07
Simplify install-*.bash

wangkuiyi:
This is a preparation of https://github.com/sql-machine-learning/sqlflow/issues/2196.  In order to distinguish install-*.bash files into development tools and runtime dependencies.

If this PR merges, I will move /scripts/docker/install-*.bash into /docker/dev/


wangkuiyi:
Please don't worry about the loss of the way we add it2check.  We can search over the Git history using the shell script https://stackoverflow.com/a/4468394/724872.

lhw362950217 (approved):
LGTM

PR #2203, opened 2020-05-09
Split the devbox image from /Dockerfile into /docker/dev/Dockerfile

### LINKED ISSUE #2196: docker build command often rebuild the image

wangkuiyi:
Fix https://github.com/sql-machine-learning/sqlflow/issues/2196


Dockerfile
```diff
@@ -31,93 +10,31 @@ COPY doc/datasets/popularize_churn.sql \
...
+ENV PYTHONPATH=/usr/local/sqlflow/python:$PYTHONPATH
 
-# The SQLFlow magic command for Jupyter.
-ENV IPYTHON_STARTUP /root/.ipython/profile_default/startup/
-COPY scripts/docker/install-jupyter.bash /
-COPY scripts/docker/js /js
-RUN /install-jupyter.bash
+# Install Couler, Fluid, and model zoo.
+COPY build/*.whl /usr/local/sqlflow/python/
+RUN pip install /usr/local/sqlflow/python/*.whl
```
typhoonzero, line 20: `RUN pip install /usr/local/sqlflow/python/*.whl && rm -f /usr/local/sqlflow/python/*.whl` to make the image smaller

scripts/travis/build.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
```
typhoonzero, line 1: build.sh -> build_images.sh?

scripts/travis/push.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
```
typhoonzero, line 1: push.sh -> push_images.sh?

docker/dev/Dockerfile
```diff
@@ -0,0 +1,72 @@
...
+# The SQLFlow magic command for Jupyter.
+ENV IPYTHON_STARTUP /root/.ipython/profile_default/startup/
+COPY install-jupyter.bash /
+COPY js /js
+RUN /install-jupyter.bash
+
+# Set build.sh as the entrypoint and assume that SQLFlow source tree
+# is at /work.
+COPY build.sh /
+CMD ["/build.sh", "/work"]
```
typhoonzero, line 72: The dev box is just some environment that needed by SQLFlow building and developing, so maybe remove here, and do `docker run --rm -v /path/to/sqlflow:/go/src/sqlflow.org/sqlflow sqlflow/sqlflow:dev /go/src/sqlflow.org/sqlflow/scripts/build.sh` ?

Yancey0623 (approved):
LGTM

wangkuiyi:
According to the records on DockerHub.com

Screen Shot 2020-05-10 at 7 22 17 PM

before this change, we have the all-in-one Docker image `sqlflow/sqlflow:latest` of size 2.5GB; after this change, we have two images, `sqlflow/sqlflow:dev` takes 1.89 GB, and the new all-in-one image `sqlflow/sqlflow:ci` is 2.11 GB and is smaller than the old one.

PR #2217, opened 2020-05-11
Support to exit sqlflow cli by "exit" and "quit" command

sneaxiy:
Currently, sqlflow cli only supports `Ctrl + D` to exit. However, like most of the softwares, mysql supports `exit` and `quit` command to exit cli. This PR adds `exit` and `quit` command support in sqlflow cli, which may be friendly to users.

### DIFF
--- a/cmd/sqlflow/main.go
+++ b/cmd/sqlflow/main.go
@@ -180,6 +180,17 @@ func sqlRequest(program string, ds string) *pb.Request {
 	return &pb.Request{Sql: program, Session: se}
 }
 
+func isExitStmt(stmt string) bool {
+	separatorIndex := strings.Index(stmt, ";")
+	if separatorIndex < 0 {
+		separatorIndex = len(stmt)
+	}
+
+	firstStmt := stmt[0:separatorIndex]
+	firstStmt = strings.ToUpper(strings.TrimSpace(firstStmt))
+	return firstStmt == "EXIT" || firstStmt == "QUIT"
+}
+
 func runStmt(serverAddr string, stmt string, isTerminal bool, ds string) error {
 	// special case, process USE to stick SQL session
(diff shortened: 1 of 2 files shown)


wangkuiyi (approved):

typhoonzero (approved):

cmd/sqlflow/main.go
```diff
@@ -356,7 +367,12 @@ func main() {
 		if !*noAutoCompletion {
 			// TODO(lorylin): get autocomplete dicts for sqlflow_models from sqlflow_server
 		}
-		runPrompt(func(stmt string) { runStmt(*serverAddr, stmt, true, *ds) })
+		runPrompt(func(stmt string) {
```
lhw362950217, line 370: maybe we should move this logic to [there](https://github.com/sql-machine-learning/sqlflow/blob/2ad28c15dd7b773f19aa2991a8de907bde649ca4/cmd/sqlflow/main.go#L186) and keep the code clean here.

btw, we may say 'Goodby!' as we do in response to `ctrl+D` action

sneaxiy: Thanks for suggestions! I have once moved this logic to the position you mentioned. However, I keep `os.Exit(0)` to the position in this PR after thinking. My concern is that `runStmt` should be a function to execute SQL statements. So it should report execution results or errors after this function is invoked, instead of considering the process exit cases. Therefore, I think that calling `os.Exit(0)` inside `runStmt` may not be a good idea.

sneaxiy: To say `Goodbye!` before `os.Exit(0)` is a good idea! I would add it asap in the next PR.

lhw362950217: There is another special case `use db`, like `exit`, we can process these client side commands together as mentioned above.

PR #2222, opened 2020-05-11
Move os.Exit(0) in runPrompt to runStmt

sneaxiy:
Related to https://github.com/sql-machine-learning/sqlflow/pull/2217 . As discussed in https://github.com/sql-machine-learning/sqlflow/pull/2217#discussion_r422920897 , it is better to move all corner cases handling inside `runStmt` . This PR moves `os.Exit(0)` (when `exit` or `quit` command is received) to `runStmt`.

Also, add `fmt.Println("Goodbye!)"` when exiting as what `Ctrl + D` does.


lhw362950217 (approved):
LGTM

PR #2229, opened 2020-05-12
Merge the first CI stage PreCommit into the second

wangkuiyi:
The first CI stage PreCommit has only one job, which runs pre-commit to check the source code.  It seems that we can merge it into the second stage to maximize the concurrent execution of CI jobs.


tonyyang-svail (approved):
LGTM

PR #2232, opened 2020-05-13
Release model definitions and list them

typhoonzero:
Continue work with model zoo server.

### DIFF
--- a/pkg/modelzooserver/modelzooserver.go
+++ b/pkg/modelzooserver/modelzooserver.go
@@ -28,10 +28,10 @@ const modelDefTable = "sqlflow_model_zoo.model_definition"
 const trainedModelTable = "sqlflow_model_zoo.trained_model"
 
 // TODO(typhoonzero): create tables if these tables are not pre created?
-const createTableStmts = `CREATE DATABASE sqlflow_model_zoo;
-DROP TABLE IF EXISTS sqlflow_model_zoo.model_collection;
-DROP TABLE IF EXISTS sqlflow_model_zoo.model_definition;
+const createTableStmts = `CREATE DATABASE IF NOT EXISTS sqlflow_model_zoo;
 DROP TABLE IF EXISTS sqlflow_model_zoo.trained_model;
+DROP TABLE IF EXISTS sqlflow_model_zoo.model_definition;
+DROP TABLE IF EXISTS sqlflow_model_zoo.model_collection;
 
 CREATE TABLE sqlflow_model_zoo.model_collection (
     id INT AUTO_INCREMENT,
--- a/pkg/modelzooserver/modelzooserver.go
+++ b/pkg/modelzooserver/modelzooserver.go
@@ -27,8 +27,13 @@ const modelCollTable = "sqlflow_model_zoo.model_collection"
 const modelDefTable = "sqlflow_model_zoo.model_definition"
 const trainedModelTable = "sqlflow_model_zoo.trained_model"
 
-// TODO(typhoonzero): create tables if these tables are not pre created.
-const createTableStmts = `CREATE TABLE sqlflow_model_zoo.model_collection (
+// TODO(typhoonzero): create tables if these tables are not pre created?
+const createTableStmts = `CREATE DATABASE sqlflow_model_zoo;
+DROP TABLE IF EXISTS sqlflow_model_zoo.model_collection;
+DROP TABLE IF EXISTS sqlflow_model_zoo.model_definition;
+DROP TABLE IF EXISTS sqlflow_model_zoo.trained_model;
+
+CREATE TABLE sqlflow_model_zoo.model_collection (
     id INT AUTO_INCREMENT,
     name VARCHAR(255),
     version VARCHAR(255),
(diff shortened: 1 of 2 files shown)


pkg/modelzooserver/modelzooserver.go
```diff
@@ -15,28 +15,162 @@ package modelzooserver
...
 	"context"
+	"fmt"
+	"io"
+	"log"
 
+	"sqlflow.org/sqlflow/pkg/database"
 	pb "sqlflow.org/sqlflow/pkg/proto"
 )
 
+const modelCollTable = "sqlflow_model_zoo.model_collection"
```
wangkuiyi, line 26: How about col, and collection => repo?

pkg/modelzooserver/modelzooserver.go
```diff
@@ -15,28 +15,162 @@ package modelzooserver
...
+    model_def_id INT,
+    name VARCHAR(255),
+    version VARCHAR(255),
+    url VARCHAR(255),
+    description TEXT,
+    metrics TEXT,
+    PRIMARY KEY (id),
+    FOREIGN KEY (model_def_id) REFERENCES model_definition(id)
+);`
+
```
wangkuiyi, line 63: Should we reuse https://github.com/sql-machine-learning/sqlflow/blob/develop/pkg/database/testdata/sql.go by moving the file sql.go to pkg/database and expose functions defined in it?

pkg/modelzooserver/modelzooserver.go
```diff
@@ -15,28 +15,162 @@ package modelzooserver
...
 type modelZooServer struct {
+	DB *database.DB
 }
 
 func (s *modelZooServer) ListModelDefs(ctx context.Context, req *pb.ListModelRequest) (*pb.ListModelResponse, error) {
```
wangkuiyi, line 68: I took a look into the proto file.   It seems that the size field in ListModelReponse is redundant to `len(names)`.

```protobuf
message ListModelResponse {
    // A list of model definition or trained model names.
    repeated string names = 1;
    // A list of tags of the above list names, the length should be the same as "names".
    repeated string tags = 2;
    // Size of elements in names.
    int64 size = 3;
}
```

wangkuiyi: How if users want to list model definitions belonging to a specific model collection/repo?  I don't see a field in `ListModelRequest` allowing users to specify the filtering condition.

I would suggest that we follow the convention of GitHub API to have

```protobuf
message /*model_zoo.*/ListRepoRequest {...}
message /*model_zoo.*/ListRepoResponse {...}

message /*model_zoo.*/ListDefinitionRequest { string repo; ...}
message /*model_zoo.*/ListDefinitionResponse {...}

message /*model_zoo.*/ListModelRequest { string repo; string definition; ...}
message /*model_zoo.*/ListModelResponse {...}
```

typhoonzero: Planning to add this in the next PR.

pkg/modelzooserver/modelzooserver.go
```diff
@@ -15,28 +15,162 @@ package modelzooserver
...
+// TODO(typhoonzero): create tables if these tables are not pre created?
+const createTableStmts = `CREATE DATABASE IF NOT EXISTS sqlflow_model_zoo;
+DROP TABLE IF EXISTS sqlflow_model_zoo.trained_model;
+DROP TABLE IF EXISTS sqlflow_model_zoo.model_definition;
+DROP TABLE IF EXISTS sqlflow_model_zoo.model_collection;
+
+CREATE TABLE sqlflow_model_zoo.model_collection (
+    id INT AUTO_INCREMENT,
+    name VARCHAR(255),
+    version VARCHAR(255),
```
lhw362950217, line 39: Maybe we need some field to describe the zipped file's storage location?

pkg/modelzooserver/modelzooserver.go
typhoonzero: Maybe the name "repo" would let model contributors be confused about the model source code stored in git repo?

pkg/modelzooserver/modelzooserver.go
```diff
@@ -15,28 +15,162 @@ package modelzooserver
...
+);
+
+CREATE TABLE sqlflow_model_zoo.trained_model (
+    id INT AUTO_INCREMENT,
+    model_def_id INT,
+    name VARCHAR(255),
+    version VARCHAR(255),
+    url VARCHAR(255),
+    description TEXT,
+    metrics TEXT,
```
lhw362950217, line 59: TEXT's max length is 64K, I'm not sure if it is enough for us.

pkg/modelzooserver/modelzooserver.go
typhoonzero: Yes, we indeed need some Go functions to do database operations, and these functions should be able to avoid "SQL injection attacks". I added a "TODO" comment in `modelzooserver.go` to implement these functions later.

pkg/modelzooserver/modelzooserver.go
typhoonzero: For creating model collection Docker images, we don't need the zip file's location because we will pack it inside the Docker image.

pkg/modelzooserver/modelzooserver.go
typhoonzero: Should be enough, metrics field stores a JSON string of the model validation metrics like: 
```json
{"accuracy": 0.9, "AUC": 0.78}
```

lhw362950217 (approved):
LGTM

typhoonzero:
Will polish the result returning in the next PR.

PR #2233, opened 2020-05-13
clean up legacy model zoo code

typhoonzero:
Remove not used code.

The code in `model.go` was used to implement a legacy design of the model zoo, according to our new design: https://github.com/sql-machine-learning/sqlflow/blob/develop/doc/design/model_zoo.md and https://github.com/sql-machine-learning/sqlflow/blob/develop/doc/design/contribute_models_new.md, we should remove the legacy code here.

### DIFF
--- a/pkg/model/model.go
+++ b/pkg/model/model.go
@@ -69,14 +69,6 @@ func (m *Model) Save(modelURI string, trainStmt *ir.TrainStmt, session *pb.Sessi
 	if err := m.saveDB(db, modelURI, session); err != nil {
 		return err
 	}
-	// TODO(typhoonzero): support hive, maxcompute saving model zoo metas.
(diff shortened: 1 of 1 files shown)


pkg/model/model.go
```diff
@@ -214,86 +206,3 @@ func readGob(filePath string, object interface{}) error {
 	}
 	return nil
 }
-
```
wangkuiyi, line 217: How about comment this PR to explain how came these functions and why we could remove them?

wangkuiyi (commented):
LGTM

pkg/model/model.go
typhoonzero: Done

wangkuiyi (approved):

PR #2240, opened 2020-05-14
XGBoost feature column support design doc

sneaxiy:
This PR adds the feature column support in XGBoost models, so that the XGBoost models can accept string column as it inputs or do some pre-processing before training.

### DIFF
--- a/doc/design/figures/xgboost_feature_column.md
+++ b/doc/design/figures/xgboost_feature_column.md
@@ -17,7 +17,7 @@ Currently, `COLUMN` clauses are supported in SQLFlow TensorFlow models. The `COL
 into TensorFlow feature column API calls inside SQLFlow codegen implementation.
 
 However, XGBoost has no similar feature column APIs as TensorFlow. Currently, XGBoost models can only support simple column names like `c1, c2, c3` in `COLUMN` clauses,
-and any data pre-processing is not supported. It makes that we cannot use XGBoost to train NLP models which accept string column as their input.
+and any data pre-processing is not supported. It makes that we cannot use XGBoost to train models which accept string column as their input.
 
 This design explains how SQLFlow supports feature columns in XGBoost model.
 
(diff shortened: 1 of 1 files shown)


doc/design/figures/xgboost_feature_column.md
```diff
@@ -0,0 +1,131 @@
...
+COLUMN INDICATOR(CATEGORY_HASH(string_column1, 10)), int_column2
+LABEL class 
+INTO sqlflow_xgboost_model.my_model;
+```
+
+Currently, `COLUMN` clauses are supported in SQLFlow TensorFlow models. The `COLUMN` clauses are transformed
+into TensorFlow feature column API calls inside SQLFlow codegen implementation.
+
+However, XGBoost has no similar feature column APIs as TensorFlow. Currently, XGBoost models can only support simple column names like `c1, c2, c3` in `COLUMN` clauses,
+and any data pre-processing is not supported. It makes that we cannot use XGBoost to train NLP models which accept string column as their input.
```
typhoonzero: Not only NLP models will be supported but also models that takes categorical string feature inputs.

doc/design/figures/xgboost_feature_column.md
```diff
@@ -0,0 +1,131 @@
...
+            
+        for idx, b in enumarate(boudaries):
+            if input >= b
+                return idx
+        return len(boundaries)
+
+# CrossedColumnTransformer is designed as a decorator of CategoryColumnTransformer
+# It accepts several CategoryColumnTransformers, and crosses their transformed
+# results using hash algorithm.
+class CrossedColumnTransformer(CategoryColumnTransformer):
```
typhoonzero: Cross columns should not be a subclass of `CategoryColumnTransformer` since it may not be a categorical column.

doc/design/figures/xgboost_feature_column.md
```diff
@@ -0,0 +1,131 @@
...
+        pass
+
+    def __call__(self, inputs):
+        transformed_inputs = []
+        for transformer in self.category_transformers:
+            transformed_inputs.append(transformer(inputs))
+         
+        self._cross(transformed_inputs)
+        
+class ListedFeatureColumnTransformer(BaseFeatureColumnTransformer):
```
typhoonzero, line 97: Maybe `ListedFeatureColumnTransformer` is not needed, we can call the transformers in a function call manner like: `IndicatorColumnTransformer(CategoryColumnWithHashBucketTransformer(...))`

doc/design/figures/xgboost_feature_column.md
sneaxiy: Thanks. Done.

doc/design/figures/xgboost_feature_column.md
sneaxiy: Thanks. I have misunderstond `CrossColumn`. Fixed.

doc/design/figures/xgboost_feature_column.md
sneaxiy: In this design, any transformer would only handle one `COLUMN` clause, and each clause would produce a list of transformer calls. `ListedFeatureColumnTransformer` is used to combine all `COLUMN` clauses. Therefore, maybe `ListedFeatureColumnTransformer` is needed to combine each column output.

typhoonzero (approved):
LGTM++

PR #2242, opened 2020-05-14
Use Qiniu than AWS for releasing

wangkuiyi:
It takes 30~60 minutes for users in Hangzhou to download the sqlflow client tool from the Amazon S3 bucket in Hong Kong.  After comparing download speed, we decided to use Qiniu.

### DIFF
--- a/scripts/travis/deploy_client.sh
+++ b/scripts/travis/deploy_client.sh
@@ -54,30 +54,18 @@ cd $TRAVIS_BUILD_DIR
 go generate ./...
 GOBIN=/tmp go install ./cmd/sqlflow
 
-echo "Install AWS client for $TRAVIS_OS_NAME ..."
-if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
-    curl -s "https://awscli.amazonaws.com/AWSCLIV2.pkg" -o "AWSCLIV2.pkg"
-    sudo installer -pkg AWSCLIV2.pkg -target /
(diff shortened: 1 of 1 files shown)


scripts/travis/deploy_client.sh
```diff
@@ -54,30 +54,18 @@ cd $TRAVIS_BUILD_DIR
...
-    s3 cp /tmp/sqlflow \
-    s3://sqlflow-release/$RELEASE_TAG/$TRAVIS_OS_NAME/sqlflow
-aws --region ap-east-1 --output text \
-    s3api put-object-acl \
-    --bucket sqlflow-release \
-    --key $RELEASE_TAG/$TRAVIS_OS_NAME/sqlflow \
-    --acl public-read
+echo "Publish /tmp/sqlflow to Qiniu Object Storage ..."
+qshell account "$QINIU_AK" "$QINIU_SK" "wu"
+qshell rput sqlflow-release \
```
shendiaomo, line 69: Do we need to put the `sqlflow` binary to different directories to support various OS?
For example:
```
windows/sqlflow
OSX/sqlflow
Linux/sqlflow
```

lhw362950217 (approved):
LGTM

scripts/travis/deploy_client.sh
wangkuiyi: Yes. That is what we are doing here. `sqlflow-release` is the bucket name. The next line is the file path.

PR #2252, opened 2020-05-14
Use axel than curl

### LINKED ISSUE #2249: Download ODPS client breaks often in CI
Need to mirror the tool on AWS and Qiniu and use `axel` instead of `curl` to download.

### LINKED ISSUE #2250: Change download tool from curl to axel
For two reasons, we need to be able to download a file from more than one sources:

1. Packages like the Go compiler has mirrors all over the world, and SQLFlow has developers all over the world.  We want to be able to automatically choose the faster mirror.  The benefit is shown in the following figure:

   image

1. We might need to create mirrors of some dependencies whose publishers didn't make mirrors.  For example, Travis CI often breaks due to the unstable Aliyun download link of their ODPS tool.

wangkuiyi:
Fix https://github.com/sql-machine-learning/sqlflow/issues/2250
Fix https://github.com/sql-machine-learning/sqlflow/issues/2249

### DIFF
--- a/docker/dev/install.sh
+++ b/docker/dev/install.sh
@@ -68,13 +68,17 @@ rm -rf $HOME/.cache/pip/*
 
 
 echo "Install Go compiler ..."
-GO_DEV="https://dl.google.com/go/go1.13.4.linux-amd64.tar.gz"
-curl -sL $GO_DEV  | tar -C /usr/local -xzf -
+GO_MIRROR_0="https://dl.google.com/go/go1.13.4.linux-amd64.tar.gz"
+GO_MIRROR_1="http://mirrors.ustc.edu.cn/golang/go1.13.4.linux-amd64.tar.gz"
+axel --quiet --output go.tar.gz $GO_MIRROR_0 $GO_MIRROR_1
+tar -C /usr/local -xzf go.tar.gz
+rm go.tar.gz
 export GOPATH="/root/go"
 export PATH="/usr/local/go/bin:$GOPATH/bin:$PATH"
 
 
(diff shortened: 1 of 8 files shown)


docker/dev/install.sh
```diff
@@ -122,12 +126,9 @@ echo '<settings>
...
-     -o /usr/local/bin/google_checks.xml
-curl -sLJ \
-     "https://github.com/checkstyle/checkstyle/releases/download/checkstyle-8.29/checkstyle-8.29-all.jar" \
-     -o /usr/local/bin/checkstyle-8.29-all.jar
+axel --quiet --output /usr/local/bin/google-java-format-1.6-all-deps.jar \
+     "https://github.com/google/google-java-format/releases/download/google-java-format-1.6/google-java-format-1.6-all-deps.jar"
+axel --quiet --output /usr/local/bin/google_checks.xml \
+     "https://raw.githubusercontent.com/checkstyle/checkstyle/master/src/main/resources/google_checks.xml"
+axel --quiet --output /usr/local/bin/checkstyle-8.29-all.jar \
+     "https://github.com/checkstyle/checkstyle/releases/download/checkstyle-8.29/checkstyle-8.29-all.jar"
```
tonyyang-svail, line 134: Should we use `curl` when downloading with one URL?

tonyyang-svail (approved):

docker/dev/install.sh
wangkuiyi: I have an ambitious plan to mirror these links on AWS S3 to accelerate the CI, because Travis CI VMs run on AWS.

PR #2269, opened 2020-05-16
Release and query trained models in model zoo

typhoonzero:


pkg/modelzooserver/modelzooserver.go
```diff
@@ -109,7 +109,60 @@ func (s *modelZooServer) ListModelDefs(ctx context.Context, req *pb.ListModelReq
...
+		modelVersions = append(modelVersions, v)
+		urls = append(urls, url)
+		descs = append(descs, desc)
+		metrics = append(metrics, m)
+		imageURL := fmt.Sprintf("%s:%s", imagename, imagetag)
+		imageurls = append(imageurls, imageURL)
+	}
+
+	return &pb.ListTrainedModelResponse{
+		Names:          names,
```
lhw362950217: can we return list of wrapped object instead of lists of fields?

typhoonzero: Good idea!

pkg/proto/modelzooserver.proto
```diff
@@ -58,6 +62,7 @@ message ListModelDefResponse {
     int64 size = 5;
 }
```
weiguoz, line 68: Shall we add a `ModelDefResponse` to The `ListModelDefResponse`? Like:
``` protobuf
message ListModelDefResponse {
    message ModelDefRsp {
        string class_name = 1;
        string image_url = 2;
        string tag = 3;
        string arg_desc = 4;
    }
    repeated ModelDefRsp model_defs = 1;
}
```
So that we don't have to tell users neither `len(class_names) = len(image_urls) = ... = size` nor the sequence is important.

update:
I just realized [the comment](https://github.com/sql-machine-learning/sqlflow/pull/2269/files#r426343888).  Is that what you want to do? @typhoonzero @lhw362950217

typhoonzero: Yes, that's indeed better.

lhw362950217 (approved):
LGTM

PR #2292, opened 2020-05-18
Support upload model repo files and release the model

typhoonzero:
Part of https://github.com/sql-machine-learning/sqlflow/issues/2307

TODO: will update `input_check.go` in next PR.

### DIFF
--- a/pkg/modelzooserver/tar.go
+++ b/pkg/modelzooserver/tar.go
@@ -49,7 +49,7 @@ func untarGzDir(tarball string, output string) error {
 	if err != nil {
 		return err
 	}
-	return uncompress(rd, ".")
+	return uncompress(rd, output)
 }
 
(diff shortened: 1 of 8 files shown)


pkg/modelzooserver/tar.go
```diff
@@ -0,0 +1,150 @@
...
+import (
+	"archive/tar"
+	"bytes"
+	"compress/gzip"
+	"io"
+	"os"
+	"path/filepath"
+)
+
+func tarGzDir(dir string, output string) error {
```
lhw362950217, line 25: Can we put these tool func to a distinct package so that the sqlflow cli can reuse them ?

typhoonzero: We can move them out when needed.

lhw362950217 (approved):
LGTM

PR #2298, opened 2020-05-19
Split setup_minikube.sh into multiple scripts reusable by desktop

wangkuiyi:
This PR splits `scripts/travis/setup_minikube.sh` into multiple scripts named `install_*.sh` and `start_*.sh`.  So, the sqflow.org/desktop project can call `install_*.sh` from Vagrant provision shell script and call `start_*.sh` from inside the running VM.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,56 +34,56 @@ install:
 - echo "skip install"
 jobs:
   include:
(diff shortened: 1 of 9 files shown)


Yancey0623 (approved):
LGTM

PR #2302, opened 2020-05-19
Fix BUCKET cannot accept column name and negative boundaries bug

sneaxiy:
Add `BUCKET` column support of:

- `BUCKET(col_name)`, where `col_name` is the string name of the column. Only `BUCKET(NUMERIC(col_name))` is supported before this PR.
- `BUCKET(col_name, [-10])`, where the boundaries contain negative number.

### DIFF
--- a/pkg/sql/ir_generator.go
+++ b/pkg/sql/ir_generator.go
@@ -42,6 +42,7 @@ const (
 	square        = "SQUARE"
 	dense         = "DENSE"
 	comma         = "COMMA"
+	negative      = "-"
 )
 
 func generateTrainStmtWithInferredColumns(slct *parser.SQLFlowSelectStmt, connStr string, verifyLabel bool) (*ir.TrainStmt, error) {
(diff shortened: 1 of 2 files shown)


pkg/sql/ir_generator.go
```diff
@@ -538,27 +564,43 @@ func parseNumericColumn(el *parser.ExprList) (*ir.NumericColumn, error) {
...
-	if err != nil {
-		return nil, err
-	}
-	if _, ok := source.(*ir.NumericColumn); !ok {
-		return nil, fmt.Errorf("key of BUCKET must be NUMERIC, which is %s", source)
+
+	var source ir.FeatureColumn
+	var err error
+
+	if sourceExprList.Sexp == nil {
```
typhoonzero, line 578: Better to use `sourceExprList.Type` to determine whether the subexpression is a string or an expression.

pkg/sql/ir_generator.go
```diff
@@ -440,13 +441,23 @@ func parseExpression(e interface{}) (interface{}, error) {
...
 					}
-				} else {
-					value, err := parseExpression(&expr.Sexp)
-					if err != nil {
-						return nil, err
+					continue
+				}
+
+				// parse negative integer
+				if len(expr.Sexp) == 2 && (*expr.Sexp[0]).Value == negative {
```
typhoonzero, line 448: Can add a comment to reference extended_syntax_parser.y:368 for further reading.

typhoonzero (approved):
Just some minor comments.

pkg/sql/ir_generator.go
sneaxiy: OK. I will fix it in next PR asap.

pkg/sql/ir_generator.go
sneaxiy: Same above.

pkg/sql/ir_generator.go
sneaxiy: OK. I will fix it in next PR asap.

PR #2310, opened 2020-05-19
Call scripts/travis/upload_coveralls.sh in Travis CI VM

wangkuiyi:
Currently, the `.travis.yml` run `scripts/travis/upload_coveralls.sh` in a container.  Indeed, the script starts a container and call `goveralls` to upload the `coverage.out` file to coveralls.io.  This leads to `goveralls` running in a container in a container in the Travis VM.  This change removes a nested layer of container.

Also, the container that `upload_coveralls.sh` starts executes the `sqlflow:ci` image, which derives from `sqlflow:dev`.  Actually, `upload_coveralls.sh` can run `sqlflow:dev` other than `sqlflow:ci` because the former contains `goveralls` and is smaller than the latter.


tonyyang-svail (approved):
LGTM

PR #2312, opened 2020-05-19
Polish Travis CI

wangkuiyi:
- Use Ubuntu Bionic other than Xeniel in Travis CI.  This is not necessary to keep CI running, but reasonable because our dev and ci Docker images all base on Bionic.  Another potential benefit is that we try to keep using recent versions of minikube and kubectl in Travic CI VMs.  It is good to keep up with recent Linux versions in the VM.

- Remove `--cpu` and `--memory` when starting minikube with `--driver=none`, because minikube prompts that `--cpu` and `--memory` options do not matter when `--driver=none`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: go
-
+dist: bionic
 
 # SQLFlow uses sqlflow.org/sqlflow as vanity import path. TravisCI
 # supports it via go_import_path.  Ref:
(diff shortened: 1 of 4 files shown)


tonyyang-svail (commented):
@wangkuiyi Could you please mention the reason for switching from Xenial to Bionic?

wangkuiyi:
> @wangkuiyi Could you please mention the reason for switching from Xenial to Bionic?

@tonyyang-svail Sure. I updated the description of this PR.

tonyyang-svail (approved):
LGTM.

PR #2325, opened 2020-05-21
Polish language guide document

### LINKED ISSUE #2289: SELECT statement should be the same in TRAIN/PREDICT/EXPLAIN
- The `SELECT` clauses should be the same in TRAIN/PREDICT/EXPLAIN. For example, when there is `SELECT a * 2, b` in training, it should be also `SELECT a * 2, b` in predicting and explaining. We should add some docs to explain it.

- The column names of tables in TRAIN/PREDICT/EXPLAIN should be exactly the same. For example, the column names in train table is `a, b` and the column name in test table must be `a, b` too. Should we only need the equality of column number instead of column names? Moreover, can we use alias column names in PREDICT/EXPLAIN?

### LINKED ISSUE #2290: EVALUATE clause support
Do we support `TO EVALUATE` ? I see it in codes and unittests, but there is no doc on it.

https://github.com/sql-machine-learning/sqlflow/blob/94153c2b5c1133de7184b36df0663ad545d4ad09/pkg/parser/extended_syntax_parser.y#L257

https://github.com/sql-machine-learning/sqlflow/blob/32b5dc5cb5d34a4140aec7d4f4b329a3d55be27b/cmd/sqlflowserver/main_test.go#L1504

typhoonzero:
Fix https://github.com/sql-machine-learning/sqlflow/issues/2290
Fix https://github.com/sql-machine-learning/sqlflow/issues/2289

### DIFF
--- a/doc/language_guide.md
+++ b/doc/language_guide.md
@@ -260,7 +260,7 @@ TO PREDICT result_table_reference
 USING model_table_reference;
 ```
 
-**NOTE: The select statement must have the same logic used in training, e.g. you use `SELECT a, b * 2 FROM train_data TO TRAIN` to train a model, when predicting, you must use also `SELECT a, b * 2 FROM pred_data TO PREDICT` so that the model will behave properly. You should do the same when using `TO EXPLAIN` and `TO EVALUATE` clause.**
+**Please be aware that the `SELECT ...` statement before the `TO PREDICT ...` and `TO EXPLAIN ...` clause must generate the same data schema as the one before the `TO TRAIN` clause does. This requirement ensures that the data used in prediction has the same schema as those used in training.**
 
 The select statement syntax is the same as the select statement syntax in the training syntax. SQLFlow uses the column name to guarantee the prediction data has the same order as the training data. For example, if we have used `c1`, `c2`, `c3` and `label` column to train a model, the select statement in the prediction job should also retrieve columns that contain exactly the same names.
 
@@ -332,7 +332,7 @@ USING TreeExplainer;
 
 ## Evaluation Syntax
 
-A SQLFlow prediction statement consists of a sequence of select, predict, and using clauses.
+A SQLFlow prediction statement consists of select, evaluate, and into clauses.
 
 ```sql
 SELECT select_expr [, select_expr ...]
@@ -345,11 +345,13 @@ TO EVALUATE model_table_reference
 INTO evaluate_result_table;
 ```
 
-The select statement syntax is the same as the select statement syntax in the training syntax. When doing the evaluation, you should use a dataset that is not used for training to evaluate the model's performance when dealing with "new" data. The evaluation result will be saved in `evaluate_result_table` when the evaluation job completes.
+The SELECT statement before TO EVALUATE must generate the same data schema as that used in training. Please be aware that, by the machine learning theory, you are not supposed to use the training dataset for evaluation. If you do that, you are likely to see a very good evaluation result, which indeed, doesn't tell much useful information. Please write a SELECT statement as the prefix that generates different dataset but have the same schema.
+
+The INTO clause names the table for saving evaluation results, which, in this example, is `evaluate_result_table`.
 
 ### Evaluate and Into Clause
 
-The *evaluate clause* describes which model will be used to evaluate, what metrics will be used for evaluation, and which table will be used to save the result.
+The `TO EVALUATE` clause specifies the model to evaluate, what metrics will be used for evaluation, and which table will be used to save the result.
 
 ```sql
 TO EVALUATE model_table_reference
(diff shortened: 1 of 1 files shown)


doc/language_guide.md
```diff
@@ -328,6 +330,39 @@ WITH
...
+
+```sql
+TO EVALUATE model_table_reference
+[WITH
+  attr_expr [, attr_expr ...]]
+INTO evaluate_result_table;
+```
+
+- *model_table_reference* indicates that this model will be used to do the evaluation. The result metrics can show the performance of this model when using this model to predict "new" data.
+- *attr_expr* set attributes when doing evaluation. You can set `validation.metrics` to indicate which metrics will be outputed to the result table, e.g. `validation.metrics="Accuracy,AUC"`, you can find more supported metric names for Keras models [here](https://www.tensorflow.org/api_docs/python/tf/keras/metrics) and XGBoost models [here](https://xgboost.readthedocs.io/en/latest/parameter.html).
```
sneaxiy: `*attr_expr* sets attributes` ?

doc/language_guide.md
```diff
@@ -328,6 +330,39 @@ WITH
...
+FROM table_references
+  [WHERE where_condition]
+  [LIMIT row_count]
+TO EVALUATE model_table_reference
+[WITH
+  attr_expr [, attr_expr ...]]
+INTO evaluate_result_table;
+```
+
+The select statement syntax is the same as the select statement syntax in the training syntax. When doing the evaluation, you should use a dataset that is not used for training to test evaluate the model dealing with "new" data. The evaluation result will be saved in `evaluate_result_table` when the evaluation job completes.
```
sneaxiy: `... that is not used for training to evaluate...`?

sneaxiy (commented):
Almost LGTM for tiny changes.

sneaxiy (dismissed):
LGTM.

doc/language_guide.md
```diff
@@ -260,6 +260,8 @@ TO PREDICT result_table_reference
 USING model_table_reference;
 ```
 
+**NOTE: The select statement must have the same logic used in training, e.g. you use `SELECT a, b * 2 FROM train_data TO TRAIN` to train a model, when predicting, you must use also `SELECT a, b * 2 FROM pred_data TO PREDICT` so that the model will behave properly. You should do the same when using `TO EXPLAIN` and `TO EVALUATE` clause.**
```
wangkuiyi: - We don't need **NOTE** here.

----

Please be aware that the `SELECT ...` statement before the `TO PREDICT ..` clause must generate the same data schema as the one before the `TO TRAIN` clause does.  This requirement ensures that the data used in prediction has the same schema as those used in training.

doc/language_guide.md
```diff
@@ -328,6 +330,39 @@ WITH
 USING TreeExplainer;
 ```
 
+## Evaluation Syntax
+
+A SQLFlow prediction statement consists of a sequence of select, predict, and using clauses.
```
wangkuiyi: ~~a sequence of~~

doc/language_guide.md
```diff
@@ -328,6 +330,39 @@ WITH
...
+FROM table_references
+  [WHERE where_condition]
+  [LIMIT row_count]
+TO EVALUATE model_table_reference
+[WITH
+  attr_expr [, attr_expr ...]]
+INTO evaluate_result_table;
+```
+
+The select statement syntax is the same as the select statement syntax in the training syntax. When doing the evaluation, you should use a dataset that is not used for training to evaluate the model's performance when dealing with "new" data. The evaluation result will be saved in `evaluate_result_table` when the evaluation job completes.
```
wangkuiyi: The SELECT statement before `TO EVALUATE` must generate the same data schema as that used in training.  Please be aware that, by the machine learning theory, you are not supposed to use the training dataset for evaluation.  If you do that, you are likely to see a very good evaluation result, which indeed, doesn't tell much useful information.  Please write a SELECT statement as the prefix that generates different dataset but have the same schema.

The INTO clause names the table for saving evaluation results, which, in this example, is `evaluate_result_table`.

doc/language_guide.md
```diff
@@ -328,6 +330,39 @@ WITH
...
+
+### Evaluate and Into Clause
+
+The *evaluate clause* describes which model will be used to evaluate, what metrics will be used for evaluation, and which table will be used to save the result.
```
wangkuiyi: *evaluate clause* => `TO EVALUATE` clause

describes which model => specifies the model

will be used to evaluate => to evaluate

wangkuiyi (commented):
A few comments on English writing.

doc/language_guide.md
typhoonzero: Thanks for the hints. Done and updated.

wangkuiyi (approved):

PR #2329, opened 2020-05-21
Fix create predict result table error

typhoonzero:
Fix errors on PAI prediction:

```
failed executing create table predict_result (sepal_length double, sepal_width double, petal_length double, petal_width double, class bigint, class INT);: "ODPS-0130071:[1,144] Semantic analysis exception - column repeated in creation: class\n"
```


Yancey0623 (approved):
LGTM

PR #2331, opened 2020-05-22
Fix xgboost dist train data sharding problem

### LINKED ISSUE #2326: Data sharding problem in distributed XGBoost training using rabit
I am not sure how XGBoost distributed training using rabit shards input data. Supposing that the whole dataset is D, there are 2 usually methods to shard input data in distributed training:

- Method 1: we shard dataset D beforehand into N pieces, where N is the number of workers. Then, we distribute each piece of dataset to each worker. In this way, each worker reads unique data.
- Method 2: all workers can read the whole dataset D, and the each worker only pick up 1/N data inside the whole dataset D, say, what [tf.data.Dataset.shard](https://www.tensorflow.org/api_docs/python/tf/data/Dataset?version=nightly#shard) does.

In the doc of XGBoost and rabit, I have not found whether method 1 or 2 is used. But in the implementation of XGBoost, I found that maybe XGBoost uses method 2.

Link to my issue of XGBoost repo: https://github.com/dmlc/xgboost/issues/5694

sneaxiy:
Fix https://github.com/sql-machine-learning/sqlflow/issues/2326 .

### DIFF
--- a/python/sqlflow_submitter/xgboost/dataset.py
+++ b/python/sqlflow_submitter/xgboost/dataset.py
@@ -130,12 +130,11 @@ def load_dmatrix(filename):
     See https://github.com/sql-machine-learning/sqlflow/issues/2326
     in detailed.
     '''
-    # NOTE(sneaxiy): We can add `if xgb.rabit.get_world_size() > 1` here
-    # to call `xgb.DMatrix(filename)` directly, but it would
-    # make the code more complex. So I decide to unify the codes of
-    # non-distributed and distributed training.
-    csr_data = load_svmlight_file(filename, zero_based=True)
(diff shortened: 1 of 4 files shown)


python/sqlflow_submitter/xgboost/dataset.py
```diff
@@ -17,6 +17,7 @@
 from pathlib import Path
 
 import xgboost as xgb
+from sklearn.datasets import load_svmlight_file
```
weiguoz: Is it applicable to Python2？
We need a test in PAI environment.

weiguoz (approved):
LGTM
Excellent PR!

PR #2335, opened 2020-05-24
Overall design of the migration to python

shendiaomo:


doc/design/migration_to_python.md
```diff
@@ -0,0 +1,479 @@
+# Migrating to Python
+
+The design is about migrating SQLFlow golang code that's related to ML to python.
```
wangkuiyi: golang => Go
python => Python

doc/design/migration_to_python.md
```diff
@@ -0,0 +1,479 @@
...
+
+The last two sections explain how the proposed architecture solves the problem of the current architecture, as well as the preparations to be noted.
+
+## Flaws In The Current Architecture
+
+For SQLFlow, migrating to python might mean rewriting lots of existing golang code in python. Why should we migrate the golang *ML code* to python? Is is necessary? Is it the right time?
+
+After demonstrating the flaws of the current architecture in the following scenarios, the answer will be obvious.
+
+### Scenario 1: An ML Expert Wants A New Column Type
```
typhoonzero, line 19: Maybe ML experts ( model contributors ) do not have a chance to directly add a new column type, all this work should be done by the developers of SQLFlow.

doc/design/migration_to_python.md
```diff
@@ -0,0 +1,479 @@
...
+   3. Add a `switch` statement in the function `initColumnMap` in [pkg/step/feature/derivation.go](https://github.com/sql-machine-learning/sqlflow/tree/400c691470c6503393453d47856913df3365503e/pkg/step/feature/derivation.go).
+   4. Add a `case` in the large `switch` statement in function `generateFeatureColumnCode` in [pkg/sql/codegen/tensorflow/codegen.go](https://github.com/sql-machine-learning/sqlflow/tree/400c691470c6503393453d47856913df3365503e/pkg/sql/codegen/tensorflow/codegen.go). This step generates a snippet of python code with simple text substitution.
+   5. Check inputs for `image_embedding_column ` to avoid possible run-time exception from the python part to ensure user experience.
+3. Test and debug: a bug can be in the python file, in `pkg/ir/feature_column.go`, in `pkg/sql/ir_generator.go`, in `pkg/step/feature/derivation.go`, in `pkg/sql/codegen/tensorflow/codegen.go`, or in the python code text substitution template.
+
+See https://github.com/sql-machine-learning/sqlflow/pull/1897 and https://github.com/sql-machine-learning/sqlflow/pull/1901 for an real-world example.
+
+### Scenario 2: An ML Expert Wants To Contribute A New Type Of Model
```
typhoonzero, line 37: Same as above. Yet, after this refactor, we can make SQLFlow easier to do continuously evolving.

doc/design/migration_to_python.md
```diff
@@ -0,0 +1,479 @@
...
+Add a package named `sqlflow` in [sqlflow/python](https://github.com/sql-machine-learning/sqlflow/tree/400c691470c6503393453d47856913df3365503e/python) to contain the python code imgrated from golang. The directory structure of SQLFlow core code under `sqlflow/python/sqlflow` should be:
+
+| Package/Module Name | Description                                                  |
+| ------------------- | ------------------------------------------------------------ |
+| `platform`          | Modules under this package must implement the same set of functions(`run`, `train`, `evaluate`, `explain`, `predict` and `solve`) with the same signatures (e.g. `train(stmt: Statement)`) |
+| `columns`           | Modules under this package defines functions that can be called by users in a `COLUMN` clause. |
+| `features.py`       | Translate the `COLUMN` clause into calls to functions in the `column` package. |
+| `execute.py`        | Execute a SQLFlow program or a SQLFlow step.                 |
+| `client.py`         | A python client of sqlflow_server.                           |
+| `_database.so`      | Wrapper of the golang package `pkg/database`.                |
```
typhoonzero: GoMaxcompute use HTTP API, rewrite it using pure Python is not much effort.

doc/design/migration_to_python.md
```diff
@@ -0,0 +1,479 @@
...
+The golang package `executor` should parse the diagnostics messages generated by `diagnostics.py` and return them to the users.
+
+In the remainder of the section, we say **diagnostics are required** where `diagnostics.py` should be involved to generate an error message.
+
+##### Intermediate Representations
+
+As described above, the proposed architecture requires the golang part to pass the *ir* of a SQLFlow program to the python part to execute the program. We propose to define `ir` as a protobuf file in `pkg/proto/ir.proto`. It should have at least the following messages and fields, each other detailed design doc could add necessary fields to the `Statement` message.
+
+```protobuf
+message Statement {
```
typhoonzero, line 221: It's simple to just add JSON/YAML hints in a Go struct so that we won't need to define a proto, the IR's struct in `ir.go` itself defines the JSON/YAML serialization constraints.

typhoonzero (commented):
Maybe add some figures to show the current structure and refactored structure will be nice.

doc/design/migration_to_python.md
brightcoder01: Agree with @typhoonzero

doc/design/migration_to_python.md
```diff
@@ -0,0 +1,479 @@
...
+
+- Call the standard module `ast` to generate ASTs for the function call expressions in `columns`
+- Do semantic checking for every function nodes in the ASTs. **Diagnostics are required.**
+- Collect analysis requirements of the column functions from the nodes in the AST and construct analysis `select` statements
+- Construct a `ColumnsDAG` that represents the `columns` from the AST
+- Return the DAG and the analysis `select` statements to `workflow`. `workflow` should put the DAG in an Argo artifact that could be passed to the following steps. 
+
+`generate_feature_columns` does at least the following task:
+
+- Generate python column objects ( `tf.feature_columns` or a function that returns tensors e.g.) according to the `ColumnsDAG` and the results of the analysis `select` statements (which could be in a database table or in an artifact).
```
brightcoder01: Generate a python function built upon feature column api or [keras preprocessing layers](https://github.com/sql-machine-learning/elasticdl/tree/develop/elasticdl_preprocessing). The function accept input features, preprocess them and output tensors.

doc/design/migration_to_python.md
```diff
@@ -0,0 +1,479 @@
...
+| `client.py`         | A python client of sqlflow_server.                           |
+| `_database.so`      | Wrapper of the golang package `pkg/database`.                |
+| `database.py`       | Wrapper to `_database.so` for ease of use.                   |
+| `ir_pb2.py`         | Generated from `pkg/proto/ir.proto`.                         |
+| `contracts.py`      | See below.                                                   |
+| `diagnostics.py`    | See below.                                                   |
+
+#### Python Version And The Style
+
+The `sqlflow` package will only run in the *Docker container* and never be submitted to the PAI platform, so we propose to develop in python3 in the migration.
```
brightcoder01, line 164: Le's take tensorflow for example, the transform function for `column` clause will call the keras preprocessing layers or feature column api. It's a part of the complete model definition and will run on the PAI platform.

doc/design/migration_to_python.md
```diff
@@ -0,0 +1,479 @@
...
+  
+# ColumnsDAG is a structure to be designed in the detailed design doc.
+def generate_feature_columns(columns: ColumnsDAG):
+    """
+    `generate_feature_columns` generates python feature columns objects
+    """
+    pass
+```
+
+`generate_analysis_steps` does at least the following tasks:
```
brightcoder01, line 387: We can add the link to new version of COLUMN clause design here: https://github.com/sql-machine-learning/elasticdl/blob/develop/docs/designs/data_transform.md

doc/design/migration_to_python.md
```diff
@@ -0,0 +1,479 @@
...
+
+- Call the standard module `ast` to generate ASTs for the function call expressions in `columns`
```
brightcoder01, line 389: Do you mean `ast` python module? I think `AS` keyword in COLUMN clause will break the execution of `ast` module.

doc/design/migration_to_python.md
```diff
@@ -0,0 +1,479 @@
...
+
+16. [sqlfs](https://github.com/sql-machine-learning/sqlflow/tree/400c691470c6503393453d47856913df3365503e/pkg/sqlfs) wraps a database table into a file system interface and is required by the `model` package. We would migrate this package to python together with `model`.
+
+17. [table_writer](https://github.com/sql-machine-learning/sqlflow/tree/400c691470c6503393453d47856913df3365503e/pkg/table_writer) is a utility package for rendering database tables for a UI. We **don't** have to migrate this package to python. It's required by the `step` package and should be move to the `step` directory.
+
+## The Proposed Architecture
+
+### The Proposed Golang Package Structure
+
+Based on the foregoing analysis, the directory structure SQLFlow core code under `sqlflow/pkg` should be:
```
brightcoder01, line 131: It seems that we don't have the `code_gen` module in either go or python package. Does it mean that we won't generate code inside SQLFlow?

doc/design/migration_to_python.md
```diff
@@ -0,0 +1,479 @@
...
+    pass
+  
+if __name__ == '__main__':
+    main(sys.argv[1] if len(sys.argv) > 1 else "execute")
+```
+
+The python function `execute` executes the SQLFlow `program` and should do at least the following tasks:
+
+- Do semantic checking for the `program` using `contracts.py`. **Diagnostics are required**.
+- Call `features.py:generate_analysis_steps` to generate additional data transforming steps.
```
brightcoder01, line 329: For `TO TRAIN` statement, we always have an analysis step. Inside this step, we will generate the analysis SQL and execute.
Maybe we won't generate the analysis SQL before generating the workflow.

doc/design/migration_to_python.md
```diff
@@ -0,0 +1,479 @@
...
+
+The python function `execute` executes the SQLFlow `program` and should do at least the following tasks:
+
+- Do semantic checking for the `program` using `contracts.py`. **Diagnostics are required**.
+- Call `features.py:generate_analysis_steps` to generate additional data transforming steps. 
+- Call `couler` to generate an Argo `.yaml` including all `statement` steps and the additional steps and submit the workflow to Kubernetes.
+
+The python function `fetch` gets output of an executing `Program` in a incremental way.
+
+The python function `step` executes a SQLFlow `statement` as a workflow step. `step` should do at least the following tasks:
```
brightcoder01: A SQLFlow statement may generate more than one steps.

doc/design/migration_to_python.md
```diff
@@ -0,0 +1,479 @@
...
+
+- From the `platform` package, **locate** the python module of the **configured** platform.
+  - The configure mechanism should be described in detail in the detailed design doc (the current architecture use the environment variable `SQLFlow_submitter` to config the platform)
+  - The locating mechanism should be described in detail in the detailed design doc.
+-  Call one of `train`, `predict`, `explain` ,`run`, `solve`, `evaluate` of the python module of the platform according to `statement.type`
+
+There should be at least three modules in the `platform` package:
+
+- `pai.py`
+- `alisa.py`
```
brightcoder01, line 344: I think we only have `alisa` and `default`. Because we submit PAI-TF job via `alisa` too.

doc/design/migration_to_python.md
```diff
@@ -0,0 +1,479 @@
...
+    pass
+```
+
+`generate_analysis_steps` does at least the following tasks:
+
+- Call the standard module `ast` to generate ASTs for the function call expressions in `columns`
+- Do semantic checking for every function nodes in the ASTs. **Diagnostics are required.**
+- Collect analysis requirements of the column functions from the nodes in the AST and construct analysis `select` statements
+- Construct a `ColumnsDAG` that represents the `columns` from the AST
+- Return the DAG and the analysis `select` statements to `workflow`. `workflow` should put the DAG in an Argo artifact that could be passed to the following steps.
```
brightcoder01, line 393: IMO, the Go program will check the syntax and parse the `COLUMN` clause into a data structure which standards for a DAG directly.
The Python module will use the data structure to do the following feature derivation, data analysis and transform code generation.

doc/design/migration_to_python.md
```diff
@@ -0,0 +1,479 @@
...
+    pass
+```
+
+The `sqlflow` function simply connects to sqlflow_server running at the `server_addr`, the server would then create a `Program` for this single statement with `datasource` and start a python process to `execute` the program.
+
+## Does the Proposed Architecture Solve the Flaws in the Current Architecture?
+
+To answer the question, let's review the scenarios in the beginning of the document.
+
+### Scenario 1: An ML Expert Wants A New Column Type
```
brightcoder01, line 427: I prefer that the column is only maintained by SQLFlow developers.

doc/design/migration_to_python.md
shendiaomo: Yes, the proposed architecture doesn't have to generate a python script.

doc/design/migration_to_python.md
shendiaomo: I propose to use the python `ast` module to generate ASTs because `COLUMN`s are actually calling python functions, we don't have to rewrite a parser to parse the *python style function call syntax*.

Similarly, I propose that we don't invent the new syntax `a_column(b_column() as c)` because it's neither SQL-stylish nor a valid function call expression. `b_column() as c, a(c)` is more explicit, more SQL-stylish and easier to parse.

doc/design/migration_to_python.md
shendiaomo: Is this possible? If a model contributor requires some new feature columns in her innovative DNN, does she have to wait until SQLFlow developers have supported the new column?

typhoonzero: Actually, if a model developer want to do custom transformations in his model, he can write a custom Keras layer to do so, no need to add SQLFlow column support.

doc/design/migration_to_python.md
sneaxiy: If we do not generate a python script, the workflow script is implemented by calling python `platform.train/predict/evaluate` functions? So golang should call python (rpc or native calls)? If so, how about make it more clear if there is any cross language call?

doc/design/migration_to_python.md
```diff
@@ -0,0 +1,479 @@
...
+
+For SQLFlow, migrating to python might mean rewriting lots of existing golang code in python. Why should we migrate the golang *ML code* to python? Is is necessary? Is it the right time?
+
+After demonstrating the flaws of the current architecture in the following scenarios, the answer will be obvious.
+
+### Scenario 1: An ML Expert Wants A New Column Type
+
+Feature engineering is a very important part of machine learning. SQLFlow supports feature engineering via the `COLUMN` clause. The current `COLUMN` clause of SQLFlow is based on `tf.feature_column` and works only for TensorFlow. In the next version, the SQLFlow `COLUMN` clause will support more comprehensive feature columns for TensorFlow, PyTorch and XGBoost. 
+
+Suppose the user wants a `image_embedding_column ` that converts a input image into embedding for a recommender system. Adding such a new column in the current architecture requires the following main steps:
```
sneaxiy: `image_embedding_column ` -> `image_embedding_column`

doc/design/migration_to_python.md
sneaxiy: Do we use the standard python `ast` module here? Does the standard python `ast` module support to parse SQL statement, or user-defined DSL grammar? I remember the standard python `ast` module only supports to parse python code.

doc/design/migration_to_python.md
```diff
@@ -0,0 +1,479 @@
...
+  // `columns` contain a map of string to `Columns` For multiple COLUMN clauses
+  // like "COLUMN ... FOR deep_feature, COLUMN ... FOR wide_feature". They will
+  // be parsed as {"deep_feature": [...], "wide_feature": [...]}. For single
+  // column clause like "column ...", "feature_columns" will be used as the
+  // default map key.
+  optional map<string, Columns> columns = 8;
+  // `model_save` specifies the saved model path in the INTO/USING clause.
+  optional bool model_save = 9;
+  enum Type {
+    RUN = 0;
```
sneaxiy, line 257: `RUN` here means `TO RUN` or normal SQL statement? How to distinguish them?

sneaxiy (commented):
Agreed with @typhoonzero . We should need some figures.

doc/design/migration_to_python.md
shendiaomo: > `RUN` here means `TO RUN` or normal SQL statement? How to distinguish them?

It'' be fixed in  the next commit.

doc/design/migration_to_python.md
shendiaomo: > If we do not generate a python script, the workflow script is implemented by calling python platform.train/predict/evaluate functions? So golang should call python (rpc or native calls)? If so, how about make it more clear if there is any cross-language call?

1. golang starts a python process that calls the python function `execute()`
2. `execute()` generates a `.yaml` and submits the `.yaml` to Argo 
3. Each step in the `.yaml` starts the `step` binary
4. The `step` binary calls the python function `step()`
5. The `step()` python function calls python platform.train/predict/evaluate functions respectively.

lhw362950217: a input image -> an input image

doc/design/migration_to_python.md
```diff
@@ -0,0 +1,479 @@
...
+
+   1. The `workflow`  package has the same design as the *ML code* described above, it has similar flaws. 
+   2. In the next version of the `COLUMN` clause, the `COLUMN` clause may generate its own workflow steps to derive features or get analysis data. This process will requires some *ML code* and should be implemented in python.
+
+4. [step](https://github.com/sql-machine-learning/sqlflow/tree/400c691470c6503393453d47856913df3365503e/pkg/step) is the downstream of the `workflow` package, it's binary `cmd/step` is executed as a workflow step on Argo. The `step/feature` directory should be migrated to python because it has *ML code* that deduces feature columns from the input.
```
lhw362950217: Is a step's input still a SQL statement after migration? And the step parse and execute the SQL?

doc/design/migration_to_python.md
lhw362950217: We can wrap all Go DB driver into one Python module, so, we don't have to rewrite them one by one.

typhoonzero: 1. wrapping Go drivers still need to write Python interface to have `Exec`, `Query` functions and assemble the result in Python.
2. calling HTTP API in Python is simple to implement
3. `_database.so` maybe hard to support different OS since model developers may be able to call the Python API to debug models on their laptop.

typhoonzero: By the way, we only have to implement Alisa Python API, for maxcompute we have pyodps to use handy.

doc/design/migration_to_python.md
typhoonzero: Does each step execute a SQL statement?

- Execute `step -e "SELECT ... TO TRAIN"` in each step:
    - the step image must have a step binary, java parser, generate python code, and call the python code in a sub-process
- Directly execute `sqlflow.train()` in each step:
    - the step image only contain `sqlflow` Python API for use, don't need to parse the statement again.

FYI.

shendiaomo: Yes, actually, the python function `sqlflow.execute step` is designed for the purpose, eventually, we'll call `python -m sqlflow.execute step` to execute each step. I propose to do this in the second phase rather than in the MVP phase, because this requires to reimplement code in the golang package `step` and `tablewriter`, which will distract us from the main purpose of the migration.

In the MVP phase, we still reuse the `step` binary and focus on the migration of the *ML code* in golang. In the 2nd phase of the migration, we can migrate the golang package `step` at a right time.

doc/design/migration_to_python.md
shendiaomo: Yes, see earlier discussion https://github.com/sql-machine-learning/sqlflow/pull/2335#discussion_r431035131

doc/design/migration_to_python.md
shendiaomo: In the MVP phase, golang will start a `step` binary, the `step` binary calls the python executable module `sqlflow.execute step` that calls the `python platform.train/predict/evaluate` functions.
In the second phase, a workflow step should start `sqlflow.execute step` directly.

See https://github.com/sql-machine-learning/sqlflow/pull/2335#discussion_r431035131

doc/design/migration_to_python.md
shendiaomo: Done.

doc/design/migration_to_python.md
shendiaomo: Done.

doc/design/migration_to_python.md
shendiaomo: Yes, this part proposes to use a python-like function call syntax for the `COLUMN` clause.

doc/design/migration_to_python.md
shendiaomo: The current architecture supports `pai`. I propose to duplicate all the functionalities of the current architecture at the **MVP** phase.

doc/design/migration_to_python.md
shendiaomo: That's up to the syntax of the `COLUMN` clause, this part proposes we use the python function call syntax. You (and @sneaxiy ) are right if we've decided that the `COLUMN` clause are not to be compatible with the existing Python/Go function call syntax

doc/design/migration_to_python.md
shendiaomo: It's done.

doc/design/migration_to_python.md
shendiaomo: Done.

doc/design/migration_to_python.md
shendiaomo: I prefer protobuf because it can directly map to structs/classes in both python and golang. JSON/YAML is much more fragile if used as `dicts/lists` because they're less structural and lost the protection of the type system and the syntax of python.

doc/design/migration_to_python.md
shendiaomo: You're right. I'll fix this later.

weiguoz:
I close this PR due to a lack of activity.

PR #2342, opened 2020-05-25
Add find_fastest_apt_source.sh

wangkuiyi:
for building the `sqlflow:dev` and `sqfllow:ci` Docker images.

Currently, in `install.sh`, we hard code the APT source

https://github.com/sql-machine-learning/sqlflow/blob/cd3c42b95cad2ca09e4e68c7eff4c1672d6b606f/docker/dev/install.sh#L18-L26

With the introduction of this scripts, we can do 

```bash
sudo (find_fastest_apt_source.sh > /etc/apt/sources.list)
```

----

I know that we can have a long list of APT source mirrors by calling

```bash
curl -s mirrors.ubuntu.com/mirrors.txt
```

However, it is too long to test every choice. Also, this list is far from complete -- it doesn't include any site in China.

### DIFF
--- a/docker/dev/find_fastest_apt_source.sh
+++ b/docker/dev/find_fastest_apt_source.sh
@@ -21,8 +21,6 @@ mirrors.aliyun.com
 mirrors.ustc.edu.cn
 mirrors.163.com
 archive.ubuntu.com
-ftp.de.debian.org
-packages.debian.org
 EOM
 
 # Find the fastest APT source using ping.
--- a/docker/dev/find_fastest_apt_source.sh
+++ b/docker/dev/find_fastest_apt_source.sh
@@ -1,0 +1,44 @@
+#!/bin/bash
+
+# Define a list of mirrors without using Bash arrays.
+# c.f. https://stackoverflow.com/a/23930212/724872
+read -r -d '' APT_SOURCES << EOM
+mirrors.aliyun.com
+mirrors.ustc.edu.cn
+mirrors.163.com
+archive.ubuntu.com
+ftp.de.debian.org
+packages.debian.org
+EOM
+
+# Find the fastest APT source using ping.
+SPEED=99999.9
+for i in $APT_SOURCES; do
+    # c.f. https://stackoverflow.com/a/9634982/724872
+    echo "Testig speed of $i ..."
+    CUR_SPEED=$(ping -c 4 $i | tail -1| awk '{print $4}' | cut -d '/' -f 2)
+    echo "$CUR_SPEED"
+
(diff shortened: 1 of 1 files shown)


docker/dev/find_fastest_apt_source.sh
```diff
@@ -0,0 +1,44 @@
...
+
+# Define a list of mirrors without using Bash arrays.
+# c.f. https://stackoverflow.com/a/23930212/724872
+read -r -d '' APT_SOURCES << EOM
+mirrors.aliyun.com
+mirrors.ustc.edu.cn
+mirrors.163.com
+archive.ubuntu.com
+ftp.de.debian.org
+packages.debian.org
```
lhw362950217: `ftp.de.debian.or` and `packages.debian.org` are not ubuntu mirrors

wangkuiyi: What should i do? Remove them?

docker/dev/find_fastest_apt_source.sh
```diff
@@ -0,0 +1,44 @@
...
+        BEST_APT_SOURCE="$i"
+        SPEED="$CUR_SPEED"
+    fi
+done
+
+# The default Ubuntu version is 18.04, code named bionic.
+UBUNTU_CODENAME=${UBUNTU_CODENAME-"bionic"}
+
+# Write APT source lists.
+cat <<EOF
```
lhw362950217, line 45: should write to /etc/apt/sources.list

docker/dev/find_fastest_apt_source.sh
lhw362950217: YES. maybe change them to other mirrors

docker/dev/find_fastest_apt_source.sh
wangkuiyi: Let us write to the file at invocation time, like explained in https://github.com/sql-machine-learning/sqlflow/pull/2342#issue-422525337, so could we add Bash unit test for this script in the future.

lhw362950217 (approved):
LGTM

typhoonzero:
Can we try https://blog.zackad.dev/en/2018/01/18/how-to-select-fastest-ubuntu-repository-mirror.html if it works?

wangkuiyi:
> Can we try https://blog.zackad.dev/en/2018/01/18/how-to-select-fastest-ubuntu-repository-mirror.html if it works?

@typhoonzero This approach requires `pip install`.  However, we need the fastest APT source to install `pip`.  I intended to write this script, which depends only on ping, a tool pre-install in almost all Unix-like systems.

PR #2344, opened 2020-05-25
Add script to find fastest go proxy

### LINKED ISSUE #2343: Find faster go proxy for dev build in China
We need to find a Go proxy in China to speed up the develop build process.

Use `goproxy.cn` if we find it faster than the `pkg.go.dev` site

**Additional Notes**

lhw362950217:
Fixes #2343


docker/dev/find_fastest_go_proxy.sh
```diff
@@ -0,0 +1,46 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Define a list of mirrors without using Bash arrays.
+# c.f. https://stackoverflow.com/a/23930212/724872
+DEFAULT_URL="https://pkg.go.dev/"
+read -r -d '' GO_PROXIES <<EOM
+${DEFAULT_URL}
```
wangkuiyi: How comes this var?

lhw362950217: Just use it to test if we have to add a proxy. It seems if we can access this site fast, we do not need a proxy.

wangkuiyi: If it is only for testing, how about remove it from this change?

If it has more use than testing, it needs a precise name.  DEFAULT_URL is too general a name.

wangkuiyi: BTW, how should we decide if we need a Go proxy here?

docker/dev/find_fastest_go_proxy.sh
```diff
@@ -0,0 +1,46 @@
...
+https://goproxy.cn/,direct
```
wangkuiyi, line 20: How about we don't write `https://` and `,direct` here?  If we can do so, we can reuse a Bash function `find_fastest_url` and we don't need to call `cut` to retrieve the URL to the left.

docker/dev/find_fastest_go_proxy.sh
```diff
@@ -0,0 +1,46 @@
...
+        SPEED="$CUR_SPEED"
+    fi
+done
+
+# If best is default site, return empty proxy, ohterwise
+# return best proxy url
+if [[ "${BEST_URL}" == "${DEFAULT_URL}" ]]; then
+    echo ""
+else
+    echo "$BEST_URL"
```
wangkuiyi, line 44:
```bash
echo "https://$BEST_PROXY,direct"
```

where `$BEST_PROXY` comes from the ping speed comparison.

docker/dev/find_fastest_go_proxy.sh
lhw362950217: I mean we compare this site's access speed to other real `porxies`, if it actually is faster than proxies like `goproxy.cn`, we will not use any proxy.
Maybe I need to describe it in the script.

I'm not sure if there are Go proxies in America, if it does, we can use that proxies instead of this site.

wangkuiyi: I have never used a Go proxy here in America. I guess we don't need it here for public repo development.  A Go proxy, in my mind, is good for accelerating download in China specifically, and for redirecting employees to a corp-mirror of Go packages.

wangkuiyi (approved):
LGTM

PR #2346, opened 2020-05-25
Add model zoo input checkers

typhoonzero:


pkg/modelzooserver/input_check.go
```diff
@@ -13,12 +13,39 @@
...
+	}
+	return nil
 }
 
 // checkImageURL checks a Docker image URL is valid
-func checkImageURL(imageURL string) bool {
-	return true
+func checkImageURL(imageURL string) error {
+	if len(imageURL) < 6 {
+		return fmt.Errorf("imageURL should have at least 6 characters")
```
wangkuiyi, line 39: Why 6 characters?

pkg/modelzooserver/modelzooserver.go
```diff
@@ -176,6 +176,10 @@ func (s *modelZooServer) ReleaseModelDef(stream pb.ModelZooServer_ReleaseModelDe
 	}
 	// close and flush the tar.gz file
 	fd.Close()
+	err = checkImageURL(reqName)
```
wangkuiyi: How about keep the consistent style as the following lines and write

```golang
if err := checkImageURL(..); err {
    return err
}
```

pkg/modelzooserver/input_check.go
typhoonzero: A model definition image name like `a` seems not meaningful.

pkg/modelzooserver/input_check.go
```diff
@@ -13,12 +13,39 @@
...
 
+import (
+	"fmt"
+	"regexp"
+)
+
 // checkName checks model definition and trained model is valid.
-func checkName(name string) bool {
-	return true
+func checkName(name string) error {
```
weiguoz, line 22: How about adding a few unit tests as samples?

typhoonzero: Thanks, I just forgot to `git add` the test file...

weiguoz (approved):

PR #2355, opened 2020-05-26
SQLFlow TO RUN statement design doc.

brightcoder01:


doc/design/to_run_stmt.md
```diff
@@ -0,0 +1,178 @@
...
+INTO result_table
+```
+
+The SQLFlow statement above will call the python function `a_python_func` from the file `/run/my_func/data_preprocessor.py` inside the docker image `a_data_scientist/maxcompute_functions:1.0`. The attributes in the `WITH` clause will be passed to the python function as parameters.
+
+## Challenges
+
+- TO RUN can execute functions in Kubernetes and MaxCompute.
+- TO RUN can execute functions in both a single process and a distributed cluster.
+- TO RUN can execute functions built upon various data computing frameworks such as [Dask](https://github.com/dask/dask), [Mars](https://github.com/mars-project/mars), [Ray](https://github.com/ray-project/ray) and so on.
```
wangkuiyi: I think TO RUN should be able to run only a Python function in a process/container.  It is not the responsibility of TO RUN to consider the use of Mars, Dask, or Ray.

doc/design/to_run_stmt.md
```diff
@@ -0,0 +1,178 @@
...
+
+## Design
+
+### Fix the challenges
```
wangkuiyi: Fix => Address

doc/design/to_run_stmt.md
```diff
@@ -0,0 +1,178 @@
...
+
+The semantics of `TO RUN a_data_scientist/maxcompute_functions:1.0/my_func.data_preprocessor.a_python_func` is that `a_python_func` will be executed in a docker container for Kubernetes or a PyODPS node for MaxCompute. The implementation of `a_python_func` is fully customized by users. Because the execution environments for `a_python_func` are different between Kubernetes and MaxCompute, there are some differences considering these three challenges above.
```
wangkuiyi: This section is considering the execution environment/platform of TO RUN.  

In a container on Kubernetes and in a PyODPS node on MaxCompute are just two examples.  Not a complete list of environments.

doc/design/to_run_stmt.md
```diff
@@ -0,0 +1,178 @@
...
+- Execution environment  
+  `a_python_func` is executed inside the function image `a_data_scientist/maxcompute_functions:1.0`. We can customized this image at the `docker build` stage and make sure that all the dependencies `a_python_func` need are already installed in the image.
+
+- Single process execution  
+  User writes the data transform process using pandas, numpy, tsfresh and so on inside `a_python_func`.
+
+- Distributed cluster execution  
+  User can install Dask/Mars/Ray while building the docker image. Inside the implementation of `a_python_func`, we can create a Dask/Mars/Ray cluster, build a computing graph and submit the distribution execution using the API from these packages.
+
+MaxCompute
```
wangkuiyi: I don't think there is a configuration of PyODPS node on MaxCompute. In my mind, there is only one execution environment -- Docker container on Kubernetes.  It is the function's responsibility to launch a job on MaxCompute.

doc/design/to_run_stmt.md
```diff
@@ -0,0 +1,178 @@
...
+
+- Execution environment  
+  The packages of PyODPS node are already pre-installed and user cannot customize it. `a_python_func` can only use these pre-installed python packages.
+
+- Single process execution
+  Since PyODPS node cannot guarantee it contains all the packages we need such as `tsfresh`, `a_python_func` create a mars cluster containing only one worker and tell this worker to execute another function `another_python_func` to do the data transformation via [Mars remote API](https://github.com/mars-project/mars/issues/1227). The cluster uses our function image `a_data_scientist/maxcompute_functions:1.0`, `another_python_func` and its dependencies are already installed while building this image.
+  
+- Distributed cluster execution  
+  The PyODPS node only installed Mars package, but not Dask and Ray. So we can create a cluster, build the computing graph and submit the distributed execution only using Mars API inside `a_python_func`.
```
wangkuiyi: What is a "computing graph"?

doc/design/to_run_stmt.md
```diff
@@ -0,0 +1,178 @@
...
+- Distributed cluster execution  
+  User can install Dask/Mars/Ray while building the docker image. Inside the implementation of `a_python_func`, we can create a Dask/Mars/Ray cluster, build a computing graph and submit the distribution execution using the API from these packages.
+
+MaxCompute
+
+- Execution environment  
+  The packages of PyODPS node are already pre-installed and user cannot customize it. `a_python_func` can only use these pre-installed python packages.
+
+- Single process execution
+  Since PyODPS node cannot guarantee it contains all the packages we need such as `tsfresh`, `a_python_func` create a mars cluster containing only one worker and tell this worker to execute another function `another_python_func` to do the data transformation via [Mars remote API](https://github.com/mars-project/mars/issues/1227). The cluster uses our function image `a_data_scientist/maxcompute_functions:1.0`, `another_python_func` and its dependencies are already installed while building this image.
```
wangkuiyi: Does Mars execute Docker images? Or Mars submit jobs to Kubernetes to execute Docker images? If it is the latter case, how is the relationship between the Kubernetes cluster used by Mars API and the one run SQLFlow generated steps?

doc/design/to_run_stmt.md
```diff
@@ -0,0 +1,178 @@
...
+We can express the data transform logic using Python, leverage many mature python packages and encapsulate it into a reusable function. And then we propose to add the `TO RUN` clause in SQLFlow to call it.
+
+## Syntax Extension
+
+```SQL
+SELECT * FROM source_table
+TO RUN a_data_scientist/maxcompute_functions:1.0/my_func.data_preprocessor.a_python_func
+WITH param_a = value_a,
+     param_b = value_b
+INTO result_table
```
wangkuiyi: I feel that we don't need INTO clause here.  The existence of the INTO clause implies that the Python function will generate a table. How if it doesn't generate any table, or if it generates more than one table?

We can use WITH clause to pass in table name(s) if the Python function needs it.

doc/design/to_run_stmt.md
```diff
@@ -0,0 +1,178 @@
...
+- output_table: the table name from the clause `INTO output_table`.
+- image_name: the docker image name from the clause `TO RUN image_name/func_name`.
+
+2. Attributes from `WITH` clause.
+
+Please check the following example. The first four parameters are from the context and param_1 ~ param_n are from `WITH` Attributes.
+
+```Python
+def a_python_func(
+    db_type,
```
wangkuiyi: Let us define the first parameter being a Python dictionary named `context`.  Users use the WITH clause to specify the rest of the parameters.

wangkuiyi:
@brightcoder01 Please review https://github.com/sql-machine-learning/sqlflow/pull/2371

doc/design/to_run_stmt.md
brightcoder01: Updated :)

doc/design/to_run_stmt.md
brightcoder01: Good suggestion, I'll add a section `Execution Platforms`. 
And we will focus on the implementation details in Kubernetes and MaxCompute at this doc.

doc/design/to_run_stmt.md
brightcoder01: The python program running in the container can call Mars/Dask API to create a mars/dask cluster and submit a distributed computing job to this cluster.

doc/design/to_run_stmt.md
brightcoder01: Let's take mars for example. After creating mars cluster, user will use mars API to compose the distributed data computing logic. The logic will be translated to a computing graph and submit to the cluster. [link](https://docs.pymars.org/en/latest/distributed/architecture.html)

doc/design/to_run_stmt.md
brightcoder01: Mars submit jobs to kubernetes to execute docker image.

The docker in the generated step plays the role of the client to submit the distributed data computing job. The docker in the kubernetes cluster is the one which execute the data computing logic. Even though the roles are different, we can use the same docker image. We need install the mars package the packages we use to do the data preprocessing just like `TSFresh`.

doc/design/to_run_stmt.md
brightcoder01: Good idea. According to the following discussion, we pass the context field using environment variable.

doc/design/to_run_stmt.md
brightcoder01: `INTO` clause is optional here. If we don't need output any data, we don't need the `INTO` suffix.

doc/design/to_run_stmt.md
brightcoder01: Yes, we can only execute docker run in the workflow step.
For different platforms, the commands are different for the docker run: `python a_python_file.py` for Vanilla Kubernetes and `alisa.submitter a_python_file.py` for MaxCompute. It's the responsibility of `alisa.submitter` to submit a PyODPS task to MaxCompute to execute the Python program.
We can check the `Execution Platforms` section in the update design.

wangkuiyi (approved):

Issue #2347, opened 2020-05-26
[TODO LIST] XGBoost feature column support


sneaxiy:
Related to https://github.com/sql-machine-learning/sqlflow/pull/2345 .

- [x] TensorFlow-like feature column transformer in Python
- [x] Data transformation before dumping to DMatrix in Python db generator
- [x] Golang `deriveFeatureColumnCode` code for XGBoost
- [x] Feature column support for train/predict/evaluate/explain codegen template of XGBoost
- [x] Save PMML file during training

sneaxiy:
Fixed by https://github.com/sql-machine-learning/sqlflow/pull/2361, https://github.com/sql-machine-learning/sqlflow/pull/2416 and https://github.com/sql-machine-learning/sqlflow/pull/2411 .

Issue #2190, opened 2020-05-07
COLUMN clause for XGBoost


brightcoder01:
The root discussion issue for Data analysis and transform is [here](https://github.com/sql-machine-learning/elasticdl/issues/1670).

**Transformation Requirements**
XGBoost model only accept the numerical values as input.
For numerical input, we don't need transformation such as normalization/standardization, [issue](https://github.com/dmlc/xgboost/issues/357).
For categorical input, we need convert it into numerical before feeding it to the model. The typical transformation is `convert to one hot vector`, `convert to integer id by looking up vocabulary` and `hash`. We can use `ONEHOT`, `VOCABULARIZE`, `HASH` transform function in SQLFlow COLUMN clause. 

 ```SQL
SELECT cont_1, cont_2, cont_3, cat_1, cat_2, cat_3
FROM source_table
TO TRAIN xgboost
COLUMN VOCABULARIZE(cat_1), HASH(cat_2), ONEHOT(cat_3)
INTO my_model
```

Let's take `VOCABULARIZE` function for example. The workflow contains two steps:
1. Do data analysis on the column of string type, and get the vocabulary.
2. Do the transformation of mapping string to id at the training stage.

**Discussion Points**
1. How do we execute the string to id mapping transformation?
- Do this transformation during the process of generating DMatrix from the source table.
*For example, load the source data as numpy array -> do transformation on the numpy array -> build DMatrix from this preprocessed numpy array -> feed DMatrix to XGBoost for training. Please check the [official document](https://xgboost.readthedocs.io/en/latest/python/python_intro.html#data-interface)*.
- Do this transformation in each iteration of XGBoost training. 
*Does XGBoost have the flexibility? - To be investigated*
- Build a [sklearn pipeline](https://scikit-learn.org/stable/modules/compose.html), the pipeline contains transformation and XGBoost training. [sample video](https://campus.datacamp.com/courses/extreme-gradient-boosting-with-xgboost/using-xgboost-in-pipelines?ex=7).

2. How to save the transform logic together with the model?
- Export both the model and transform logic together to PMML.
*We can try to use the sklearn pipeline to compose the preprocess logic and XGBoost model. And then use [sklearn2pmml](https://github.com/jpmml/sklearn2pmml) to export the transformation and model to PMML*

3. Distributed data preprocess and XGBoost training?
- **[Need Investigation] Can we use sklearn pipeline for XGBoost distributed training?**
[...]

PR #2361, opened 2020-05-27
Support feature column in XGBoost models

sneaxiy:
Add XGBoost feature column implementation.

See discussion here: https://github.com/sql-machine-learning/sqlflow/issues/2190 .

See TODO list here: https://github.com/sql-machine-learning/sqlflow/issues/2347 .

See design doc here: [Feature Column Support in SQLFlow XGBoost models](https://github.com/sql-machine-learning/sqlflow/blob/develop/doc/design/xgboost_feature_column.md) .

This PR includes:

- [ ] TensorFlow-like feature column transformer in Python
   - [x] `NUMERIC`, `BUCKET`, `CATEGORY_ID`, `CATEGORY_HASH`
   - [ ] `INDICATOR` . Convert a category id to a one-hot vector. To be implemented in next PR.
   - [ ] ~~`CROSS`. There is no requirement to implement crossed features yet.~~
   - [ ] ~~`SEQ_CATEGORY_ID` . Convert the column data to a varying length Tensor. Not quite sure it is useful in XGBoost model, because varying length data for XGBoost models may be difficult to implement.~~
   - [ ] ~~`EMBEDDING` . Not very possible to implement, because XGBoost cannot update the embedding parameter.~~
- [x] Data transformation before dumping to DMatrix in Python db generator
- [x] Golang `deriveFeatureColumnCode` code for XGBoost
- [x] Feature column support for train/predict/evaluate/explain codegen template of XGBoost
- [ ] Save PMML file during training. Use [sklearn2pmml](https://github.com/jpmml/sklearn2pmml) to implement.


docker/ci/install-pips.bash
```diff
@@ -30,4 +30,4 @@ pip install --quiet \
     oss2==2.9.0 \
     plotille==3.7 \
     seaborn==0.9.0 \
-    scikit-learn==0.21.0
+    scikit-learn==0.21.0
```
typhoonzero: Seems need to have a newline at the end of the file.

pkg/sql/codegen/common.go
```diff
@@ -0,0 +1,129 @@
+// Copyright 2020 The SQLFlow Authors. All rights reserved.
```
weiguoz: The meaning of function `GenerateFeatureColumnCode` is clear, so I would suggest renaming `common.go` to another name, like: `featurecolumngen.go` or something else.

pkg/sql/codegen/common.go
```diff
@@ -0,0 +1,129 @@
...
+		return fmt.Sprintf("%s.dtypes.string", module), nil
+	default:
+		return "", fmt.Errorf("unsupport dtype: %d", dtype)
+	}
+}
+
+// TODO(sneaxiy): XGBoost does not support some feature columns, such as EMBEDDING.
+// For better error message, we should find a better way to distinguish whether the
+// module is TensorFlow or XGBoost.
+func isTensorFlowModule(module string) bool {
```
brightcoder01: In the next step, we will support PyTorch.
We cannot assume that !isTensorFlowModule == isXGBoostModule

pkg/sql/codegen/common.go
```diff
@@ -0,0 +1,129 @@
...
+	case *ir.CategoryHashColumn:
+		// FIXME(typhoonzero): do we need to support dtype other than int64?
+		dtype, err := toModuleDataType(c.FieldDesc.DType, module)
+		if err != nil {
+			return "", err
+		}
+		return fmt.Sprintf("%s.feature_column.categorical_column_with_hash_bucket(key=\"%s\", hash_bucket_size=%d, dtype=%s)",
+			module, c.FieldDesc.Name, c.BucketSize, dtype), nil
+	case *ir.CrossColumn:
+		var keysGenerated = make([]string, len(c.Keys))
```
brightcoder01: Do we support cross in XGBoost?

pkg/sql/codegen/common.go
typhoonzero: Name the file like `generate_feature_column_code.go` maybe more meaningful.

sneaxiy:
I would separate this PR into 3 pieces:

- Add test data.
- Common codegen codes movement.
- XGBoost feature column support.

pkg/util/util.go
```diff
@@ -0,0 +1,24 @@
+// Copyright 2020 The SQLFlow Authors. All rights reserved.
```
typhoonzero: Try not to add `util` package. `util` is not meaningful, the functions will be hard to find if other callers need to use them. Just put the function in the package where it's used.

pkg/sql/codegen/xgboost/codegen.go
```diff
@@ -203,6 +204,41 @@ func resolveFeatureMeta(fds []ir.FieldDesc) ([]byte, []string, error) {
 	return f, featureNames, e
 }
 
+func generateFeatureColumnCode(fc ir.FeatureColumn) (string, error) {
+	return codegen.GenerateFeatureColumnCode(fc, "xgboost_extended")
+}
+
+func deriveFeatureColumnCode(fcs []ir.FeatureColumn, labelFc ir.FeatureColumn) (featureColumnsCode string, fieldDescs []ir.FieldDesc, label ir.FieldDesc, err error) {
```
typhoonzero: Feature columns and "Field Descs" are not equal, the function called `deriveFeatureColumnCode` while it's returning feature columns together with field descs.

cmd/sqlflowserver/main_test.go
```diff
@@ -226,9 +226,11 @@ func prepareTestData(dbStr string) error {
 			datasets = []string{
 				testdata.ODPSFeatureMapSQL,
 				testdata.ODPSSparseColumnSQL,
-				fmt.Sprintf(testdata.IrisMaxComputeSQL, caseDB)}
+				fmt.Sprintf(testdata.IrisMaxComputeSQL, caseDB),
+				fmt.Sprintf(testdata.ChurnMaxComputeSQL, caseDB)}
 		} else {
-			datasets = []string{fmt.Sprintf(testdata.IrisMaxComputeSQL, caseDB)}
+			datasets = []string{fmt.Sprintf(testdata.IrisMaxComputeSQL, caseDB),
```
brightcoder01, line 232: Make the format pretty?

```GO
datasets = []string{
                         fmt.Sprintf(testdata.IrisMaxComputeSQL, caseDB),
		         fmt.Sprintf(testdata.ChurnMaxComputeSQL, caseDB)
}
```

docker/ci/install-pips.bash
brightcoder01: Is it a code format warning?

pkg/sql/codegen/xgboost/codegen.go
brightcoder01: How about add more description for this method？

pkg/util/util.go
```diff
@@ -0,0 +1,24 @@
...
+package util
+
+import (
+	"fmt"
+	"strings"
+)
+
+// IntArrayToJSONString converts int array to json string
+func IntArrayToJSONString(intArray []int) string {
+	return strings.Join(strings.Split(fmt.Sprint(intArray), " "), ",")
```
brightcoder01: The logic is join a string array with `,`, the output is a common separated string instead of a JSON

python/sqlflow_submitter/xgboost/predict.py
```diff
@@ -32,40 +32,52 @@ def pred(datasource,
...
+        transform_fn=transform_fn,
+        feature_column_code=feature_column_code,
+        raw_data_dir="predict.raw.dir")  # NOTE: default to use external memory
     bst = xgb.Booster({'nthread': 4})  # init model
     bst.load_model("my_model")  # load data
     print("Start predicting XGBoost model...")
+
+    if is_pai:
+        selected_cols = db.pai_selected_cols(
+            "odps://{}/tables/{}".format(*pai_table.split(".")))
```
brightcoder01: We can set `odps://{}/tables/{}` as a constant. It's more readable.

python/sqlflow_submitter/xgboost/train.py
```diff
@@ -172,4 +189,10 @@ def save_model(model_dir, model_params, train_params, feature_metas,
         train_params,
         feature_metas,
         feature_column_names,
-        label_meta)
+        label_meta,
+        feature_column_code)
+
+
+def save_local_file(booster, model_name):
```
brightcoder01: Can we add more description about the difference between save_local_file and save_model?
It may Include the export file path (local or remote) and the content.

cmd/sqlflowserver/main_test.go
```diff
@@ -1598,6 +1606,116 @@ FROM housing.xgb_predict LIMIT 5;`)
...
+	executeSQLFunc(trainSQL)
+
+	predictTableName := fmt.Sprintf("%sxgb_fc_test_predict_table_%d", dbPrefix, uniqueID)
+	predictSQL := fmt.Sprintf(`SELECT %s FROM %s TO PREDICT %s.%s_new USING %s;`, selectColumns, table, predictTableName, label, modelName)
+	executeSQLFunc(predictSQL)
+
+	if !isPai { // PAI does not support evaluate now
+		evaluateTableName := fmt.Sprintf("%sxgb_fc_test_evaluate_table_%d", dbPrefix, uniqueID)
+		evaluateSQL := fmt.Sprintf(`
+SELECT %s FROM %s TO EVALUATE %s WITH validation.metrics="accuracy_score" LABEL %s INTO %s;`,
```
weiguoz: Join L1669 to L1668, just like L1663 does.

pkg/sql/codegen/common.go
```diff
@@ -0,0 +1,129 @@
...
+		if !isTensorFlowModule(module) {
+			return "", fmt.Errorf("INDICATOR is only supported in TensorFlow models")
+		}
+
+		sourceCode, err := GenerateFeatureColumnCode(c.CategoryColumn, module)
+		if err != nil {
+			return "", err
+		}
+		return fmt.Sprintf("%s.feature_column.indicator_column(%s)", module, sourceCode), nil
+
```
weiguoz: We don't need a blank line here. [Go switch exapmle](https://github.com/golang/go/wiki/Switch#switch)

python/sqlflow_submitter/db.py
```diff
@@ -295,31 +297,24 @@ def pai_maxcompute_db_generator(table,
...
-                    yield list(row), label
+                row = pai_reader.read(num_records=1)[0]
+                if label_index is not None:
+                    yield list(row), row[label_index]
                 else:
                     yield list(row), None
-            except Exception as e:
-                reader.close()
+            except:
+                pai_reader.close()
```
weiguoz, line 311: Shall we move the `pai_reader.close()` to a `finally` block to ensure the reader will be closed?

pkg/sql/codegen/common.go
sneaxiy: Done. `common.go` has been moved to `codegen_feature_column.go` in https://github.com/sql-machine-learning/sqlflow/pull/2402 .

pkg/sql/codegen/common.go
sneaxiy: Thanks! Renamed `isTensorFlowModule` to be `isXGBoostModule` in https://github.com/sql-machine-learning/sqlflow/pull/2402 .

pkg/sql/codegen/common.go
sneaxiy: Yes. We support `CROSS` in XGBoost.

docker/ci/install-pips.bash
sneaxiy: Fixed.

pkg/sql/codegen/common.go
sneaxiy: Done. `common.go` has been moved to `codegen_feature_column.go` in https://github.com/sql-machine-learning/sqlflow/pull/2402 .

pkg/util/util.go
sneaxiy: Done. Package `util` is removed.

pkg/sql/codegen/xgboost/codegen.go
sneaxiy: Done. Renamed to `deriveFeatureColumnCodeAndLabelDescs` .

docker/ci/install-pips.bash
sneaxiy: Yes. Fixed.

pkg/sql/codegen/xgboost/codegen.go
sneaxiy: Added.

pkg/util/util.go
sneaxiy: `fmt.Sprint([]int{1, 2, 3})` returns `[1 2 3]`. So this function is indeed converts int array to json string.

python/sqlflow_submitter/xgboost/predict.py
sneaxiy: Thanks. Done.

python/sqlflow_submitter/xgboost/train.py
sneaxiy: Thanks. I planned to use `save_local_file` to do 2 things: (1) save the booster; (2) save the PMML file. But this PR is not related to PMML, so I remove `save_local_file` here.

cmd/sqlflowserver/main_test.go
sneaxiy: Done.

pkg/sql/codegen/common.go
sneaxiy: Done.

python/sqlflow_submitter/db.py
sneaxiy: Great idea! Done.

sneaxiy:
@typhoonzero @brightcoder01 @weiguoz I have moved some original codes of this PR to https://github.com/sql-machine-learning/sqlflow/pull/2401 and https://github.com/sql-machine-learning/sqlflow/pull/2402 . Maybe it is better to review now.

python/sqlflow_submitter/db.py
sneaxiy: Great idea! But we cannot replace `except` with `finally` here because (1) `finally` block would always run no matter whether there is exception; (2) `finally` cannot suppress exceptions, and we want to suppress the out-of-range exception raised by `pai_reader.read()` here. 

I have tried another way to polish this code. Please see [here](https://github.com/sql-machine-learning/sqlflow/pull/2361/files#diff-90d40b4a29a5b5da3f86836b1a986a33R309). Notice that exception would be always raised when all data is read out.

weiguoz (dismissed):
LGTM ++
Also, please @typhoonzero @brightcoder01 take a look.

typhoonzero (dismissed):
LGTM++, please @brightcoder01 take a look at `python/sqlflow_submitter/xgboost/feature_column.py` can later merge with current columns design.

sneaxiy:
I wonder should we follow the calculation logic of TensorFlow in feature column like `CATEGORY_HASH`, `CROSS`? In this PR, I just implement them by the standard Python API `hash` (we can re-implement them in the near future). @typhoonzero @brightcoder01

pkg/sql/codegen/pai/template_xgboost.go
```diff
@@ -153,7 +169,11 @@ model_params,
 train_params,
 feature_metas,
 feature_column_names,
-label_meta) = model.load_metas("{{.OSSModelDir}}", "xgboost_model_desc")
+label_meta,
+feature_columns_code) = model.load_metas("{{.OSSModelDir}}", "xgboost_model_desc")
+
+feature_column_transformers = eval('[{}]'.format(feature_column_code))
+transform_fn = xgboost_extended.feature_column.ComposedColumnTransformer(feature_column_names, *feature_column_transformers)
```
brightcoder01, line 176: Please add a description here that in the current implementation, we are generating a `transform_fn` from COLUMN clause. The `transform_fn` is executed during the process of dumping the original data into DMatrix.

python/sqlflow_submitter/xgboost/dataset.py
```diff
@@ -131,8 +195,13 @@ def load_dmatrix(filename):
     in detailed.
     '''
     if xgb.rabit.get_world_size() > 1:
+        if '#' in filename:
```
brightcoder01, line 205: Please add description what `if '#' in filename` means.
A shard of the training data file?

python/sqlflow_submitter/xgboost/feature_column.py
```diff
@@ -0,0 +1,243 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import numpy as np
+
+__all__ = [
+    'numeric_column',
```
brightcoder01, line 17: It seems that the API name is the same as TensorFlow feature column API. We can keep the name here at the current stage.
In the next step, we can rename the transform function in the COLUMN clause according to [the design doc](https://github.com/sql-machine-learning/elasticdl/blob/develop/docs/designs/data_transform.md) for TensorFlow/PyTorch/XGBoost in one shot.

python/sqlflow_submitter/xgboost/feature_column.py
```diff
@@ -0,0 +1,243 @@
...
+# TODO(sneaxiy): implement faster and proper hash algorithm
+def hashing(x):
+    return hash(x)  # use builtin hash function
+
+
+def apply_transform_on_value(feature, transform_fn):
+    if len(feature) == 1:  # dense
+        return transform_fn(feature[0]),
+    else:
+        return feature[0], transform_fn(feature[1]), feature[2]
```
brightcoder01, line 34: Is it a sparse input? Can we add description about what each element in this tuple means.

python/sqlflow_submitter/xgboost/predict.py
```diff
@@ -32,40 +32,52 @@ def pred(datasource,
...
+        batch_size=DEFAULT_PREDICT_BATCH_SIZE,
+        transform_fn=transform_fn,
+        feature_column_code=feature_column_code,
+        raw_data_dir="predict.raw.dir")  # NOTE: default to use external memory
     bst = xgb.Booster({'nthread': 4})  # init model
     bst.load_model("my_model")  # load data
     print("Start predicting XGBoost model...")
+
+    if is_pai:
+        pai_table = "odps://{}/tables/{}".format(*pai_table.split("."))
```
brightcoder01, line 63: What if pai_table doesn't contains maxcompute project name?
Will we append the project name before code_gen?

python/sqlflow_submitter/xgboost/train.py
```diff
@@ -153,16 +163,23 @@ def train(datasource,
...
+            save_model(oss_model_dir, model_params, train_params,
+                       feature_metas, feature_column_names, label_meta,
+                       feature_column_code)
 
 
 def save_model(model_dir, model_params, train_params, feature_metas,
-               feature_column_names, label_meta):
-    model.save_file(model_dir, "my_model")
+               feature_column_names, label_meta, feature_column_code):
+    model_name = "my_model"
```
brightcoder01: Why do we hard code the model name is `my_model`?

python/sqlflow_submitter/xgboost/feature_column.py
```diff
@@ -0,0 +1,243 @@
...
+        CategoricalColumnTransformer._set_field_names(self, field_names)
+        self.column_idx = self.field_names.index(self.key)
+
+    def get_column_names(self):
+        return [self.key]
+
+    def __call__(self, inputs):
+        def transform_fn(slot_value):
+            if isinstance(slot_value, np.ndarray):
+                output = np.ndarray(slot_value.shape)
```
brightcoder01, line 176: Maybe `output = [hashing(item) for item in slot_value]` is cleaner.

python/sqlflow_submitter/xgboost/__init__.py
```diff
@@ -10,3 +10,15 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
+from sqlflow_submitter.xgboost import feature_column
+
+
+class DataTypeCollection(object):
```
brightcoder01, line 17: We can use this style:
```python
class DataType(object):
    INT64 = "int64"
    FLOAT32 = "float32"
    STRING = "string"
```

python/sqlflow_submitter/xgboost/feature_column.py
```diff
@@ -0,0 +1,243 @@
...
+    def __call__(self, inputs):
+        slot_values = [
+            apply_transform_on_value(column(inputs), hashing)
+            for column in self.columns
+        ]
+        hash_values = [v[0] if len(v) == 1 else v[1] for v in slot_values]
+        return sum(hash_values) % self.hash_bucket_size,
+
+
+def crossed_column(keys, hash_bucket_size, hash_key=None):
```
brightcoder01: IMO, we don't need cross for XGBoost.

pkg/sql/codegen/pai/template_xgboost.go
sneaxiy: Done. Please see [here](https://github.com/sql-machine-learning/sqlflow/pull/2361/files#diff-d9418c7b645edf2bcca1f77a708f70d5R98).

python/sqlflow_submitter/xgboost/dataset.py
sneaxiy: '#' in XGBoost means use external memory cache. But we use `load_svmlight_files` to load DMatrix in distributed mode. '#' is not accepted by `load_svmlight_files`. Please see comment [here](https://github.com/sql-machine-learning/sqlflow/pull/2361/files#diff-85436bed613ae85f0caae0da7b565094R198).

python/sqlflow_submitter/xgboost/feature_column.py
sneaxiy: Agreed. We should rename them in the next step.

python/sqlflow_submitter/xgboost/feature_column.py
sneaxiy: Done. Please see [here](https://github.com/sql-machine-learning/sqlflow/pull/2361/files#diff-2f0634b1ebc2f4375900e5a40aaad875R33).

python/sqlflow_submitter/xgboost/predict.py
sneaxiy: We have appended project name [here](https://github.com/sql-machine-learning/sqlflow/blob/develop/pkg/sql/pai_submitter.go#L105).

python/sqlflow_submitter/xgboost/train.py
sneaxiy: Removed "my_model" here.

But string literal "my_model" is hard-coded in many places, for example [1](https://github.com/sql-machine-learning/sqlflow/blob/develop/pkg/sql/codegen/pai/template_xgboost.go#L38). I guess the key problem is what the name of the saved model is. Any name is OK but we should not hard code them in many places. We should clean them all in the near future.

python/sqlflow_submitter/xgboost/feature_column.py
sneaxiy: It may not work. Python builtin function `hash` does not support `numpy.ndarray` and for-range iteration of the `numpy.ndarray` would yield each `numpy.ndarray` whose dimension is one less than the original one.

python/sqlflow_submitter/xgboost/__init__.py
sneaxiy: It may not work. See [here](https://github.com/sql-machine-learning/sqlflow/blob/develop/pkg/sql/codegen/codegen_feature_column.go#L25). If we want to unify the codegen of TensorFlow and XGBoost, we must support data type style like `xgboost.dtypes.float32` (something corresponds to `tf.dtypes.float32`) instead of `xgboost.DataType.FLOAT32`.

python/sqlflow_submitter/xgboost/feature_column.py
sneaxiy: Removed.

pkg/util/util.go
brightcoder01: Can we use GO [JSON module](https://gobyexample.com/json) to convert the GO ARRAY to JSON?
In the current implementation, we are operating on the string. I think it's not so common and readable.

brightcoder01 (approved):

shendiaomo (approved):

PR #2363, opened 2020-05-27
End2end test modelzoo releasing model definition

typhoonzero:


pkg/modelzooserver/image_build.go
```diff
@@ -0,0 +1,104 @@
...
+	"os/exec"
+	"strings"
+)
+
+func imageExistsOnRegistry(imageName, tag string) bool {
+	var imageNamePart string
+	var registryPart string
+	registryPart = os.Getenv("SQLFLOW_MODEL_ZOO_REGISTRY")
+	imageNamePart = imageName
+	if strings.Contains(imageName, ".") {
```
lhw362950217, line 31: “." -> "/" ?

lhw362950217, line 33: Are we using '.' to determine whether the name is start with a domain？Maybe add some comment

python/sqlflow_submitter/import_custom_models.py
```diff
@@ -22,5 +22,6 @@ def import_model_def(estimator_name, namespace):
                 try:
                     namespace[model_name_parts[0]] = __import__(
                         model_name_parts[0])
+                    return model_name_parts[0]
```
lhw362950217: shall we add some log if the estimator_name is not in "a.b" format? Or, we just intend to ignore it.

typhoonzero: I think just ignore it will be fine.

lhw362950217 (approved):
LGTM

wangkuiyi (approved):

PR #2365, opened 2020-05-27
Make AutoClassifier compatible with python2

shendiaomo:
See https://github.com/sql-machine-learning/models/pull/68
This fixes an internal issue because `super()` is only available in python3 while we have to support python2 in PAI

### DIFF
--- a/docker/dev/build.sh
+++ b/docker/dev/build.sh
@@ -71,7 +71,7 @@ if [[ ! -d models ]]; then
     git clone https://github.com/sql-machine-learning/models
 fi
 cd models
-git checkout c897963f821d515651de79cb4ef1fbf6126ecaa5
+git checkout 07a75db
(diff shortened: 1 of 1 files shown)


typhoonzero (approved):

sneaxiy (approved):

docker/dev/build.sh
```diff
@@ -71,7 +71,7 @@ if [[ ! -d models ]]; then
     git clone https://github.com/sql-machine-learning/models
 fi
 cd models
-git checkout c897963f821d515651de79cb4ef1fbf6126ecaa5
+git checkout 07a75db
```
wangkuiyi, line 74: @shendiaomo Is there a reason that we need this special version of `models`? Should we comment and document the reason, or should we git tag the models repo?

PR #2369, opened 2020-05-28
Support hints for alisa

### LINKED ISSUE #2220: Support hints for maxcompute
Database engines, such as MaxCompute, supports [set operations](https://partners-intl.aliyun.com/help/doc-detail/96004.htm) to set MaxCompute system variables.
Example: `set odps.sql.mapper.split.size=4096`

The scope of this operation is inside a SQL program, no matter where it appears.

Also, such hints appear in `alisa` and [hive](https://cwiki.apache.org/confluence/display/Hive/LanguageManual+Commands).

weiguoz:
Fix #2220

### DIFF
--- a/pkg/sql/executor_ir.go
+++ b/pkg/sql/executor_ir.go
@@ -176,7 +176,7 @@ func runSingleSQLFlowStatement(wr *pipe.Writer, sql *parser.SQLFlowStmt, db *dat
 	return r.Execute(submitter)
 }
 
-// RewriteStatementsWithHints comines hints into stardard SQL
+// RewriteStatementsWithHints combines the hints into the standard SQL(s)
 func RewriteStatementsWithHints(stmts []*parser.SQLFlowStmt, dialect string) []*parser.SQLFlowStmt {
 	hints, sqls := splitHints(stmts, dialect)
 	if len(hints) > 0 {
--- a/pkg/sql/executor_ir.go
+++ b/pkg/sql/executor_ir.go
@@ -162,7 +162,7 @@ func runSingleSQLFlowStatement(wr *pipe.Writer, sql *parser.SQLFlowStmt, db *dat
 			r, err = generateEvaluateStmt(sql.SQLFlowSelectStmt, session.DbConnStr, modelDir, cwd, GetSubmitter(session.Submitter).GetTrainStmtFromModel())
 		}
 	} else {
-		standardSQL := ir.NormalStmt(sqlRewriter(sql.Original, hints))
+		standardSQL := ir.NormalStmt(rewriteSQLWithHints(sql.Original, hints))
 		r = &standardSQL
 	}
 	if err != nil {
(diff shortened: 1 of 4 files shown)


pkg/sql/executor_ir.go
```diff
@@ -174,6 +176,43 @@ func runSingleSQLFlowStatement(wr *pipe.Writer, sql *parser.SQLFlowStmt, db *dat
 	return r.Execute(submitter)
 }
 
+// RewriteStatementsWithHints comines hints into stardard SQL
```
typhoonzero: comines -> combines?

typhoonzero (dismissed):
LGTM++, not sure if we need to add a test to ensure that statements like `set k=v; select ...` works?

pkg/sql/executor_ir.go
weiguoz: sure.

weiguoz:
> LGTM++, not sure if we need to add a test to ensure that statements like `set k=v; select ...` works?

I have tested such a statement in `goalisa` manually, it works.

pkg/sql/executor_ir_test.go
```diff
@@ -263,3 +264,23 @@ func TestLogChanWriter_Write(t *testing.T) {
...
 }
+
+func TestRewriteStatementsWithHints4Alisa(t *testing.T) {
+	if os.Getenv("SQLFLOW_submitter") != "alisa" {
+		t.Skip("Skip test case: submitter is not alisa.")
+	}
+	dialect := "alisa"
+	hint1, hint2 := `set odps.stage.mapper.num=1;`, `set odps.sql.mapper.split.size=4096;`
+	standardSQL, extendedSQL := `select 1;`, `select 1 to predict d.t.f using m;`
+	sqlProgram := hint1 + standardSQL + hint2 + extendedSQL
```
wangkuiyi: How about adding one or more `\n`s and whitespace and `\t`s in between the statements?

pkg/sql/executor_ir.go
```diff
@@ -174,6 +176,43 @@ func runSingleSQLFlowStatement(wr *pipe.Writer, sql *parser.SQLFlowStmt, db *dat
...
+			sqls = append(sqls, stmt)
+		}
+	}
+	return hints, sqls
+}
+
+func isHint(stmt *parser.SQLFlowStmt, dialect string) bool {
+	if !stmt.IsExtendedSyntax() {
+		if dialect == "alisa" {
+			if strings.HasPrefix(strings.ToLower(stmt.Original), "set ") {
```
wangkuiyi: Should we trim leading whitespaces and tabs here before checking the existence of `set`?

pkg/sql/executor_ir.go
```diff
@@ -174,6 +176,43 @@ func runSingleSQLFlowStatement(wr *pipe.Writer, sql *parser.SQLFlowStmt, db *dat
...
+		} else {
+			sqls = append(sqls, stmt)
+		}
+	}
+	return hints, sqls
+}
+
+func isHint(stmt *parser.SQLFlowStmt, dialect string) bool {
+	if !stmt.IsExtendedSyntax() {
+		if dialect == "alisa" {
```
wangkuiyi, line 207: It might be better to use `switch` here and log a warning if `dialect` is not `"alias"`.

pkg/sql/executor_ir_test.go
weiguoz: Done.

pkg/sql/executor_ir.go
weiguoz: Sure.

typhoonzero (approved):

PR #2371, opened 2020-05-28
Summarize the discussion of TO RUN design

wangkuiyi:


doc/design/to_run.md
```diff
@@ -0,0 +1,165 @@
...
+Docker image.
+
+```sql
+SELECT * FROM source_table ORDER BY creation_date
+TO RUN 
+  a_data_scientist/extract_ts_features:1.0
+CMD 
+  "--verbose",
+  "--window_width=120"
+  "output_table_name";
```
brightcoder01: From the discussion, the `output_table_name` is in `INTO` clause.

```SQL
SELECT * FROM source_table ORDER BY creation_date
TO RUN 
  a_data_scientist/extract_ts_features:1.0
CMD 
  "--verbose",
  "--window_width=120"
INTO output_table_1, output_table_2
```

doc/design/to_run.md
```diff
@@ -0,0 +1,165 @@
...
+
+The SQLFlow server cannot pass in all context information in a single
+environment variable, which has a limit of value size.  Instead, it sets
+environment variables prefixed with `SQLFLOW_`.
+
+- `SQLFLOW_DB`: the type of DBMS.
+- [To-be-complete]
+
+
+## Run a Python Function
```
brightcoder01, line 166: The user will provide a python program instead of python function.
The python program contains a `main` function. The main function will parse the command line parameters and execute the table I/O, data computing process.

doc/design/to_run.md
wangkuiyi: You are right. The names after INTO are supposed to be passed to the entrypoint program via environment variables.

doc/design/to_run.md
wangkuiyi: It a Python function, but not a Python program, that the user should provide. The entrypoint programs I added to this section can `eval` the function call expression.  This example entrypoint program cannot handle table I/O and data computing. But this example shows that users, if they want, could provide more powerful entrypoint programs to handle anything, including starting Dask/Mars jobs.

brightcoder01 (approved):

doc/design/to_run.md
```diff
@@ -0,0 +1,222 @@
...
+row.
+
+This requirement intrigues us to support a general way to call arbitrary
+programs from SQL, hence the idea of the `TO RUN` clause.
+
+
+## Syntax Extension
+
+We want the syntax extension support to run any program written in any langauge,
+and we also want it easy to run Python functions in a way that best fit the
```
brightcoder01: best fit => best fits

doc/design/to_run.md
```diff
@@ -0,0 +1,222 @@
...
+
+### The SELECT Prefix
+
+As the `TO RUN` clause can run any program in the Docker image, it is not
+necessary to have the SELECT statement as a prefix.  Instead, we can change the
+above syntax design to make the SELECT prefix a command-line option to the
+entrypoint program, which then calls the DBMS API to execute it.
+
+```sql
+TO RUN
```
brightcoder01, line 126: How about disabling the syntax of calling `TO RUN` directly without the SELECT prefix. It's not so `SQL`. And SQL users may be confused with this statement.

wangkuiyi: Agreed. That's my typo. Fixing now.

brightcoder01 (approved):

PR #2392, opened 2020-06-01
Remove unused pkg dependence for parser-hive

lhw362950217:
To accelerate the build of java parser building, we remove some dependencies from parser-hive.


Yancey0623 (approved):
LGTM, please add some comments to explain why to exclude these dependencies.

PR #2395, opened 2020-06-01
Change sklearn version according to models revision

sneaxiy:
Related to https://github.com/sql-machine-learning/models/pull/69 . Make sklearn version be 0.21.0 to support sklearn2pmml.


docker/dev/build.sh
```diff
@@ -73,7 +73,7 @@ fi
 cd models
 git checkout master # The residual local repo might not be on a branch.
 git pull  # Pull the most recent commits, which might include the one we want.
-git checkout 07a75db
+git checkout 26ba78f7dbe9f55a53193ca68a47a528fe5e6977
```
wangkuiyi: How about we tag this Git commit and use the tag name here? We can give it any reasonable tag name, for example, `v0.0.1`.

wangkuiyi (approved):
LGTM!

I added the tag v0.0.1 to the models repo and refer it from here as `v0.0.1`.

PR #2396, opened 2020-06-01
Model zoo server receive client uploaded trained model

typhoonzero:


pkg/modelzooserver/modelzooserver.go
```diff
@@ -275,54 +276,92 @@ func (s *modelZooServer) DropModelDef(ctx context.Context, req *pb.ModelDefReque
...
+	sqlf = nil
+
+	// Create database sqlflow_public_models to store public trained models.
+	if _, err := s.DB.Exec("CREATE DATABASE IF NOT EXISTS sqlflow_public_models;"); err != nil {
+		return err
+	}
+
+	for { // read stream request
+		// NOTE: other fields in req must be the same in every stream request.
+		streamReq, err := stream.Recv()
```
lhw362950217, line 295: variable `streamReq` seems unnecessary, can we just use `req, err = stream.Recv`?

typhoonzero: Well, it's needed as a temporary variable. If `err == io.EOF`, the `streamReq` is actually `nil`, so we need to get the last  none nil object as `req`.

lhw362950217 (approved):
LGTM

PR #2402, opened 2020-06-02
Move feature column codegen to common path for XGBoost models support

sneaxiy:
Mainly move `generateFeatureColumnCode` to a new file `codegen_feature_column.go`.

### DIFF
--- a/pkg/sql/codegen/feature/codegen_feature_column.go
+++ b/pkg/sql/codegen/feature/codegen_feature_column.go
@@ -118,7 +118,7 @@ func GenerateFeatureColumnCode(fc ir.FeatureColumn, module string) (string, erro
 			module, sourceCode, c.Dimension, c.Combiner), nil
 	case *ir.IndicatorColumn:
 		if isXGBoostModule(module) {
-			return "", fmt.Errorf("INDICATOR is only supported in XGBoost models")
+			return "", fmt.Errorf("INDICATOR is not supported in XGBoost models")
 		}
 
 		sourceCode, err := GenerateFeatureColumnCode(c.CategoryColumn, module)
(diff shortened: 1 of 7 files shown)


pkg/step/feature/derivation.go
```diff
@@ -133,9 +133,9 @@ func newRowValue(columnTypeList []*sql.ColumnType) ([]interface{}, error) {
 	for idx, ct := range columnTypeList {
 		typeName := ct.DatabaseTypeName()
 		switch unifyDatabaseTypeName(typeName) {
-		case "VARCHAR", "TEXT", "STRING":
+		case "CHAR", "VARCHAR", "TEXT", "STRING":
 			rowData[idx] = new(string)
-		case "INT":
+		case "INT", "TINYINT":
```
typhoonzero, line 138: Will there be some test cases for these data types?

typhoonzero (dismissed):
LGTM generally

pkg/sql/codegen/feature/codegen_feature_column.go
```diff
@@ -0,0 +1,133 @@
+// Copyright 2020 The SQLFlow Authors. All rights reserved.
```
typhoonzero: Should we move package `feature` under `pkg/step/feature/` or keep it here? Seems we already have a package named `feature` there.

pkg/step/feature/derivation.go
sneaxiy: Done.

pkg/sql/codegen/feature/codegen_feature_column.go
sneaxiy: Thanks for recommendation. Moved to `codegen`.

typhoonzero (approved):
LGTM++

PR #2405, opened 2020-06-02
End2end test using a modelzoo model when predicting

typhoonzero:
TODO:

- support uploading, using model zoo models on PAI


pkg/model/model.go
```diff
@@ -90,10 +93,77 @@ func Load(modelURI, dst string, db *database.DB) (*Model, error) {
...
+			return nil, err
+		}
+	}
+
+	m := &Model{}
+	if e := gob.NewDecoder(&buf).Decode(m); e != nil {
+		return nil, fmt.Errorf("gob-decoding train select failed: %v", e)
+	}
+
+	if dst != "" { // empty in invalid param for tar -C
```
lhw362950217: in -> is ?

lhw362950217 (dismissed):
LGTM

pkg/model/model.go
```diff
@@ -90,10 +93,77 @@ func Load(modelURI, dst string, db *database.DB) (*Model, error) {
...
 		}
+	} else if strings.Contains(modelURI, "/") {
+		// general model zoo urls like some-domain.com:port/model_name:tag
+		// download traind model and extract to dst
+		return loadModelFromModelZoo(modelURI, dst)
 	}
 	return loadDB(db, modelURI, dst)
 }
 
+func loadModelFromModelZoo(modelURI, dst string) (*Model, error) {
```
wangkuiyi: loadModelFromZoo?

pkg/model/model.go
```diff
@@ -90,10 +93,77 @@ func Load(modelURI, dst string, db *database.DB) (*Model, error) {
...
+		modelName = modelNameAndTag
+		modelTag = "" // default use empty tag
+	} else if len(parts) == 2 {
+		modelName = parts[0]
+		modelTag = parts[1]
+	} else {
+		return nil, fmt.Errorf("model name after USING must be like [model-zoo.com:port/]model_name[:tag]")
+	}
+	// TODO(typhoonzero): use SSL connection
+	conn, err := grpc.Dial(modelZooServerAddr, grpc.WithInsecure())
```
wangkuiyi, line 107: Should the model zoo service a gRPC (backend) service, or an HTTP ( frontend) service?

pkg/model/model.go
```diff
@@ -90,10 +93,77 @@ func Load(modelURI, dst string, db *database.DB) (*Model, error) {
...
+	modelZooClient := pb.NewModelZooServerClient(conn)
+	stream, err := modelZooClient.DownloadModel(context.Background(), &pb.ReleaseModelRequest{
+		Name: modelName,
+		Tag:  modelTag,
+	})
+	if err != nil {
+		return nil, err
+	}
+	var buf bytes.Buffer
+	for {
```
wangkuiyi, line 121: This loop should be separated into a function definition.

pkg/model/model.go
```diff
@@ -90,10 +93,77 @@ func Load(modelURI, dst string, db *database.DB) (*Model, error) {
...
+		}
+	}
+
+	m := &Model{}
+	if e := gob.NewDecoder(&buf).Decode(m); e != nil {
+		return nil, fmt.Errorf("gob-decoding train select failed: %v", e)
+	}
+
+	if dst != "" { // empty is invalid param for tar -C
+		cmd := exec.Command("tar", "xzf", "-", "-C", dst)
```
wangkuiyi: These lines should be a function untar?

pkg/model/model.go
typhoonzero: model zoo server is a gRPC server to provide core functions, and we can use https://github.com/grpc/grpc-web to provide a front end web pages for users to view available models in the model zoo

wangkuiyi: Should model.go connect to an HTTP(s) proxy of the model zoo gRPC service, other than connecting directly to the gRPC service?  My purpose is that the use of model zoo service should be very flexible.  For example, an on-premise deployment of the SQLFlow service might (optionally) use a model zoo service deployment on public clouds.

wangkuiyi (approved):

pkg/model/model.go
typhoonzero: For on-premise deployments, we can deploy some TCP reverse proxies to proxy internal traffic out to a public model zoo server. Or we can configure the firewall to allow traffic to the specific domain of the public model zoo server. And if we make the model zoo an HTTP service, this type of work is still needed if there's traffic controls in the on-premise environment.

PR #2411, opened 2020-06-03
Support pmml saving of XGBoost models

sneaxiy:
Save XGBoost models as PMML file for future online prediction.

### DIFF
--- a/python/sqlflow_submitter/xgboost/train.py
+++ b/python/sqlflow_submitter/xgboost/train.py
@@ -166,21 +166,72 @@ def train(datasource,
         print("Evaluation result: %s" % re)
 
     if rank == 0:
-        # TODO(sneaxiy): save pmml as well
-        model_name = "my_model"
-        bst.save_model(model_name)
+        filename = "my_model"
+        save_local_file(bst, model_params, filename)
 
         if is_pai and len(oss_model_dir) > 0:
-            save_model(oss_model_dir, model_name, model_params, train_params,
+            save_model(oss_model_dir, filename, model_params, train_params,
                        feature_metas, feature_column_names, label_meta,
                        feature_column_code)
 
 
-def save_model(model_dir, model_name, model_params, train_params,
-               feature_metas, feature_column_names, label_meta,
-               feature_column_code):
-    model.save_file(model_dir, model_name)
-    # TODO(sneaxiy): save pmml as well
+def save_local_file(booster, model_params, filename):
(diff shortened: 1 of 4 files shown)


python/sqlflow_submitter/xgboost/train.py
```diff
@@ -166,21 +166,72 @@ def train(datasource,
...
                        feature_metas, feature_column_names, label_meta,
                        feature_column_code)
 
 
-def save_model(model_dir, model_name, model_params, train_params,
-               feature_metas, feature_column_names, label_meta,
-               feature_column_code):
-    model.save_file(model_dir, model_name)
-    # TODO(sneaxiy): save pmml as well
+def save_local_file(booster, model_params, filename):
```
typhoonzero: `save_model_local`?

typhoonzero (dismissed):

python/sqlflow_submitter/xgboost/train.py
sneaxiy: Done. Changed to `save_model_to_local_file` .

python/sqlflow_submitter/xgboost/train.py
```diff
@@ -166,21 +166,72 @@ def train(datasource,
...
+        model.classes_ = model._le.classes_
+
+        meta["_le"] = {"classes_": model.classes_.tolist()}
+        meta["classes_"] = model.classes_.tolist()
+    elif objective.startswith("reg:"):
+        model = xgb.XGBRegressor()
+    elif objective.startswith("rank:"):
+        model = xgb.XGBRanker()
+    else:
+        raise ValueError(
```
weiguoz, line 212: It seems that we lost some universality(general models) in order to support PMML.

sneaxiy: Yes, because sklearn-style API of XGBoost only supports `XGBClassifier`, `XGBRegressor` and `XGBRanker`. How about warning or error @weiguoz ?

weiguoz (approved):
LGTM ++
I'm pretty sure it's a useful feature.
Thank you @sneaxiy .

python/sqlflow_submitter/xgboost/train.py
weiguoz: > How about warning or error 

I prefer using `error` here.

Meanwhile, I would suggest adding notes to user documentation, such as [/doc/model_parameter.md](https://github.com/sql-machine-learning/sqlflow/blob/develop/doc/model_parameter.md). Let's do this in another PR later.

PR #2416, opened 2020-06-04
Add INDICATOR support in XGBoost models

sneaxiy:
ATT.

### DIFF
--- a/cmd/sqlflowserver/main_test.go
+++ b/cmd/sqlflowserver/main_test.go
@@ -1616,7 +1616,8 @@ func getUniqueID() int {
 	return uniqueID
 }
 
-func caseXGBoostFeatureColumnImpl(t *testing.T, table string, label string, selectColumns string, columnClauses string, nclasses int, nworkers int, isPai bool) {
+func caseXGBoostFeatureColumnImpl(t *testing.T, table string, label string, selectColumns string, columnClauses string, nclasses int, nworkers int, isPai bool,
+	skipExplain bool) {
 	tableSplits := strings.SplitN(table, ".", 2)
 	dbPrefix := ""
 	if len(tableSplits) == 2 {
@@ -1670,12 +1671,14 @@ func caseXGBoostFeatureColumnImpl(t *testing.T, table string, label string, sele
 		executeSQLFunc(evaluateSQL)
 	}
 
-	paiExplainExtra := ""
-	if isPai {
-		paiExplainExtra = fmt.Sprintf(`, label_col="%s" INTO %sxgb_fc_test_explain_table_%d`, label, dbPrefix, uniqueID)
+	if !skipExplain {
+		paiExplainExtra := ""
+		if isPai {
+			paiExplainExtra = fmt.Sprintf(`, label_col="%s" INTO %sxgb_fc_test_explain_table_%d`, label, dbPrefix, uniqueID)
+		}
+		explainSQL := fmt.Sprintf(`SELECT %s FROM %s TO EXPLAIN %s WITH summary.plot_type=bar %s;`, selectColumns, table, modelName, paiExplainExtra)
+		executeSQLFunc(explainSQL)
 	}
-	explainSQL := fmt.Sprintf(`SELECT %s FROM %s TO EXPLAIN %s WITH summary.plot_type=bar %s;`, selectColumns, table, modelName, paiExplainExtra)
-	executeSQLFunc(explainSQL)
 
 	if !isPai { // PAI does not support SHOW TRAIN, because the model is not saved into database
 		showTrainSQL := fmt.Sprintf(`SHOW TRAIN %s;`, modelName)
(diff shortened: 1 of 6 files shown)


pkg/sql/codegen/codegen_feature_column.go
```diff
@@ -39,9 +41,14 @@ func isXGBoostModule(module string) bool {
 	return strings.HasPrefix(module, "xgboost")
 }
 
-// IntArrayToJSONString converts int array to json string
+// IntArrayToJSONString converts int array to JSON string
```
wangkuiyi: It doesn't make sense to have an exported function to export []int into JSON.  Such a function should be don't with json.Marshal.  If we have such a function, it must be a handy tool used locally, like `encodeIntSliceOrDie`.

cmd/sqlflowserver/main_test.go
```diff
@@ -1616,7 +1616,8 @@ func getUniqueID() int {
 	return uniqueID
 }
 
-func caseXGBoostFeatureColumnImpl(t *testing.T, table string, label string, selectColumns string, columnClauses string, nclasses int, nworkers int, isPai bool) {
+func caseXGBoostFeatureColumnImpl(t *testing.T, table string, label string, selectColumns string, columnClauses string, nclasses int, nworkers int, isPai bool,
+	skipExplain bool) {
```
wangkuiyi, line 1622: Do we need the skipExplain flag here? I mean this is a test case and we want it to test the subject thoroughly -- why should we skip the testing of explaining?

sneaxiy: Because XGBoost explanation uses `pandas.DataFrame` . See [here](https://github.com/sql-machine-learning/sqlflow/pull/2416/files#diff-9c0c52328de10b86c937f3425060d52dR53). It does not support `INDICATOR` now since `INDICATOR` would make a scalar be a vector. If we want to support `INDICATOR` in XGBoost `INDICATOR`, we should separate each element in the one hot vector to several columns of `pandas.DataFrame`. I would fix it in the next PR. Marked.

pkg/sql/codegen/codegen_feature_column.go
sneaxiy: This function is also used in [tensorflow/codegen.go](https://github.com/sql-machine-learning/sqlflow/blob/develop/pkg/sql/codegen/tensorflow/codegen.go#L84) . As you suggested, changed it to `MarshalOrDie`.

python/sqlflow_submitter/xgboost/feature_column.py
```diff
@@ -170,16 +202,20 @@ def _set_field_names(self, field_names):
...
+        return self.hash_bucket_size
+
     def __call__(self, inputs):
         def transform_fn(slot_value):
             if isinstance(slot_value, np.ndarray):
-                output = np.ndarray(slot_value.shape)
+                output = np.ndarray(slot_value.shape, dtype=np.int64)
                 for i in six.moves.range(slot_value.size):
-                    output[i] = hashing(slot_value[i])
+                    output.put(
```
brightcoder01: Is `output[i] = hashing(slot_value.take(i)) % self.hash_bucket_size` cleaner?

pkg/sql/codegen/codegen_feature_column.go
```diff
@@ -39,9 +41,14 @@ func isXGBoostModule(module string) bool {
 	return strings.HasPrefix(module, "xgboost")
 }
 
-// IntArrayToJSONString converts int array to json string
-func IntArrayToJSONString(intArray []int) string {
-	return strings.Join(strings.Split(fmt.Sprint(intArray), " "), ",")
+// MarshalOrDie converts any data to JSON string.
+// Exit the process if there is any error.
+func MarshalOrDie(in interface{}) string {
```
brightcoder01: MarshalToJsonOrDie

Marshal is too general, we can marshal an object to JSON/XML/Bytes/...

python/sqlflow_submitter/xgboost/feature_column.py
```diff
@@ -190,6 +226,46 @@ def categorical_column_with_hash_bucket(key, hash_bucket_size, dtype='string'):
...
+
+    def __call__(self, inputs):
+        slot = self.categorical_column(inputs)
+        assert len(
+            slot
+        ) == 1, "indicator_column does not accept sparse categorical feature"
+
+        def transform_fn(slot_value):
+            num_classes = self.categorical_column.num_classes()
+            if isinstance(slot_value, np.ndarray):
```
brightcoder01, line 251: Can we add comments that this `np.array` is only 2 dimensional?

python/sqlflow_submitter/xgboost/feature_column_test.py
```diff
@@ -0,0 +1,187 @@
...
+
+        self.assertEqual(len(inputs), len(expected_outputs))
+
+        column = fc.ComposedColumnTransformer(column_names, column)
+        outputs = column((inputs, ))[0]
+
+        self.assertEqual(len(inputs), len(outputs))
+
+        if isinstance(outputs[0], np.ndarray):
+            if not np.array_equal(outputs[0], expected_outputs[0]):
```
brightcoder01: Can we move this print into self.assertTrue as the message of assertion failure?

python/sqlflow_submitter/xgboost/feature_column_test.py
```diff
@@ -0,0 +1,187 @@
...
+                print(outputs[0], expected_outputs[0])
+
+            self.assertTrue(np.array_equal(outputs[0], expected_outputs[0]))
```
brightcoder01, line 54: Why do we only check the first element of outputs and expected_outputs?

python/sqlflow_submitter/xgboost/feature_column.py
sneaxiy: It may not work. The previous code is wrong. `output[i]` returns `np.ndarray` instead of a single  element.

python/sqlflow_submitter/xgboost/feature_column.py
sneaxiy: This `np.array` may be multi-dimensional. See [here](https://github.com/sql-machine-learning/sqlflow/pull/2416/files/456b7e3d91312b1ab0b8cb74bf97886ed2a3acad#diff-2f0634b1ebc2f4375900e5a40aaad875R255).

python/sqlflow_submitter/xgboost/feature_column_test.py
sneaxiy: Sorry, this is the debug info when developing. I would remove it.

python/sqlflow_submitter/xgboost/feature_column_test.py
sneaxiy: This is checking all elements. The `output`'s format is `tuple(output_array)` .

pkg/sql/codegen/codegen_feature_column.go
sneaxiy: Done.

python/sqlflow_submitter/xgboost/feature_column.py
brightcoder01: I recommend to use [numpy.vectorize](https://numpy.org/doc/stable/reference/generated/numpy.vectorize.html) to wrap the function which are applied to a numpy array elementwise.

python/sqlflow_submitter/xgboost/feature_column.py
brightcoder01: Is slot_value also multi-dimensional?

python/sqlflow_submitter/xgboost/feature_column.py
sneaxiy: Great! Done.

python/sqlflow_submitter/xgboost/feature_column.py
sneaxiy: Yes!

brightcoder01 (approved):

pkg/sql/codegen/codegen_feature_column.go
```diff
@@ -39,9 +41,14 @@ func isXGBoostModule(module string) bool {
...
 
-// IntArrayToJSONString converts int array to json string
-func IntArrayToJSONString(intArray []int) string {
-	return strings.Join(strings.Split(fmt.Sprint(intArray), " "), ",")
+// MarshalToJSONOrDie converts any data to JSON string.
+// Exit the process if there is any error.
+func MarshalToJSONOrDie(in interface{}) string {
+	bytes, err := json.Marshal(in)
+	if err != nil {
+		log.Fatal(err)
```
typhoonzero, line 49: Does this error end the sqlflowserver process? If so, this is not recommended since we don't need to stop the server process if there's an error occurred when processing one statement.

sneaxiy: Fix in https://github.com/sql-machine-learning/sqlflow/pull/2433 .

PR #2422, opened 2020-06-05
support incremental training in XGBoost

sneaxiy:
Extend `TO TRAIN` syntax to support incremental training. The SQL is like:

```sql
SELECT * FROM ...
TO TRAIN ...
WITH ...
LABEL ...
[USING pre_trained_model_name_to_load]
INTO new_trained_model_name_to_save;
```


pkg/sql/codegen/pai/codegen.go
```diff
@@ -122,20 +122,24 @@ func Train(ir *ir.TrainStmt, session *pb.Session, tarball, paramsFile, modelName
 			return
 		}
 	} else if strings.HasPrefix(strings.ToLower(ir.Estimator), "xgboost") {
-		ossURI := OSSModelURL(ossModelPath)
-		if code, e = xgboost.DistTrain(ir, session, cc.Worker.Count, ossURI); e != nil {
+		ossURIToSave := OSSModelURL(ossModelPathToSave)
+		ossURIToLoadPreTrainedModel := ""
```
typhoonzero: Simplify to `ossURIToLoad`?

typhoonzero: Check if `ossModelPathToSave == ossModelPathToLoadPreTrained`?

sneaxiy: > Simplify to ossURIToLoad?

Done.

> Check if ossModelPathToSave == ossModelPathToLoadPreTrained?

There is no need to check. We can support both `save_path == load_path` and `save_path != load_path`.

pkg/parser/extended_syntax_parser.y
```diff
@@ -76,11 +76,12 @@ type StandardSelect struct {
...
-	Estimator  string
-	TrainAttrs Attributes
-	Columns    columnClause
-	Label      string
-	Save       string
+	Estimator       string
+	TrainAttrs      Attributes
+	Columns         columnClause
+	Label           string
+	PreTrainedModel string
```
typhoonzero: parser and IR represents the parsed SQL statement information, `PreTrainedModel` is more likely to be how this `USING` field will be used, not representing what the parse the field is. Maybe change it to `TrainUsing` ?

typhoonzero (dismissed):
LGTM with one naming issue.

pkg/sql/codegen/pai/codegen.go
typhoonzero: I see.

pkg/parser/extended_syntax_parser.y
sneaxiy: Sounds good. Done.

typhoonzero (approved):
LGTM++

PR #2433, opened 2020-06-08
Change MarshalToJSONOrDie to MarshalToJSON

sneaxiy:
To avoid sqlflowserver die when error occurs.

### DIFF
--- a/pkg/sql/codegen/codegen_feature_column.go
+++ b/pkg/sql/codegen/codegen_feature_column.go
@@ -16,7 +16,6 @@ package codegen
 import (
 	"encoding/json"
 	"fmt"
-	"log"
 	"sqlflow.org/sqlflow/pkg/ir"
 	"strings"
 )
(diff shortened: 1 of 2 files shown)


pkg/sql/codegen/codegen_feature_column.go
```diff
@@ -41,34 +40,37 @@ func isXGBoostModule(module string) bool {
 	return strings.HasPrefix(module, "xgboost")
 }
 
-// MarshalToJSONOrDie converts any data to JSON string.
-// Exit the process if there is any error.
-func MarshalToJSONOrDie(in interface{}) string {
+// MarshalToJSONString converts any data to JSON string.
+func MarshalToJSONString(in interface{}) (string, error) {
```
wangkuiyi, line 44: I don't think this function is necessary. We can use `json.Marshal`.

pkg/sql/codegen/codegen_feature_column.go
```diff
@@ -41,34 +40,37 @@ func isXGBoostModule(module string) bool {
...
-	}
-	return string(bytes)
+	return string(bytes), err
 }
 
 // GenerateFeatureColumnCode generates feature column code for both TensorFlow and XGBoost models
 func GenerateFeatureColumnCode(fc ir.FeatureColumn, module string) (string, error) {
 	switch c := fc.(type) {
 	case *ir.NumericColumn:
+		shapeStr, err := MarshalToJSONString(c.FieldDesc.Shape)
```
wangkuiyi, line 53:
```go
s, e ;= json.Marshal(c.FieldDesc.Shape)
return fmt.Sprintf("%s", s)  // %s automatically converts []byte into string.
```

pkg/sql/codegen/codegen_feature_column.go
sneaxiy: This function is also used [here](https://github.com/sql-machine-learning/sqlflow/blob/develop/pkg/sql/codegen/tensorflow/codegen.go#L336), which would be used in [Python codegen template](https://github.com/sql-machine-learning/sqlflow/blob/develop/pkg/sql/codegen/tensorflow/template_train.go#L79). If we use `json.Marshal` directly in `codegen_feature_column.go`, we have to write duplicate codes [here](https://github.com/sql-machine-learning/sqlflow/blob/develop/pkg/sql/codegen/tensorflow/codegen.go#L336).

wangkuiyi (approved):

pkg/sql/codegen/codegen_feature_column.go
sneaxiy: Same reason why I choose to write a new function as [here](https://github.com/sql-machine-learning/sqlflow/pull/2433#discussion_r437066025).

PR #2444, opened 2020-06-10
Optimize syntax support group by

typhoonzero:
To support syntax like:

```sql
SELECT src.plants, src.markets, src.distance, plants.capacity, markets.demand FROM transportation AS src
LEFT JOIN plants ON src.plants = plants.plants
LEFT JOIN markets ON src.markets = markets.markets
TO MINIMIZE SUM(distance * 90 / 1000)
CONSTRAINT SUM(markets) <= capacity GROUP BY plants,
           SUM(plants) >= demand GROUP BY markets
WITH variables="plants,markets",
     plants="Integers",
     markets="Integers"
[USING glpk]
INTO result_table;
```

ref: https://github.com/sql-machine-learning/sqlflow/blob/develop/doc/design/programming_solver.md

### DIFF
--- a/pkg/parser/extended_syntax_parser.y
+++ b/pkg/parser/extended_syntax_parser.y
@@ -20,6 +20,13 @@ type Expr struct {
 
 type ExprList []*Expr
 
+type ConstraintExpr struct {
+	expr *Expr
+	groupby string
+}
+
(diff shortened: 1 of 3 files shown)


sneaxiy (approved):
LGTM.

PR #2458, opened 2020-06-12
Polish model zoo server

typhoonzero:
Remove TODO comments and add input checks to avoid SQL injection.

### DIFF
--- a/pkg/modelzooserver/input_check.go
+++ b/pkg/modelzooserver/input_check.go
@@ -33,6 +33,18 @@ func checkName(name string) error {
 	return nil
 }
 
+// checkTag checks tag is valid. Tags should be consist of alphabet, numbers, dashes (-), underscrolls (_) and dots (.).
+func checkTag(name string) error {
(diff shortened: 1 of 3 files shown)


pkg/modelzooserver/input_check.go
```diff
@@ -23,7 +23,7 @@ func checkName(name string) error {
 	if len(name) < 6 {
 		return fmt.Errorf("model name should have at least 6 characters")
 	}
-	match, err := regexp.MatchString(`^[a-zA-Z0-9_-]{6,256}$`, name)
+	match, err := regexp.MatchString(`^[a-zA-Z0-9_\-]{6,256}$`, name)
```
lhw362950217, line 26: when '-' is the first or the last char in '[]', we do not need to escape it

lhw362950217 (approved):
LGTM

pkg/modelzooserver/input_check.go
typhoonzero: Good point. Will refine in following pull requests.

PR #2466, opened 2020-06-15
Support native keras models

typhoonzero:


python/sqlflow_submitter/tensorflow/keras_example.py
```diff
@@ -22,6 +22,36 @@
 from sqlflow_submitter.tensorflow.predict import pred
 from sqlflow_submitter.tensorflow.train import train
 
+native_keras_code = """import tensorflow as tf
+
+class RawDNNClassifier(tf.keras.Model):
+    def __init__(self, hidden_units=[100,100], n_classes=3):
+        super(RawDNNClassifier, self).__init__()
+        self.feature_layer = None
```
sneaxiy: How about add some comments on why `self.feature_layer` may be not None, i.e., explain `WrappedKerasModel`?

python/sqlflow_submitter/tensorflow/train_predict_test.py
```diff
@@ -56,6 +56,14 @@ def test_keras(self):
                                  env=os.environ.copy(),
                                  check=True)
             self.assertEqual(ret.returncode, 0)
+
+            ret = subprocess.run([
+                "/usr/local/bin/python",
```
sneaxiy: How about use `sys.executable` instead of hard code path?

python/sqlflow_submitter/tensorflow/keras_example.py
typhoonzero: `native_keras_code` should be removed actually. we are using https://github.com/sql-machine-learning/models/blob/develop/sqlflow_models/native_keras.py this model. `feature_layer` should always be `None`.

sneaxiy (approved):
LGTM++

PR #2469, opened 2020-06-15
Disable JupyterNotebook's token validation for playground

lhw362950217:
Current single node playground mainly focus on single user, so,  it seems that we do not need to add a password for Jupyter Notebook.


Yancey0623 (approved):
LGMT

PR #2470, opened 2020-06-15
Put testing utility to a more common file

shendiaomo:


pkg/sql/executor_ir_test.go
```diff
@@ -115,6 +117,35 @@ USING sqlflow_models.my_xgb_regression_model;
 `
 )
 
+func goodStream(stream chan interface{}) (bool, string) {
```
typhoonzero: How about put to `pkg/sql/testing.go`, then other places like `main_test.go` can also reuse this function.

shendiaomo: Done.

shendiaomo: Done.

typhoonzero (approved):

PR #2476, opened 2020-06-16
Save model meta into 'model_meta.json'

### LINKED ISSUE #2085: Save model meta to OSS when training on PAI platform
Current training process on PAI will not save metadata on OSS, so we can't use the `SHOW TRAIN` statement.

Save metadata with the model like we do on other platforms

**Additional Notes**

lhw362950217:
Fixes #2085
This part just save model's metadata into `model_meta.json`, next PR will try to use it in Model Zoo cli.


typhoonzero (dismissed):
LGTM, actually we can load the saved metas when predicting so that we do not need to do feature derivation again.

python/sqlflow_submitter/xgboost/train.py
```diff
@@ -224,12 +232,13 @@ def save_model_to_local_file(booster, model_params, filename):
...
     meta["type"] = model_type
-    meta = json.dumps(meta)
+    metaJson = json.dumps(meta)
 
     # Meta data is needed for saving sklearn pipeline. See here:
     # https://github.com/dmlc/xgboost/blob/d19cec70f1b40ea1e1a35101ca22e46dd4e4eecd/python-package/xgboost/sklearn.py#L356
-    booster.set_attr(scikit_learn=meta)
+    booster.set_attr(scikit_learn=metaJson)
     booster.save_model(filename)
+    save_model_metadata("model_meta.json", meta)
```
sneaxiy, line 241: Do not need to save the sklearn `meta`. The sklearn `meta` here is the meta data needed for sklearn model saving. It would be written into model binary when calling `booster.save_model`. We should only save the meta info which SQLFlow needs.

lhw362950217: done

sneaxiy (dismissed):
LGTM.

wangkuiyi (dismissed):

sneaxiy (approved):
LGTM.

PR #2477, opened 2020-06-16
Migration to Python: encapsulating sqlflow_submitter

shendiaomo:
This is based on #2432 
1. `ir.proto`: `Program` and `Statement` map concept of SQLFlow to Python and Go
2. Package `python/sqlflow/column/` defines column functions for `tensorflow` and `xgboost`
3. `python/sqlflow/features.py` extracts table column statistics and maps functions in a `COLUMN` clause to Python function calls.
4. Package `python/sqlflow/platform/` defines `train`, `predict` and other functions that encapsulate `sqlflow_submitter` for each type of SQLFlow statement. The commit only supports the `default.py` for default deployment

### DIFF
--- a/pkg/proto/ir.proto
+++ b/pkg/proto/ir.proto
@@ -1,0 +1,67 @@
+syntax = "proto3";
+
+package proto;
+
+message Statement {
+  // `select` is the query for fetching data. For example,
+  // "SELECT * FROM iris.train;"
+  string select = 2;
+  // `validation_select` is the query for fetching the validation data.
+  // For example, "SELECT * FROM iris.val;".
+  string validation_select = 3;
+  // `model_image` is the name of the model's Docker image, for example, in the
+  // statement "TO TRAIN a_data_scientist/regressors:v0.2/MyDNNRegressor", the
+  // name "a_data_scientist/regressors:v0.2" is a Docker image.
+  string model_image = 4;
+  // `estimator` specifies the estimator type. For example, after parsing
+  // "SELECT ... TRAIN DNNClassifier WITH ...", `estimator` will be
+  // "DNNClassifier".
+  string estimator = 5;
+  // `attributes` is a map of parsed attribute in the WITH Clause. For example,
+  // after parsing "SELECT ... TO TRAIN ... with train.epoch=1000,
+  // model.hidden_units = [10, 10]", the `attributes` will be
+  // {"train.epoch": "1000", "model.hidden_units": "[10, 10]"}
+  map<string, string> attributes = 6;
+  string label = 7;
+  message Columns {
+  // The COLUMN clause will be split into `column`. For example, in the COLUMN
+  // clause "COLUMN NUMERIC(sepal_length), INDICATOR(ID)" will be saved in
+  // `columns` as ["NUMERIC(sepal_length)", "INDICATOR(ID)"]
+    repeated string columns = 1;
+  }
+  // `columns` contain a map of string to `Columns` for multiple COLUMN clauses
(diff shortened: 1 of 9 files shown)


pkg/proto/ir.proto
```diff
@@ -0,0 +1,67 @@
+syntax = "proto3";
```
wangkuiyi, line 1: If the codegen is going to template-fill the IR into generated code, why do we need to redefine the IR in protobuf?

python/sqlflow/column/tensorflow.py
```diff
@@ -0,0 +1,62 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import tensorflow as tf
+from tensorflow import feature_column as tfc
+
+
+def numeric(field, dim=0):
```
wangkuiyi, line 18: I feel very difficult to understand the Python code here. Python doesn't allow programmers to declare parameter types and return types. I have no idea what `field` is without an exhaustively detailed comment.

There is nothing like reading Python code -- it is deciphering.

If we are going to have a large Python codebase, we must follow the code quality of Google Tangent.  For example: https://github.com/google/tangent/blob/master/tangent/compile.py.  If we are not confident with ourselves on this level of coding quality, I would say the less Python code the better.

TensorFlow is a notorious example that a bunch of normal-level Python programmers can easily ruin the code quality set up by a few elite Google engineers and make it impossible to get back to the right track without abandoning the whole thing and rewriting the whole thing.

python/sqlflow/features.py
```diff
@@ -0,0 +1,127 @@
...
+def get_feature_metas(conn, select, label=None):
+    cursor = conn.cursor()
+    cursor.execute(select)
+    feature_metas = {}
+    feature_column_names = []
+    rows = list(cursor.fetchmany(1024))
+    for i, column in enumerate(cursor.description):
+        name, typ = column[0], column[1]
+        sep, shape, maxid, vocab = "", [1], 0, set()
+        if typ in (4, 5):
```
wangkuiyi, line 29: What are these magic numbers, 4 and 5?

python/sqlflow/features.py
```diff
@@ -0,0 +1,127 @@
...
+            dtype = "float32"
+        elif typ in (1, 3, 8):
+            dtype = "int64"
+            vocab.update(r[i] for r in rows)
+        elif typ in (252, 253):
```
wangkuiyi, line 34: What are these magic numbers?

python/sqlflow/column/tensorflow.py
```diff
@@ -0,0 +1,62 @@
...
+
+def seq_category_id(field, size, sep=''):
+    field = sparse(field, size, sep) if not field["delimiter"] else field
+    ret = tfc.sequence_categorical_column_with_identity(
+        field["feature_name"], size)
+    ret.key = field["feature_name"]
+    return ret
+
+
+EVAL_GLOBALS = {'comma': ',', 'mean': 'mean', 'sum': 'sum', 'sqrtn': 'sqrtn'}
```
wangkuiyi, line 62: It is very confusing that 

1. `EVAL_GLOBALS` is defined twice -- here and in `xgboost.py`, and
1. this only place that uses `EVAL_GLOBALS` is `features.py`.

wangkuiyi (commented):
I feel basically impossible to understand the Python code. I cannot get the idea it silhouettes.

Indeed, this PR makes me even more afraid of having a large Python codebase in SQLFlow than ever -- Python code is evilly hard to understand and maintain.

PR #2478, opened 2020-06-16
[Programming] Support optflow task submission in PAI

sneaxiy:
ATT.


pkg/ir/ir.go
```diff
@@ -230,18 +231,38 @@ type OptimizeExpr struct {
...
+	// Direction, "maximize" or "minimize"
+	Direction string
 	// Constraints
 	Constraints []*OptimizeExpr
+	// Solver
+	Solver string
 	// ResultTable is the table name to store results.
 	ResultTable string
-	// When SQLFLOW_submitter == "pai", tmp tables will be created for solving tasks
-	TmpTrainTable string
```
typhoonzero, line 246: `TmpTrainTable` maybe needed when the `SELECT` statement contains `JOIN`, then we need to create a tmp table using `CREATE TABLE AS SELECT ...`

pkg/sql/pai_submitter.go
```diff
@@ -395,6 +396,55 @@ func (s *paiSubmitter) ExecuteEvaluate(cl *ir.EvaluateStmt) error {
 	return e
 }
 
+// TODO(sneaxiy): need to add some tests to this function, but it requires
+// optflow installed in docker image
+func (s *paiSubmitter) ExecuteOptimize(cl *ir.OptimizeStmt) error {
```
typhoonzero, line 400: `ExecuteOptimize` actually is not submitting jobs to PAI. Can put this function under `submitter.go`, call the function in here so that when `SQLFLOW_submitter` is configured to `pai`, we can also submit the job.

python/sqlflow_submitter/optimize/optimize.py
```diff
@@ -0,0 +1,357 @@
...
+                           minimize)
+
+# TODO(sneaxiy): support more aggregation functions if needed
+AGGREGATION_FUNCTIONS = ['sum']
+
+# FIXME(sneaxiy): do not know why pyomo requires that DATA_FRAME must be a global variable
+DATA_FRAME = None
+
+
+def split_to_tokens(expression):
```
typhoonzero: The parsed result is already tokens, so how about changing the IR `expression` from `string` to `[]string` to store tokens?

sneaxiy: Done.

pkg/sql/pai_submitter.go
sneaxiy: Done.

pkg/ir/ir.go
sneaxiy: `TmpTrainTable` is not needed. It only occurs in `ExecuteOptimize` method.

typhoonzero (approved):
LGTM++

PR #2482, opened 2020-06-17
[Programming]Add optimize attribute checker

sneaxiy:
Support attributes `worker.*`, `data.*` and `solver.*`.


typhoonzero (dismissed):

pkg/sql/codegen/optimize/template_optimize.go
```diff
@@ -75,27 +83,36 @@ def submit():
...
+                       batch_size=batch_size)
     
     input_conf = InputConf(df1=df1)
 
     optflow_version = os.environ.get("SQLFLOW_OPTFLOW_VERSION")
     if not optflow_version:
         raise ValueError("Environment variable SQLFLOW_OPTFLOW_VERSION must be set")
     	
-    if "{{.IsPAI}}":
+    if is_pai:
```
typhoonzero: check worker count > 1. `is_pai` should not be used in optimize

sneaxiy: As offline discussion, `is_pai` is removed. We can only support kubemaker mode.

pkg/sql/codegen/optimize/codegen.go
```diff
@@ -15,21 +15,84 @@ package optimize
...
+	"encoding/json"
 	"fmt"
 	"sqlflow.org/sqlflow/pkg/ir"
 	pb "sqlflow.org/sqlflow/pkg/proto"
+	"sqlflow.org/sqlflow/pkg/sql/codegen/attribute"
 	"strings"
 	"text/template"
 )
 
+func checkIsPositiveInteger(i interface{}, name string) error {
```
wangkuiyi: Move `checkIsPositiveInteger` into package `attribute.IsPositiveInteger`?

pkg/sql/codegen/optimize/codegen.go
```diff
@@ -15,21 +15,84 @@ package optimize
...
+	if i.(int) <= 0 {
```
wangkuiyi:
```go
if v, ok := i.(int); !ok || v <= 0 {
```
 
Check https://tour.golang.org/methods/15 for more examples.

pkg/sql/codegen/optimize/codegen.go
```diff
@@ -15,21 +15,84 @@ package optimize
...
+		return checkIsPositiveInteger(i, "worker.num")
+	}},
+	"worker.core": {attribute.Int, 8, "Worker core number", func(i interface{}) error {
+		return checkIsPositiveInteger(i, "worker.core")
+	}},
+	"worker.memory": {attribute.Int, 4096, "Worker memory", func(i interface{}) error {
+		return checkIsPositiveInteger(i, "worker.memory")
+	}},
+	"solver.*": {attribute.Unknown, nil, "Solver options", nil},
+}
```
wangkuiyi: The Go idiomatic way is

```go
var attributeDict = attribute.NewDict().
   Bool("data.enable_slice", false, "the explanation", nil).
   Int32("worker.core", 8, "number of cores per worker", attribute.PositiveInt32)
```

Then, we don't need to abuse `interface {}` in `attribute.PostiveInterger`

```go
// In package attribute.assert.
func PositiveInt32(x int, name string) error {
    if x <= 0 {
        return fmt.Errorf("Expecting %s being positive integer, but got %d", name, x)
    }
    return nil
}
```

pkg/sql/codegen/optimize/codegen.go
```diff
@@ -15,21 +15,84 @@ package optimize
...
+
+const (
+	dataAttrPrefix   = "data."
+	solverAttrPrefix = "solver."
+	workerAttrPrefix = "worker."
```
wangkuiyi: Either make sure that the construction of the `attribute.Dictionary` uses these prefixes too, or, if they are only used in `func GenerateOptflowOptimzieCode`, move them into that function definition.

pkg/sql/codegen/optimize/template_optimize.go
```diff
@@ -57,16 +57,20 @@ class CustomOptFlowRunner(BaseOptFlowRunner):
 
 const optFlowSubmitText = `
 import os
+import json
```
wangkuiyi: It doesn't make sense to have a template this long -- please file a PR to minimize the template into only one function call which takes all template-filler parameters, before merging this one.

wangkuiyi (changes requested):

pkg/sql/codegen/optimize/codegen.go
sneaxiy: Done.

pkg/sql/codegen/optimize/codegen.go
sneaxiy: Done.

pkg/sql/codegen/optimize/template_optimize.go
sneaxiy: Done. Move most Python code to `runner.py`.

wangkuiyi (approved):

PR #2484, opened 2020-06-17
Restructure the `codegen` and the `attribute` packages

shendiaomo:
After the restructure, each essential part of the compiler will be a top package:
```bash
+ pkg/
    |- parser/
    |- attribute/ # Semantic checking
    |- verifier/  # Semantic checking
    |- ir/  # Intermediate representation
    |- codegen/  # Code generator
```

The next step is to decouple `executor` and `codegen`: `ir` generates code, `executor` executes the code.

These commits are related to #2481


wangkuiyi:
@shendiaomo It makes sense to me that attribute value range checking is part of semantic analysis.  However, to make a decision to move `codegen/attribute` to `attribute`, we need to confirm that `codegen` is not the only caller of `attribute`.  Is this the case already? Or, are you going to make it the case?

wangkuiyi (approved):

PR #2486, opened 2020-06-18
Add typed API to pkg/attribute

wangkuiyi:
This PR addresses https://github.com/sql-machine-learning/playground/issues/42#issue-640787739

### DIFF
--- a/pkg/attribute/attribute.go
+++ b/pkg/attribute/attribute.go
@@ -55,6 +55,61 @@ type Description struct {
 	Checker func(i interface{}) error
 }
 
+// Int declares an attribute of int-typed in Dictionary d.
+func (d Dictionary) Int(name string, value int, doc string, checker func(int) error) Dictionary {
+	d[name] = &Description{
+		Type:    Int,
+		Default: value,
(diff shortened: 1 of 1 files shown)


tonyyang-svail (approved):
LGTM

pkg/attribute/attribute.go
```diff
@@ -55,6 +55,61 @@ type Description struct {
 	Checker func(i interface{}) error
 }
 
+// Int declares an attribute of int-typed in Dictionary d.
+func (d Dictionary) Int(name string, value int, doc string, checker func(int) error) Dictionary {
+	d[name] = &Description{
+		Type:    Int,
+		Default: value,
+		Doc:     doc,
+		Checker: func(x interface{}) error { return checker(x.(int)) },
```
wangkuiyi, line 64: Here I should make it

```go
func(x interface{}) error {
    v, ok := x(int)
    if ! ok {
         return fmt.Errorf("Attribute %s is of type int, but the value %v cannot be casted into int", name, x)
    }
    return checker(v)
}
```

PR #2487, opened 2020-06-18
Polish keras wrapped model init with feature columns

typhoonzero:
Polish keras wrapped model init with feature columns and fix running with Python 2 (Python 2 do not support `inspect.signature`)

### DIFF
--- a/python/sqlflow_submitter/tensorflow/diag.py
+++ b/python/sqlflow_submitter/tensorflow/diag.py
@@ -32,7 +32,6 @@ def load_pretrained_model_estimator(estimator,
         # The constructor of Estimator contains named parameter "warm_start_from"
         warm_start_from_key = "warm_start_from"
         if warm_start_from_key in estimator_spec.args:
-            model_params = copy.copy(model_params)
             warm_start_from = os.path.abspath(warm_start_from)
 
             if is_tf_estimator(estimator):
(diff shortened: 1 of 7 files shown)


python/sqlflow_submitter/tensorflow/diag.py
```diff
@@ -22,7 +22,9 @@ class SQLFlowDiagnostic(Exception):
     pass
 
 
-def check_and_load_estimator(estimator, model_params, warm_start_from=None):
+def load_pretrained_model_estimator(estimator,
+                                    model_params,
+                                    warm_start_from=None):
```
sneaxiy, line 27: Does not work to load pretrained model because `model_params` is not updated?

python/sqlflow_submitter/tensorflow/keras_with_feature_column_input.py
```diff
@@ -23,3 +27,28 @@ def __init__(self, keras_model, model_params, feature_columns):
...
+
+       NOTE: initalize estimator model can also use this function since estimators all have
+       "feature_column" argument.
+    """
+    argspec = inspect.getargspec(estimator)
+    has_feature_columns_arg = False
+    for k in argspec.args:
+        if k == "feature_columns":
+            has_feature_columns_arg = True
+            break
```
sneaxiy: Simplify to be `has_feature_columns_arg  = "feature_column" in argspec.args`?

sneaxiy (approved):
LGTM.

PR #2491, opened 2020-06-18
Move ir_generator.go to the ir package and split ir_generate_test.go into two files

shendiaomo:
This is a follow-up of #2484 and #2481


pkg/ir/ir_generator_test.go
```diff
@@ -0,0 +1,269 @@
+// Copyright 2020 The SQLFlow Authors. All rights reserved.
```
wangkuiyi, line 1: In future PR, we can rename `ir/ir_generator*.go` into `ir/generator*.go`.

wangkuiyi (approved):
LGTM!

PR #2499, opened 2020-06-19
Fill model metadata when releasing to model zoo

### LINKED ISSUE #2498: Release model to model zoo with metadata
We need to carry model metadata when releasing a model to model zoo.

Save model metadata with model itself.
When releasing a model, dump the model and extract the metadata, carry the metadata with the model to the zoo.

**Additional Notes**

lhw362950217:
Fixes #2498


pkg/model/model.go
```diff
@@ -34,11 +35,13 @@ import (
 
 const modelZooDB = "sqlflow"
 const modelZooTable = "sqlflow.trained_models"
+const modelMetaFileName = "model_meta.json"
 
 // Model represent a trained model, which could be saved to a filesystem or sqlfs.
 type Model struct {
 	workDir     string // We don't expose and gob workDir; instead we tar it.
 	TrainSelect string // TrainSelect is gob-encoded during I/O.
+	Meta        []byte // Meta json object
```
typhoonzero: Why not keep a JSON object or `map[string]interface{}`?

pkg/model/model.go
```diff
@@ -48,6 +51,14 @@ func New(cwd, trainSelect string) *Model {
 		TrainSelect: trainSelect}
 }
 
+// GetMeta return specified metadata as string
+func (m *Model) GetMeta(key string) string {
```
typhoonzero: Should be called `GetMetaAsString`

go.sum
```diff
@@ -178,6 +178,7 @@ github.com/joho/godotenv v1.3.0 h1:Zjp+RcGpHhGlrMbJzXTrZZPrWj+1vfm90La1wgB6Bhc=
 github.com/joho/godotenv v1.3.0/go.mod h1:7hK45KPybAkOC6peb+G5yklZfMxEjkZhHbwpqxOKXbg=
 github.com/jonboulle/clockwork v0.1.0/go.mod h1:Ii8DK3G1RaLaWxj9trq07+26W01tbo22gdxWY5EU2bo=
 github.com/json-iterator/go v0.0.0-20180612202835-f2b4162afba3/go.mod h1:+SdeFBvtyEkXs7REEP0seUULqWtbJapLOCVDaaPEHmU=
+github.com/json-iterator/go v1.1.8 h1:QiWkFLKq0T7mpzwOTu6BzNDbfTE8OLrYhVKYMLF46Ok=
```
shendiaomo: Remove the change of `go.mod` and `go.sum`? I think these two files work automatically.

pkg/model/model.go
lhw362950217: I searched some json lib, like [jsoniter](http://jsoniter.com/migrate-from-go-std.html), [jsonparser](https://github.com/buger/jsonparser), go std json, none of them provide full JSON object support. And it still not easy to use when we parse the json string to map[string]interface{}. I'm thinking about keeping the original byte array and use bellowing method to get anything we want. It is not the fastest, but may be easy to use. It's the same way as `jsonparser`. How do you think?
```go
jsoniter.Get(bytes, key).ToVal(OurStruct)  // get a Struct
jsoniter.Get(bytes, key).String()  // get a string
```

typhoonzero: Can you please try: https://blog.golang.org/json Decoding arbitrary data?

lhw362950217: I tried this, provide that parse json string to map[string]interface{} is possible, but, it not so easy to use the field, especially when the filed's path is deep. We can try `jsonparser` to save the meta as a Json struct. It is not a byte array and also keep the simplicity to get fields.

typhoonzero (approved):

PR #2509, opened 2020-06-22
Add parameter name to attribute checker function

sneaxiy:
Some attributes may be like `model.*`, and the actual attribute name would be provided in runtime. We should change attribute checker type from `func (i interface{}) error` to be `func (i interface{}, name string)` for better error messages.

### DIFF
--- a/pkg/attribute/attribute.go
+++ b/pkg/attribute/attribute.go
@@ -52,19 +52,19 @@ type description struct {
 	typ          reflect.Type
 	defaultValue interface{}
 	doc          string
-	checker      func(i interface{}, name string) error
+	checker      func(value interface{}, name string) error
 }
 
 // Int declares an attribute of int-typed in Dictionary d.
 func (d Dictionary) Int(name string, value interface{}, doc string, checker func(int, string) error) Dictionary {
-	interfaceChecker := func(v interface{}, name string) error {
-		if intValue, ok := v.(int); ok {
+	interfaceChecker := func(value interface{}, name string) error {
+		if intValue, ok := value.(int); ok {
 			if checker != nil {
 				return checker(intValue, name)
 			}
 			return nil
 		}
-		return fmt.Errorf("attribute %s must be of type int, but got %T", name, v)
+		return fmt.Errorf("attribute %s must be of type int, but got %T", name, value)
 	}
 
 	if value != nil {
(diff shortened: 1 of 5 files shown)


pkg/attribute/attribute.go
```diff
@@ -52,23 +52,23 @@ type description struct {
 	typ          reflect.Type
 	defaultValue interface{}
 	doc          string
-	checker      func(i interface{}) error
+	checker      func(i interface{}, name string) error
```
shendiaomo: Rename `i` as `value`?

sneaxiy: Done.

pkg/attribute/attribute.go
```diff
@@ -52,23 +52,23 @@ type description struct {
...
 // Int declares an attribute of int-typed in Dictionary d.
-func (d Dictionary) Int(name string, value interface{}, doc string, checker func(int) error) Dictionary {
-	interfaceChecker := func(v interface{}) error {
-		if intValue, ok := v.(int); ok {
+func (d Dictionary) Int(name string, value interface{}, doc string, checker func(int, string) error) Dictionary {
+	interfaceChecker := func(value interface{}, name string) error {
+		if intValue, ok := value.(int); ok {
 			if checker != nil {
-				return checker(intValue)
+				return checker(intValue, name)
```
typhoonzero: Can we just call `err := checker` and return the error message adding the attribute name with the error string, e.g. `return fmt.Errorf("attribute %s error: %s", name, string(err))`

sneaxiy: It may be impossible. Because if we do not add `name` parameter in `checker`, there is no way that the `checker` would know which attribute it is checking if the attribute name is `model.*` when registerring.

sneaxiy: Great idea! I think it would be more simpler. Done.

typhoonzero: I mean, can we just remove the `name` argument to achieve this?

shendiaomo (approved):

pkg/attribute/attribute.go
sneaxiy: > I mean, can we just remove the name argument to achieve this?

It may be not. Supposing that we call `Dictionary.Int("model.*", ...)` to register a checker to `model.*`. But the real attribute name (e.g. `model.n_class` for example) is provided in runtime. We cannot get the runtime attribute name when calling `Dictionary.Int("model.*", ...)`.

pkg/attribute/attribute.go
```diff
@@ -52,23 +52,25 @@ type description struct {
...
 	doc          string
-	checker      func(i interface{}) error
+	checker      func(value interface{}, name string) error
 }
 
 // Int declares an attribute of int-typed in Dictionary d.
 func (d Dictionary) Int(name string, value interface{}, doc string, checker func(int) error) Dictionary {
-	interfaceChecker := func(v interface{}) error {
-		if intValue, ok := v.(int); ok {
+	interfaceChecker := func(value interface{}, name string) error {
```
shendiaomo, line 60: It seems we don't have to pass `name` to `interfaceChecker` because `name` is included in the closure of `interfaceChecker`. Anyway, we can fix it in another pull request.

sneaxiy: We may need to pass `name` to `interfaceChecker`, because the `name` (for example, `model.*`) when calling `Dictionary.Int` may be different from the name when calling `checker(value, name)` (for example `model.n_class`).

pkg/attribute/attribute.go
typhoonzero: I see!

PR #2510, opened 2020-06-22
Use the same parser configuration for maxcompute and alisa.

brightcoder01:
- The data sources behind MaxCompute and Alisa are exactly the same - MaxCompute. The parser should be the same.
- While we test the SQL statement with Alisa data source in Playground, we first choose ODPS parser, if not found (that's the current situation), we will use calcite parser instead.

### DIFF
--- a/pkg/parser/external/parser.go
+++ b/pkg/parser/external/parser.go
@@ -38,13 +38,11 @@ func NewParser(dialect string) (Parser, error) {
 		return newJavaParser("hive"), nil
 	case "calcite":
 		return newJavaParser("calcite"), nil
-	case "maxcompute":
+	case "maxcompute", "alisa":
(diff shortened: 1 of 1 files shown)


lhw362950217 (approved):
LGTM

shendiaomo:
The change is not enough:
1. The "maxcompute" key will map calcite parser.
1. The "odps" parser cannot be release public.

brightcoder01:
> The change is not enough:
> 
> 1. The "maxcompute" key will map calcite parser.
> 2. The "odps" parser cannot be release public.

The `Maxcompute` key will first find ODPS parser, if not found, will choose calcite parser.
Yes, ODPS parser won't be released publicly. We will use calcite parser if the data source is MaxCompute or Alisa.

PR #2513, opened 2020-06-22
Fix %!s(<nil>) in  doc gen

sneaxiy:
Change the doc of `unknownType` parameter from 
```
<td>max_bin</td>
<td>%!s(<nil>)</td>
<td>Only used if tree_method is set to hist, Maximum number of discrete bins to bucket continuous features.</td>
```
to be
```
<td>max_bin</td>
<td>unknown</td>
<td>Only used if tree_method is set to hist, Maximum number of discrete bins to bucket continuous features.</td>
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,6 +10,7 @@ require (
 	github.com/asaskevich/govalidator v0.0.0-20190424111038-f61b66f89f4a
 	github.com/baiyubin/aliyun-sts-go-sdk v0.0.0-20180326062324-cfa1a18b161f // indirect
 	github.com/beltran/gohive v1.0.0 // indirect
+	github.com/bitly/go-simplejson v0.5.0
 	github.com/c-bata/go-prompt v0.0.0-20190826134812-0f95e1d1de2e
 	github.com/cosiner/argv v0.0.1 // indirect
 	github.com/docopt/docopt-go v0.0.0-20180111231733-ee0de3bc6815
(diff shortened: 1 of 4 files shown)


pkg/attribute/attribute.go
```diff
@@ -321,8 +321,14 @@ func (d Dictionary) GenerateTableInHTML() string {
 	<td>%s</td>
 	<td>%s</td>
 </tr>`
+
+		var typ interface{} = desc.typ
+		if typ == nil {
+			typ = "undetermined"
```
shendiaomo: Should "undetermined" be consistent with "unknown" in the codebase?

go.mod
```diff
@@ -10,6 +10,7 @@ require (
 	github.com/asaskevich/govalidator v0.0.0-20190424111038-f61b66f89f4a
 	github.com/baiyubin/aliyun-sts-go-sdk v0.0.0-20180326062324-cfa1a18b161f // indirect
 	github.com/beltran/gohive v1.0.0 // indirect
+	github.com/bitly/go-simplejson v0.5.0
```
weiguoz, line 13: Why we need the `go-simplejson` here?

sneaxiy: I do not know why. But the linter did add these.

pkg/attribute/attribute.go
sneaxiy: Done.

shendiaomo (approved):
lgtm

PR #2527, opened 2020-06-24
Log connection error of CLI to sqlflowserver

sneaxiy:
If `sqlflowserver` is crashed after CLI has been started, there would be no error printed in CLI.

Before this PR:
```
sqlflow> SELECT * xxx; // sqlflowserver is disconnected, there would be no error printed out.
sqlflow>
```

After this PR:
```
sqlflow> SELECT * xxx;
rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing dial tcp 127.0.0.1:50051: connect: connection refused"
sqlflow>
```


lhw362950217 (approved):
LGTM

PR #2537, opened 2020-06-26
Polish type pkg/parser/ast.Expr

wangkuiyi:
- Rename `Expr.typ` into `Expr.token` because the parser should provide this value a `%token` like `NUMER`, `STRING`, etc.
- Rename `Expr.val` into `Expr.value` to use English words other than shortened form. This field should be something like `123.45` if `token` is `NUMBER` or `"hello"` if `token` is `STRING`.
- Enrich comments.

This PR addresses the question https://github.com/sql-machine-learning/sqlflow/pull/2533#discussion_r445263396 with a short answer "no, I am afraid that we cannot".


sneaxiy (approved):
LGTM.

PR #2544, opened 2020-06-28
Use lex member as parser result

### LINKED ISSUE #2541: Parse result shouldn't be a global variable
Currently, our parser contains a global variable

https://github.com/sql-machine-learning/sqlflow/blob/36eab7802cff31d862c1983d23407647339fd18a/pkg/parser/extended_syntax_parser.y#L154

And the start grammar rule writes the parsing result (something like an AST) into this global variable

https://github.com/sql-machine-learning/sqlflow/blob/36eab7802cff31d862c1983d23407647339fd18a/pkg/parser/extended_syntax_parser.y#L221-L227

This global variable prevents our parsing code from being thread-safe, and we need a mutex to synchronize concurrent calls to the parser.

https://github.com/sql-machine-learning/sqlflow/blob/36eab7802cff31d862c1983d23407647339fd18a/pkg/parser/extended_syntax_parser.y#L592-L606

We should correct the design by setting the result into a data member of the lexer instance using techniques described in https://about.sourcegraph.com/go/gophercon-2018-how-to-write-a-parser-in-go#using-goyacc.

typhoonzero:
Fix https://github.com/sql-machine-learning/sqlflow/issues/2541

### DIFF
--- a/pkg/parser/extended_syntax_parser.y
+++ b/pkg/parser/extended_syntax_parser.y
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"log"
 	"strings"
-	"sync"
 )
 
 /* expr defines an expression as a Lisp list.  If len(val)>0,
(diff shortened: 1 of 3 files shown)


sneaxiy (approved):
LGTM++.

PR #2550, opened 2020-06-28
Enable log in sqlflowserver

lhw362950217:
the test code disable log in init() func, let's enable it in non-test mode. Supply a default file name '-' as stdout, like most of the  tools in Linux do.


pkg/log/config.go
```diff
@@ -52,6 +53,8 @@ func setOutput(filename string) {
 	filename = strings.Trim(filename, " ")
 	if filename == "/dev/null" {
 		logrus.SetOutput(ioutil.Discard)
+	} else if filename == "-" {
```
wangkuiyi: Should we use the condition `filename == ""` instead of `filename == "-"` to judge if log to stdout?

wangkuiyi (approved):

weiguoz:
https://github.com/sql-machine-learning/sqlflow/blob/8e60da41d3337f812b10bcd2e65cf2b49f8ca63b/pkg/log/config.go#L56-L57
It looks like the changed file doesn't follow [the comment](https://github.com/sql-machine-learning/sqlflow/pull/2550#issue-441053566) which cares about `-`.

lhw362950217:
> https://github.com/sql-machine-learning/sqlflow/blob/8e60da41d3337f812b10bcd2e65cf2b49f8ca63b/pkg/log/config.go#L56-L57
> 
> 
> It looks like the changed file doesn't follow [the comment](https://github.com/sql-machine-learning/sqlflow/pull/2550#issue-441053566) which cares about `-`.

I take @wangkuiyi 's advice and delete the '-'

PR #2551, opened 2020-06-28
Support libsvm column

sneaxiy:


wangkuiyi:
@sneaxiy I would suggest you file a design doc PR for this change before filing code change PRs. It is much more important to learn to lead a project than coding it.  Also, to practice coding well, we need to put ourselves on the position of being a leader, other than the coder.

If I am going to add this feature of libsvm columns, I would write a document about the following points.

- What ML toolkits we already supported and why we need to support libsvm.
- What are columns and how do we implement the idea.
- What is special about libsvm columns and how should we change the code to add libsvm columns.

wangkuiyi (approved):
If it is urgent to implement this feature to meeting users' requests, please feel free to merge this PR and polish later.

PR #2564, opened 2020-06-29
Fix deleting previous model dir on OSS

typhoonzero:
Go OSS API `IsObjectExist` may return error information, removing this is fine since the below code can also detect objects not exist errors.


weiguoz (approved):
LGTM

PR #2579, opened 2020-06-30
[Design doc] Redesign numeric column clause

sneaxiy:
Remove the `NUMERIC` column clause. We can use the `DENSE` and `SPARSE` to do the same things.


doc/design/numeric_typed_column_clause.md
```diff
@@ -0,0 +1,164 @@
...
+LABEL label
+COLUMN DENSE(c1, 10), DENSE(c2, 10)
+INTO result_table;
+```
+- For sparse data, the SQL statement would be:
+```sql
+SELECT c1, c2, label FROM train_table
+TO TRAIN DNNRegressor
+LABEL label
+COLUMN SPARSE(c1, 10), SPARSE(c2, 10)
```
typhoonzero: Also add examples of how to embed SPARSE to other columns like `EMBEDDING(SPARSE(...))`?

typhoonzero (dismissed):
LGTM basically

doc/design/numeric_column_clause.md
```diff
@@ -0,0 +1,164 @@
...
+        SELECT c1, c2, label FROM train_table
+        TO TRAIN DNNRegressor
+        LABEL label
+        COLUMN NUMERIC(SPARSE(c1, 10)), NUMERIC(SPARSE(c2, 10)) -- 10 is the dense length of the decoded sparse vector
+        INTO result_table;
+        ```
+
+There are some problems in the current design:
+
+- Both `NUMERIC` and `DENSE/SPARSE` have the `length/shape` parameter.
```
brightcoder01, line 87: Do we support 2 or more dimension? If not, we may not need shape, length is enough

sneaxiy: Yes, we should support 2 or more dimensions. For example, some operators like convolution is sensitive to shape but not length.

doc/design/numeric_typed_column_clause.md
sneaxiy: Done.

typhoonzero (approved):
LGTM

PR #2593, opened 2020-07-01
Remove NUMERIC column clause

sneaxiy:
As discussed in https://github.com/sql-machine-learning/sqlflow/pull/2593 , we should remove the `NUMERIC`, and use the `DENSE/SPARSE` only.

### DIFF
--- a/doc/design/xgboost_feature_column.md
+++ b/doc/design/xgboost_feature_column.md
@@ -27,7 +27,7 @@ SQLFlow `COLUMN` clauses support the following listed feature columns, and they 
 
 | SQLFlow keywords | TensorFlow API                                              | Description                                                          | 
 |------------------|-------------------------------------------------------------|----------------------------------------------------------------------|
-| NUMERIC          | tf.feature_column.numeric_column                            | Raw numeric feature column without any pre-processing                | 
+| DENSE|SPARSE     | tf.feature_column.numeric_column                            | Raw numeric feature column without any pre-processing                | 
 | BUCKET           | tf.feature_column.bucketized_column                         | Transform input integer to be the bucket id divided by boundaries    |
 | CATEGORY_ID      | tf.feature_column.categorical_column_with_identity          | Identity mapping of integer feature column                           |
 | CATEGORY_HASH    | tf.feature_column.categorical_column_with_hash_bucket       | Using hash algorithm to map string or integer to category id         |
--- a/doc/design/xgboost_feature_column.md
+++ b/doc/design/xgboost_feature_column.md
@@ -27,7 +27,7 @@ SQLFlow `COLUMN` clauses support the following listed feature columns, and they 
 
 | SQLFlow keywords | TensorFlow API                                              | Description                                                          | 
 |------------------|-------------------------------------------------------------|----------------------------------------------------------------------|
-| DENSE|SPARSE     | tf.feature_column.numeric_column                            | Raw numeric feature column without any pre-processing                | 
+| DENSE            | tf.feature_column.numeric_column                            | Raw numeric feature column without any pre-processing                | 
 | BUCKET           | tf.feature_column.bucketized_column                         | Transform input integer to be the bucket id divided by boundaries    |
 | CATEGORY_ID      | tf.feature_column.categorical_column_with_identity          | Identity mapping of integer feature column                           |
 | CATEGORY_HASH    | tf.feature_column.categorical_column_with_hash_bucket       | Using hash algorithm to map string or integer to category id         |
--- a/doc/language_guide.md
+++ b/doc/language_guide.md
@@ -115,7 +115,7 @@ COLUMN column_expr [, column_expr ...]
     [COLUMN column_expr [, column_expr ...] FOR column_name ...]
 ```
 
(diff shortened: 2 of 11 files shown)


doc/language_guide.md
```diff
@@ -158,15 +158,15 @@ SQLFlow supports specifying various feature columns in the column clause and lab
  feature column type | usage | field type | example
 ---|---|---|---
  X | field | int/float/double | 3.14
- NUMERIC | NUMERIC(field, n[,delimiter]) | string/varchar[n] | "0.2,1.7,0.6"
+ DENSE | DENSE(field, n[,delimiter]) | string/varchar[n] | "0.2,1.7,0.6"
```
shendiaomo, line 161: There's no `SPARSE` here?

pkg/ir/ir_generator.go
```diff
@@ -533,47 +531,17 @@ func parseDefaultNumericColumn(el *parser.Expr) (*NumericColumn, error) {
...
-			Shape:    shape,
-			IsSparse: false,
-		},
+		FieldDesc: fieldDesc,
 	}, nil
 }
 
 func parseBucketColumn(el *parser.ExprList) (*BucketColumn, error) {
-	help := "BUCKET([NUMERIC(...)|col_name], BOUNDARIES)"
+	help := "BUCKET([DENSE(...)|col_name], BOUNDARIES)"
```
shendiaomo, line 544: Can't we specify `SPARSE`  here?

pkg/ir/derivation.go
```diff
@@ -161,6 +161,11 @@ func fillCSVFieldDesc(cellData string, fieldDescMap FieldDescMap, fieldName stri
 		// implicit set shape if shape is not provided in SQL COLUMN clause
 		fieldDescMap[fieldName].Shape = []int{len(values)}
 	}
+
+	if fieldDescMap[fieldName].IsSparse {
+		fieldDescMap[fieldName].DType = Int // force to set sparse tensor in CSV format to be of type Int
```
typhoonzero: Why? We currently do not have test cases that use sparse float tensors, should add FIXME comment here maybe.

pkg/ir/ir_generator.go
```diff
@@ -492,11 +490,11 @@ func inferStringValue(expr string) interface{} {
 func parseFeatureColumn(el *parser.ExprList) (FeatureColumn, error) {
 	head := (*el)[0].Value
 	if head == "" {
-		return nil, fmt.Errorf("column description expects format like NUMERIC(key) etc, got %v", el)
+		return nil, fmt.Errorf("column description expects format like DENSE(key) etc, got %v", el)
 	}
 
 	switch strings.ToUpper(head) {
-	case numeric:
+	case dense, sparse:
```
typhoonzero, line 497: Can sparse tensor used in numeric columns?

doc/design/xgboost_feature_column.md
```diff
@@ -27,7 +27,7 @@ SQLFlow `COLUMN` clauses support the following listed feature columns, and they
 
 | SQLFlow keywords | TensorFlow API                                              | Description                                                          | 
 |------------------|-------------------------------------------------------------|----------------------------------------------------------------------|
-| NUMERIC          | tf.feature_column.numeric_column                            | Raw numeric feature column without any pre-processing                | 
+| DENSE|SPARSE     | tf.feature_column.numeric_column                            | Raw numeric feature column without any pre-processing                |
```
brightcoder01: tf.feature_column.numeric_column cannot accept a sparse tensor.

doc/language_guide.md
sneaxiy: Added. Thanks!

pkg/ir/ir_generator.go
sneaxiy: Yes. The `tf.feature_column.bucketized_column` does not support sparse tensor.

pkg/ir/derivation.go
sneaxiy: In fact, it is the previous logic.

https://github.com/sql-machine-learning/sqlflow/blob/73cda1925d13cc92b08a7864018708ff988f5685/pkg/ir/ir_generator.go#L889-L892

I guess it is because we only support sparse tensor encoded in CSV format currently, and the numeric values in the cell data in CSV format are the indices where the tensor values are 1. For example, "3,5,7" indicates a sparse tensor `x`, and `x[3] = x[5] = x[7] = 1`, and the other values of `x` are 0. Since the index is always integral, the sparse tensor encoded in CSV format would be tensor of integers. 

I add a `FIXME` comment, and we should remove this constraint if we will support other sparse tensor data format in the future.

doc/design/xgboost_feature_column.md
sneaxiy: Removed. Thanks!

pkg/ir/ir_generator.go
sneaxiy: Sparse tensor is not allowed in TensorFlow numeric columns, but it would work in XGBoost numeric columns defined by ourselves.

pkg/ir/derivation.go
typhoonzero: Cool. Thanks

pkg/ir/ir_generator.go
typhoonzero: I see.

typhoonzero (approved):
LGTM++

PR #2597, opened 2020-07-01
Go mod tidy

weiguoz:


wangkuiyi (approved):
It seems that we should add `go mod tidy` to the pre-commit hooks.  We can reuse the hook defined in https://github.com/dnephin/pre-commit-golang

weiguoz:
> It seems that we should add `go mod tidy` to the pre-commit hooks. We can reuse the hook defined in https://github.com/dnephin/pre-commit-golang

Good idea.
Once the `go-mod-tidy` hook [supports sub-folders checking](https://github.com/dnephin/pre-commit-golang/commit/3c835496ef7101a0d1e1e44afa02f361683382b1#commitcomment-40060453) and a new git-tag is built, I will add `go mod tidy` to the hooks.

PR #2604, opened 2020-07-02
Release model runs on the server

### LINKED ISSUE #2574: Releasing models in a protected network environment for on-premise deployments
Our current [design](https://github.com/sql-machine-learning/sqlflow/blob/develop/doc/design/cli_model_zoo_cmd.md) uses the command-line client to download trained models from a database and then upload it to the model zoo server for sharing. In some on-premise deployments, we may have a database deployment that not accessible from the developer's laptop. To access the database, we should use a specific web interface or login into a machine that has access to the database. In this case, the command-line tool can not directly download a model from the database.

After some discussion with @lhw362950217, we agree that the best way should be:

1. when releasing model repo, use the command `sqlflow release repo ...` to directly upload a local repo directory to model zoo server.
2. when releasing a trained model, the command `sqlflow release model ...` need to send the database connection string with credentials to the model zoo server, then model zoo server will copy the model from private storage to a public place.

typhoonzero:
Fix https://github.com/sql-machine-learning/sqlflow/issues/2574


cmd/sqlflow/model_zoo_client.go
```diff
@@ -68,63 +65,25 @@ func tarRepo(repoDir, tarName string) (string, error) {
 }
 
 func releaseModel(opts *options) error {
-	if opts.DataSource == "" {
```
lhw362950217: Can we keep the original functionality and add another command to release model on server? So that, we don't rely on server if the client can directly access the datasource.

typhoonzero: Done

pkg/modelzooserver/modelzooserver.go
```diff
@@ -321,8 +322,125 @@ func (s *modelZooServer) DropModelRepo(ctx context.Context, req *pb.ReleaseModel
...
+	}
+	sendFile, err := os.Open(tarFile)
+	if err != nil {
+		return nil, err
+	}
+	defer sendFile.Close()
+
+	// model name is the original INTO clause, it can be like db.table
+	// reformat it to db_table as the table name in model zoo database.
+	modelTableName := fmt.Sprintf("%s.%s", publicModelDB, strings.ReplaceAll(req.Name, ".", "_"))
```
lhw362950217: This is not compatible with local release mode, local mode only use the table name. We can make them the same.

pkg/modelzooserver/modelzooserver.go
```diff
@@ -321,8 +322,125 @@ func (s *modelZooServer) DropModelRepo(ctx context.Context, req *pb.ReleaseModel
...
+	// FIXME(typhoonzero): only hive need to pass session
+	sqlf, err := sqlfs.Create(s.DB.DB, s.DB.DriverName, modelTableName, nil)
+	if err != nil {
+		return nil, fmt.Errorf("cannot create sqlfs file %s: %v", modelTableName, err)
+	}
+	defer sqlf.Close()
+
+	buf := make([]byte, 4096)
```
lhw362950217: Maybe we should save the model to DB after  all checks has been done.

lhw362950217 (dismissed):
LGTM

lhw362950217 (approved):
LGTM

PR #2609, opened 2020-07-03
Rename LibSVM to kv

sneaxiy:


pkg/ir/derivation.go
```diff
@@ -227,39 +227,41 @@ func fillFieldDescByDataType(cellData string, fieldDescMap FieldDescMap, fieldNa
 }
 
 const (
-	csv    = "csv"
-	libsvm = "libsvm"
+	csv = "csv"
```
wangkuiyi, line 230: I feel that we might not need names like CSV and KV in the source code. Should we just refer to the concepts by dense and sparse vectors?

sneaxiy: I am afraid that to just use dense and sparse vectors is not enough. Because:

- In Go side, we do feature derivation to get the data format and data type of the column values. During derivation, we need to know the data format first, and then parse the cell string according to the data format, and finally infer the data type from the parsed result. Because we will support sparse tensors in the form of "3,5,7" or "1:2 7:0.5", we cannot distinguish these 2 different forms if we only use dense and sparse concepts in Go side.

- In Python side, we also need to parse the column cell according to the data format. Therefore, we need the concepts of different data formats in Python side too.

Both Go and Python side should know the data format (csv, kv or other formats) in the source code.

typhoonzero (approved):

PR #2615, opened 2020-07-03
Add glpk solver in docker image for tests

sneaxiy:


docker/ci/install-pips.bash
```diff
@@ -38,3 +38,5 @@ pip install --quiet \
     PyUtilib==5.8.0 \
     pyomo==5.6.9
 
+# NOTE: glpk-utils is a required solver package of Pyomo.
+apt-get -qq install -y glpk-utils
```
typhoonzero: Put this line to `docker/dev/install.sh` ?

typhoonzero (dismissed):

docker/ci/install-pips.bash
sneaxiy: Great idea! Done.

typhoonzero (approved):

PR #2617, opened 2020-07-03
Add tidb parser deps support

typhoonzero:


go/parser/external/tidb_parser.go
```diff
@@ -125,6 +126,32 @@ func (p *tidbParser) doParse(program string) ([]*Statement, int, error) {
 			} else {
 				retStmts = append(retStmts, &Statement{String: nodes[0].Text()})
 				pos += len(sql)
+
+				restoreFlags := format.RestoreStringSingleQuotes | format.RestoreKeyWordUppercase
```
wangkuiyi: I am afraid that this code snippet increases the cyclomatic complexity of this function. How about moving it into a new function and call the new function here?

typhoonzero: Done

wangkuiyi (approved):
LGTM+

PR #2621, opened 2020-07-03
Add Python executor for optimize clause

sneaxiy:
Add Python executor for the `TO MAXIMIZE/MINIMIZE` clause, and run the generated code using native Pyomo APIs. In this way, SQLFlow can support the `TO MAXIMIZE/MINIMIZE` clause without using the OptFlow.


go/executor/executor.go
```diff
@@ -364,9 +365,76 @@ func generateOptFlowOptimizeCodeAndExecute(cl *ir.OptimizeStmt, submitter *pytho
...
+
+		if !isVariable {
+			continue
+		}
+
+		typ, err = fieldType(driver, typ)
+		if err != nil {
+			return err
+		}
+		tableColumnsStrList = append(tableColumnsStrList, fmt.Sprintf("%s %s", name, typ))
```
typhoonzero, line 403: Are we writing the original model definition data together with the result?

typhoonzero (approved):

go/executor/executor.go
sneaxiy: No. Pyomo does not provide an official interface to save the model. It seems that only `pickle` can save the model (please see https://stackoverflow.com/questions/50834258/how-to-save-pickle-a-model-instance-in-pyomo and https://github.com/Pyomo/pyomo/blob/8eaa247ae41694edd317606b44ef04824608853e/pyomo/core/tests/unit/test_pickle.py).

But I found that saving the model with lambda expressions using `pickle` (the lambda expressions in objective and constraint rules) would fail. I would try to save the model in future works.

typhoonzero: Well, I mean are we writing the original data in the input table to the output table together with the results?

sneaxiy: We only write the original data of variable columns in the result table. For example, if there are 2 variables named plants and markets in the model definition, the result data has 3 columns named plants, markets and shipment, where shipment is the solved  numeric result of the model. Other columns of the original table would not be written in the result table if the column is not the variable.

PR #2627, opened 2020-07-06
Rename feature_specs and label_spec to feature_metas and label_meta respectively

shendiaomo:
The naming of `feature_metas` and `label_meta` are not consistent, this PR tries to fix this problem.


wangkuiyi (approved):
Agreed with the renaming.

According to Wikipedia

> Meta is a prefix meaning more comprehensive or transcending.

It is not a word.

typhoonzero:
@wangkuiyi so which name do you prefer? `xxx_meta` or `xxx_spec`. I prefer the name `xxx_spec` since it's the specification of the field. `meta` is a too generic name.

PR #2637, opened 2020-07-07
Fix maxcompute ci job

typhoonzero:
The Maxcompute test cases are actually not running on travis ci.

`maxcompute.sh` is executed inside the Docker container, and we start the container by passing `-e SQLFLOW_TEST_DB_MAXCOMPUTE_AK=$MAXCOMPUTE_AK`, so use `SQLFLOW_TEST_DB_MAXCOMPUTE_AK` directly in the container.


typhoonzero:
The current code of dealing table columns is fragile and hard to fix. Will try to polish it in this PR.

sneaxiy (commented):
Please solve the conflict.

python/runtime/maxcompute.py
```diff
@@ -32,8 +43,7 @@ def read_feature(raw_val, feature_spec):
                                         sep=feature_spec["delimiter"])
                 indices = indices.reshape(indices.size, 1)
                 values = np.ones([indices.size], dtype=np.int32)
-                dense_shape = np.array(feature_metas[name]["shape"],
-                                       dtype=np.int64)
+                dense_shape = np.array(feature_spec["shape"], dtype=np.int64)
```
sneaxiy, line 46: Seems that we can remove the `read_feature` method?

sneaxiy (approved):
LGTM with a tiny comment.

python/runtime/maxcompute.py
typhoonzero: Will do in the next PR.

typhoonzero: `read_feature ` is called by `read_features_from_row` which will be used when predicting to get columns that are used as features to feed to the model.

PR #2647, opened 2020-07-08
Correct role name

weiguoz:


go/codegen/pai/tensorflow.go
```diff
@@ -236,7 +236,20 @@ func getCheckpointDir(ossModelPath, project string) (string, error) {
...
-	roleName := fmt.Sprintf("pai2oss_%s", project)
+	roleName := genRoleName(project)
 	// format the oss checkpoint path with ARN authorization.
 	return fmt.Sprintf("%s/?role_arn=%s/%s&host=%s", ossURL, ra.Arn, roleName, ra.Host), nil
 }
+
+func genRoleName(project string) string {
+	var rn bytes.Buffer
+	rn.WriteString("pai2oss")
+	for _, ch := range project {
```
shendiaomo, line 249: How about using `regexp` to replace this loop like:
```Go
re := regexp.MustCompile(`[^a-z0-9]`)
rn.WriteString(re.ReplaceAllString(strings.ToLower(project), ""))
```

go/codegen/pai/tensorflow_test.go
```diff
@@ -29,6 +29,15 @@ func TestGetCheckpointDir(t *testing.T) {
...
 	a.Equal(expectedCkp, ckpoint)
 }
+
+func TestGenRoleName(t *testing.T) {
+	a := assert.New(t)
+	a.Equal(genRoleName("a"), "pai2ossa")
+	a.Equal(genRoleName("abc"), "pai2ossabc")
+	a.Equal(genRoleName("a_b_c"), "pai2ossabc")
+	a.Equal(genRoleName("a_b_C"), "pai2ossabc")
+	a.Equal(genRoleName("a+b2_C"), "pai2ossab2c")
```
shendiaomo, line 42: Are these automatic conversions desired? What if we just report an error when characters except `[a-z0-9]` are present.

go/codegen/pai/tensorflow.go
weiguoz: It is a concise way, especially one pass, and easy to understand.
So I didn't choose using `regexp`.

go/codegen/pai/tensorflow_test.go
weiguoz: The `project` name contains `_` often. We should remove such characters to generate a valid role name.

typhoonzero (approved):
LGTM++

PR #2654, opened 2020-07-09
Add get schema tool function in db.py

lhw362950217:
Feature derivation and predict both rely on table schema, so we add a tool function in db.py


python/runtime/db.py
```diff
@@ -370,3 +371,37 @@ def buffered_db_writer(driver,
         yield w
     finally:
         w.close()
+
+
+def get_table_schema(datasource, table):
```
typhoonzero: Can we pass a connection as an argument, so that we don't need to create a new connection every time we call this function?

lhw362950217: Done

typhoonzero (approved):

PR #2658, opened 2020-07-10
Move OptFlow model generation codes from Python to Go

sneaxiy:
When using OptFlow, we should read data from the ODPS table and generate the Pyomo model on the Go side instead of on the Python side, since we cannot get AK/SK on the Python side.


typhoonzero (approved):
不明觉厉。I'll check the details later. You can merge this fast so that we can test it in our deployment.

Issue #2656, opened 2020-07-10
Migrate PAI Submitter to Python runtime


lhw362950217:
**Is your feature request related to a problem? Please describe.**
Now we should migrate PAI TensorFlow train/predict to Python because feature derivation and following process should execute in Python runtime.

**Describe the solution you'd like**
Modify below code in Python runtime package.

- [x] Support PAI TensorFlow train task submitting.
- [ ] Support PAI TensorFlow prediction task submitting.
- [ ] Support PAI TensorFlow explain task submitting.
- [ ] Support PAI TO RUN statement task submitting.
- [ ] Migrate other algorithm like `kmeans`, `RF` and `XGBoost`
- [ ] Refine db.py to be more Object Oriented so easy to use

**Describe alternatives (optional)**

**Additional Notes**

PR #2661, opened 2020-07-13
Add PAI submitter for TensorFlow prediction

lhw362950217:
Related to #2656

### DIFF
--- a/python/runtime/db.py
+++ b/python/runtime/db.py
@@ -375,7 +375,7 @@ def get_table_schema(conn, table):
     """Get column name and type of given table
 
     Args:
-        conn: an open database connection, this function will leave it open
+        conn: a database connection, this function will leave it open
         table: table name or db.table
 
     Returns:
@@ -404,14 +404,14 @@ def get_table_schema(conn, table):
 
 
 def exec(conn, sql_stmt):
-    """Execute given sql statement and return if successful
+    """Execute the given sql statement and return True on success
 
     Args:
-        conn: an open database connection, this function will leave it open
+        conn: a database connection, this function will leave it open
         sql_stmt: the sql statement to execute
     
     Returns:
-        If the query is successful
+        True on success and False on failure
     """
     if conn.driver == "maxcompute":
         inst = conn.execute_sql(sql_stmt)
--- a/python/runtime/db.py
+++ b/python/runtime/db.py
@@ -375,7 +375,7 @@ def get_table_schema(conn, table):
     """Get column name and type of given table
 
     Args:
-        datasource: datasource to connect, including auth info
+        conn: an open database connection, this function will leave it open
         table: table name or db.table
 
     Returns:
--- a/python/runtime/db.py
+++ b/python/runtime/db.py
@@ -401,3 +401,27 @@ def get_table_schema(conn, table):
             fields.append((field[0], field[1]))
         cursor.close()
     return fields
+
+
+def exec(conn, sql_stmt):
+    """Execute given sql statement and return if successful
+
+    Args:
+        conn: an open database connection, this function will leave it open
+        sql_stmt: the sql statement to execute
+    
+    Returns:
+        If the query is successful
+    """
+    if conn.driver == "maxcompute":
+        inst = conn.execute_sql(sql_stmt)
+        return inst.is_successful
+    else:
+        try:
+            cur = conn.cursor()
+            cur.execute(sql_stmt)
+            conn.commit()
+            cur.close()
+            return True
+        except:
+            return False
--- a/python/runtime/pai/entry/tensorflow.py
+++ b/python/runtime/pai/entry/tensorflow.py
@@ -14,18 +14,22 @@
 import pickle
 import types
 
-from runtime.tensorflow import is_tf_estimator, train
+from runtime.tensorflow import is_tf_estimator, predict, train
+from runtime.tensorflow.diag import SQLFlowDiagnostic
 from runtime.tensorflow.pai_distributed import (define_tf_flags,
                                                 set_oss_environs)
 
 from .. import model
 
 
-def load_oss_model(oss_model_dir, estimator):
-    set_oss_environs(FLAGS)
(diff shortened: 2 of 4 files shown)


python/runtime/db.py
```diff
@@ -375,7 +375,7 @@ def get_table_schema(conn, table):
     """Get column name and type of given table
 
     Args:
-        datasource: datasource to connect, including auth info
+        conn: an open database connection, this function will leave it open
```
typhoonzero: an open database connection -> a database connection

python/runtime/db.py
```diff
@@ -401,3 +401,27 @@ def get_table_schema(conn, table):
             fields.append((field[0], field[1]))
         cursor.close()
     return fields
+
+
+def exec(conn, sql_stmt):
+    """Execute given sql statement and return if successful
```
typhoonzero: Execute given sql statement and return if successful -> Execute the given SQL statement and return true when success

python/runtime/db.py
```diff
@@ -401,3 +401,27 @@ def get_table_schema(conn, table):
...
+
+    Args:
+        conn: an open database connection, this function will leave it open
```
typhoonzero: a database connection

python/runtime/db.py
```diff
@@ -401,3 +401,27 @@ def get_table_schema(conn, table):
...
+        sql_stmt: the sql statement to execute
+    
+    Returns:
+        If the query is successful
```
typhoonzero: True if the query is successful

python/runtime/db.py
```diff
@@ -401,3 +401,27 @@ def get_table_schema(conn, table):
...
+    """
+    if conn.driver == "maxcompute":
+        inst = conn.execute_sql(sql_stmt)
+        return inst.is_successful
+    else:
+        try:
+            cur = conn.cursor()
+            cur.execute(sql_stmt)
+            conn.commit()
```
typhoonzero, line 423: Do we need to `commit` every time?

python/runtime/pai/entry/tensorflow.py
```diff
@@ -70,11 +79,21 @@ def save_oss_model(oss_model_dir, estimator, num_workers, feature_column_names,
...
 
+def do_predict(params):
+    predict.pred(params["datasource"], params["estimator"], params["select"],
+                 params["result_table"], params["feature_columns"],
+                 params["feature_column_names"],
+                 params["feature_column_names_map"], params["result_col_name"],
+                 **params)
+
+
 def entry_point():
```
typhoonzero: entry_point -> entrypoint

python/runtime/pai/submitter.py
```diff
@@ -377,3 +378,139 @@ def submit_pytf_train(datasource, estimator, select, validation_select,
     # save trained model to sqlfs
     save_model_to_sqlfs(datasource, path_to_save, model_name)
     drop_tmp_tables([train_table, val_table], datasource)
+
+
+def get_oss_saved_model_type(model_name, project):
```
typhoonzero: `get_oss_saved_model_type_and_estimator_name`?

python/runtime/pai/submitter.py
```diff
@@ -377,3 +378,139 @@ def submit_pytf_train(datasource, estimator, select, validation_select,
...
+    """Get oss model type and estimator name, model can be:
+    1. randomforests: model is saved by pai
```
typhoonzero: randomforests -> PAI ML models

python/runtime/pai/submitter.py
```diff
@@ -377,3 +378,139 @@ def submit_pytf_train(datasource, estimator, select, validation_select,
...
+        select: sql statement to get prediction data set
+        result_table: the table name to save result
+        label_column: name of the label column, if not exist in select
+            result, we will add a int column in the result table
+    """
+    conn = db.connect_with_data_source(datasource)
+    db.exec(conn, "DROP TABLE IF EXISTS %s" % result_table)
+    create_table_sql = "CREATE TABLE %s AS SELECT * FROM %s LIMIT 0" % (
+        result_table, select)
+    if conn.driver == "hive":
```
typhoonzero: It's a pai submitter, maybe Hive will never be used here.

python/runtime/pai/submitter.py
```diff
@@ -377,3 +378,139 @@ def submit_pytf_train(datasource, estimator, select, validation_select,
...
+        datasource: current datasource
+        select: sql statement to get prediction data set
+        result_table: the table name to save result
+        label_column: name of the label column, if not exist in select
+            result, we will add a int column in the result table
+    """
+    conn = db.connect_with_data_source(datasource)
+    db.exec(conn, "DROP TABLE IF EXISTS %s" % result_table)
+    create_table_sql = "CREATE TABLE %s AS SELECT * FROM %s LIMIT 0" % (
+        result_table, select)
```
typhoonzero, line 474: The result label name may be different from the label name used when training.

python/runtime/db.py
lhw362950217: We'd better do this in case some transactions are not committed when closing connection. https://www.python.org/dev/peps/pep-0249/#Connection.close

python/runtime/pai/submitter.py
lhw362950217: Line 472 add a column when the result label is not present in the datasource

typhoonzero: We may have below cases when predicting:

1. train with `select f1, f2, label TO TRAIN ...` and predict with `select f1, f2 TO PREDICT pred.pred_result`, the label column is not selected int the predict select statement.
1. train with `select f1, f2, label TO TRAIN ...` and predict with `select f1, f2, label TO PREDICT pred.label`, label column is selected and the output column name is the same as the train label column.
1. train with `select f1, f2, label TO TRAIN ...` and predict with `select f1, f2, label TO PREDICT pred.result`, label column is selected but the output column name is different with the train label column.

typhoonzero (approved):

PR #2669, opened 2020-07-14
Fix Dataworks Python Program Run Error

sneaxiy:
In Dataworks, no TensorFlow and Docker would be found in the step container. Therefore, we cannot run the Python program generated when running the optimize clause.

This PR works on:
- fixes this error
- make the step image to run OptFlow only depends on `oss2` and `requests` (which are used in OptFlow job submission and status querying) instead of `pyomo`.

### DIFF
--- a/go/executor/cmd.go
+++ b/go/executor/cmd.go
@@ -71,8 +71,11 @@ func sqlflowCmd(cwd, driverName string) (cmd *exec.Cmd) {
 		cmd = exec.Command("docker", "run", "--rm",
 			fmt.Sprintf("-v%s:/work", cwd),
 			"-w/work", "--network=host", "-i", tfImg, "python")
+	} else if hasPython() {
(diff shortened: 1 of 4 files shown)


go/executor/cmd.go
```diff
@@ -71,8 +71,11 @@ func sqlflowCmd(cwd, driverName string) (cmd *exec.Cmd) {
 		cmd = exec.Command("docker", "run", "--rm",
 			fmt.Sprintf("-v%s:/work", cwd),
 			"-w/work", "--network=host", "-i", tfImg, "python")
+	} else if hasPython() {
```
typhoonzero, line 74: Maybe add comments on why we should have this branch?

typhoonzero (approved):

go/executor/cmd.go
sneaxiy: Will do it in the next PR.

PR #2670, opened 2020-07-14
[OptFlow Programming]Support combinational optimize cases and remove the code to obtain table data 

sneaxiy:
This PR works for math programming using OptFlow:

- support combinational optimize cases. For example, `variables = "shipment(plants, markets)"`. We create a temporary table with 3 columns: `plants, markets, plants__markets`, where `plants__markets` are the concatenated values of the column `plants` and `markets`. 

- do not get the whole table data in Go or Python side when there is any `GROUP BY`. If the constraint expression is `SUM(shipment) <= capacity GROUP BY plants`, we generate the OptFlow code like this:

```
for value, index in zip(*__import__("numpy").unique(@input["plants"], return_index=True)): 
    sum([@X[i_0] for i_0 in __import__("numpy").where(@input["plants"] == value)[0]])<=@input["capacity"][index]
```


go/codegen/optimize/codegen.go
```diff
@@ -327,6 +96,23 @@ func tryConvertToAggregationFunction(token string) string {
 	return aggregationFunctions[strings.ToUpper(token)]
 }
 
+// tryConvertComparisionToken tries to convert the comparision token in the
+// SQL statement to the desired comparison token which would be accepted by
+// the optimize model. If the token is not a comparision token, it returns
+// "".
+func tryConvertComparisionToken(token string) string {
+	var comparisionTokenMap = map[string]string{
+		"LE": "<=",
```
typhoonzero: Why do we need to support things like `LE` in the SQL statements? Can we use only expressions like `<=`?

go/executor/pai.go
```diff
@@ -466,7 +466,33 @@ func executeOptimizeUsingOptFlow(pythonExecutor *pythonExecutor, stmt *ir.Optimi
...
+	if len(stmt.Variables) > 1 {
+		joinedVarName := strings.Join(stmt.Variables, "__")
+		concatColumnNames := make([]string, 0)
+		for i, v := range stmt.Variables {
+			if i >= 1 {
+				concatColumnNames = append(concatColumnNames, `','`)
+			}
+			concatColumnNames = append(concatColumnNames, v)
+		}
+		concatColumnExpr := fmt.Sprintf("CONCAT(%s) AS %s", strings.Join(concatColumnNames, ","), joinedVarName)
```
typhoonzero, line 483: Why need a column stores the concated variable names?

go/codegen/optimize/codegen.go
sneaxiy: That's OK to only use `<=`. Only `=` -> `==` is necessary here.

go/executor/pai.go
sneaxiy: It is because the OptFlow HTTP interface only supports one column as the variable, and write the column value into the result table. Users need to see the result as:
| INDEX|VAR | X |
|---|---|---|
|0|a1, b1| xxx |
|1|a1, b2| xxx |
|2|a2, b1| xxx |
|3|a2, b2| xxx|

We need to change the implementation of this PR after we have refined OptFlow's interface.

typhoonzero (approved):

PR #2672, opened 2020-07-14
PAI submitter: add explain function

lhw362950217:
Related to #2656

### DIFF
--- a/python/runtime/pai/submitter.py
+++ b/python/runtime/pai/submitter.py
@@ -578,14 +578,14 @@ def create_explain_result_table(datasource, data_table, result_table,
     create_stmt = ""
     if model_type == model.MODEL_TYPE_TF:
         if estimator.startsWith("BoostedTrees"):
-            columnDef = ""
+            column_def = ""
             if conn.driver == "mysql":
-                columnDef = "(feature VARCHAR(255), dfc FLOAT, gain FLOAT)"
+                column_def = "(feature VARCHAR(255), dfc FLOAT, gain FLOAT)"
             else:
                 # Hive & MaxCompute
-                columnDef = "(feature STRING, dfc STRING, gain STRING)"
+                column_def = "(feature STRING, dfc STRING, gain STRING)"
             create_stmt = "CREATE TABLE IF NOT EXISTS %s %s;" % (result_table,
-                                                                 columnDef)
+                                                                 column_def)
         else:
             if not label_column:
                 raise SQLFlowDiagnostic(
--- a/python/runtime/pai/submitter.py
+++ b/python/runtime/pai/submitter.py
@@ -517,7 +517,7 @@ def submit_pytf_predict(datasource, select, result_table, label_column,
     # format resultTable name to "db.table" to let the codegen form a submitting
     # argument of format "odps://project/tables/table_name"
     project = get_project(datasource)
-    if result_table.count(".") == 1:
+    if result_table.count(".") == 0:
         result_table = "%s.%s" % (project, result_table)
 
     create_predict_result_table(datasource, data_table, result_table,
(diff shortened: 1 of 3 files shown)


python/runtime/pai/submitter.py
```diff
@@ -535,3 +535,161 @@ def submit_pytf_predict(datasource, select, result_table, label_column,
                               model_name, data_table, result_table, model_type,
                               cwd)
     submit_pai_task(cmd, datasource)
+    drop_tmp_tables([data_table], datasource)
```
sneaxiy, line 539: Maybe we should find a better way to drop the temporary tables if there are some errors raised between `create_predict_result_table` and `drop_tmp_tables`? For example, something like the `with` statement in Python?

python/runtime/pai/submitter.py
```diff
@@ -535,3 +535,161 @@ def submit_pytf_predict(datasource, select, result_table, label_column,
...
+
+
+def get_create_shap_result_sql(conn, data_table, result_table, label_column):
```
sneaxiy, line 542: Why is the 3rd parameter not `select` like the Go side code? It seems that you think the 3rd parameter as a select parameter, as you call this method [here](https://github.com/sql-machine-learning/sqlflow/pull/2672/files#diff-165c68c527ee5baf4aad24fe190bcb56R593).

python/runtime/pai/submitter.py
```diff
@@ -535,3 +535,161 @@ def submit_pytf_predict(datasource, select, result_table, label_column,
...
+    Args:
+        conn: a database connection
+        data_table: table name to read data from
+        result_table: result table name
+        label_column: column name of label 
+    
+    Returns:
+        a sql statement to create shap result table
+    """
+    schema = db.get_table_schema(conn, data_table)
```
sneaxiy, line 554: Seems that `get_table_schema` cannot get the schema from a select statement.

python/runtime/pai/submitter.py
```diff
@@ -535,3 +535,161 @@ def submit_pytf_predict(datasource, select, result_table, label_column,
...
+    drop_tmp_tables([data_table], datasource)
+
+
+def get_create_shap_result_sql(conn, data_table, result_table, label_column):
+    """Get a sql statement which create a result table for shap
+
+    Args:
+        conn: a database connection
+        data_table: table name to read data from
+        result_table: result table name
```
sneaxiy, line 548: Change the comment if the 3rd parameter should be `select`.

python/runtime/pai/submitter.py
```diff
@@ -535,3 +535,161 @@ def submit_pytf_predict(datasource, select, result_table, label_column,
...
+            columnDef = ""
+            if conn.driver == "mysql":
+                columnDef = "(feature VARCHAR(255), dfc FLOAT, gain FLOAT)"
+            else:
+                # Hive & MaxCompute
+                columnDef = "(feature STRING, dfc STRING, gain STRING)"
+            create_stmt = "CREATE TABLE IF NOT EXISTS %s %s;" % (result_table,
+                                                                 columnDef)
+        else:
+            if len(label_column) == 0:
```
sneaxiy: Try to use `if not label_column` ?

python/runtime/pai/submitter.py
```diff
@@ -535,3 +535,161 @@ def submit_pytf_predict(datasource, select, result_table, label_column,
...
+                raise SQLFlowDiagnostic(
+                    "need to specify WITH label_col=lable_col_name "
+                    "when explaining deep models")
+            create_stmt = get_create_shap_result_sql(conn, result_table,
+                                                     select, label_column)
+    elif model_type == model.MODEL_TYPE_XGB:
+        if len(label_column) == 0:
```
sneaxiy: Try to use `if not label_column` ?

python/runtime/pai/submitter.py
```diff
@@ -535,3 +535,161 @@ def submit_pytf_predict(datasource, select, result_table, label_column,
...
+                "need to specify WITH label_col=lable_col_name "
+                "when explaining xgboost models")
+        create_stmt = get_create_shap_result_sql(conn, result_table, select,
+                                                 label_column)
+    else:
+        raise SQLFlowDiagnostic(
+            "not supported modelType %d for creating Explain result table" %
+            model_type)
+
+    db.exec(conn, create_stmt)
```
sneaxiy: Check whether there is an error?

python/runtime/pai/submitter.py
```diff
@@ -535,3 +535,161 @@ def submit_pytf_predict(datasource, select, result_table, label_column,
...
+        result_table: result table name
+        label_column: name of the label column
+    
+    Returns:
+        a PAI cmd to explain the data using given model
+    """
+    # NOTE(typhoonzero): for PAI random forests predicting, we can not load the TrainStmt
+    # since the model saving is fully done by PAI. We directly use the columns in SELECT
+    # statement for prediction, error will be reported by PAI job if the columns not match.
+    if len(label_column) == 0:
```
sneaxiy: Same above.

python/runtime/pai/submitter.py
```diff
@@ -535,3 +535,161 @@ def submit_pytf_predict(datasource, select, result_table, label_column,
...
+        model_name: model used to do prediction
+        model_params: dict, Params for training, crossponding to WITH clause
+    """
+    params = locals()
+    params["entry_type"] = "explain"
+
+    cwd = os.getcwd()
+    # TODO(typhoonzero): Do **NOT** create tmp table when the select statement is like:
+    # "SELECT fields,... FROM table"
+    data_table = create_tmp_table_from_select(select, datasource)
```
sneaxiy, line 669: Check whether there is an error?

python/runtime/pai/submitter.py
```diff
@@ -535,3 +535,161 @@ def submit_pytf_predict(datasource, select, result_table, label_column,
...
+    elif model_type == model.MODEL_TYPE_XGB:
+        # (TODO:lhw) add XGB explain cmd
+        pass
+    else:
+        cmd = get_pai_tf_cmd(conf, JOB_ARCHIVE_FILE, PARAMS_FILE,
+                             ENTRY_DIR + "tensorflow.py", model_name,
+                             oss_model_path, data_table, "", result_table,
+                             project, cwd)
+    submit_pai_task(cmd, datasource)
+    drop_tmp_tables([data_table], datasource)
```
sneaxiy, line 701: Same above.

python/runtime/pai/submitter.py
lhw362950217: Let's do it in next PR, there are quite a few things to do like this to make the code more robust.

python/runtime/pai/submitter.py
lhw362950217: Thanks, here is right and the caller place is a bug, I meant to pass a input table and output table name to this function.

sneaxiy (approved):
Should add some unittests in the future PRs.

PR #2680, opened 2020-07-15
Make the attribute check for XGBoost model compatible with reg:linear objective function.

### LINKED ISSUE #2678: Using objective function reg:linear to train XGBoost model will fail because of attribute check.
I'm trying to train a XGBoost model on PAI using the following SQL statement.

```SQL
SELECT * FROM source_table
TO TRAIN xgboost.gbtree
WITH 
    objective ="reg:linear",
    train.num_boost_round = 30
LABEL target_amount
INTO sqlflow_regression_model;
```

It will fails because of attribute check. The error log is as follows:

```TXT
runSQLProgram error: attribute objective error: expected value in [reg:squaredlogerror multi:softmax multi:softprob rank:ndcg rank:pairwise reg:gamma binary:logitraw reg:logistic reg:tweedie reg:squarederror binary:hinge binary:logistic rank:map survival:cox], actual: reg:linear
```

The objective function `reg:linear` is deprecated in the versions after 0.8.2 and recommend to use `reg:squarederror`. 
But the XGBoost version on PAI is v0.8.2, we don't have `reg:squarederror` and can only use `reg:linear`. We need make the attribute check compatible with XGBoost v0.8.2

brightcoder01:
Fix #2678


Yancey0623 (dismissed):
LGTM

sneaxiy (dismissed):
LGTM++

shendiaomo (dismissed):

Yancey0623 (approved):

PR #2684, opened 2020-07-15
Fix commonThirdPartyCases ut

sneaxiy:
The `OdpsParserAdaptor` cannot parse `/*...*/`. Therefore, the unittest `commonThirdPartyCases ` would fail when using `OdpsParserAdaptor`.

### DIFF
--- a/go/parser/external/parser_test.go
+++ b/go/parser/external/parser_test.go
@@ -21,6 +21,27 @@ import (
 )
 
 func commonThirdPartyCases(p Parser, a *assert.Assertions) {
+	isMaxComputeParser := false
+	if jp, ok := p.(*javaParser); ok && jp.typ == "maxcompute" {
+		isMaxComputeParser = true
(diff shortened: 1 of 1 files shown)


go/parser/external/parser_test.go
```diff
@@ -21,6 +21,27 @@ import (
...
 
 func commonThirdPartyCases(p Parser, a *assert.Assertions) {
+	isMaxComputeParser := false
+	if jp, ok := p.(*javaParser); ok && jp.typ == "maxcompute" {
+		isMaxComputeParser = true
+	}
+
+	testSelectCases := make([]string, 0)
+	for _, sql := range SelectCases {
+		// NOTE: OdpsParserAdaptor cannot parse /*...*/
```
typhoonzero, line 31: Also add this to the PR description comment?

typhoonzero (approved):

PR #2687, opened 2020-07-16
Move oss operation functions to oss.py

typhoonzero:
The file `model.py` and `oss.py` basically does the same thing.


sneaxiy (approved):
LGTM.

PR #2688, opened 2020-07-16
Fix indent

weiguoz:
I was a little too late to review [ref](https://github.com/sql-machine-learning/sqlflow/pull/2685/files#diff-a5823644bb9fb3aca72238f4e758f912R59-R61), so I filed a tiny fix here.


sneaxiy (approved):
LGTM.

PR #2690, opened 2020-07-16
Remove duplicate codes in derivation.go

sneaxiy:
- Replace the `decomp` method in derivation.go with the `verifier.Decomp` method to avoid duplicate codes.
- Replace the `fieldTypes` in derivation.go with the `verifier.FieldTypes` to avoid duplicate codes.

### DIFF
--- a/go/ir/derivation.go
+++ b/go/ir/derivation.go
@@ -29,18 +29,6 @@ import (
 // TODO(typhoonzero): fieldTypes are copied from verifier.go, need refactor.
 type fieldTypes map[string]string
 
-// TODO(typhoonzero): decomp is copied from verifier.go, need to refactor the structure to avoid copying.
-func decomp(ident string) (tbl string, fld string) {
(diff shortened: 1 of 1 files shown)


shendiaomo (approved):

go/ir/derivation.go
```diff
@@ -277,7 +262,7 @@ func getMaxIndexOfKeyValueData(str string) (int, error) {
 
 func fillFieldDesc(columnTypeList []*sql.ColumnType, rowdata []interface{}, fieldDescMap FieldDescMap, rowCount int, originalSizes map[string]int) error {
 	for idx, ct := range columnTypeList {
-		_, fld := decomp(ct.Name())
+		_, fld := verifier.Decomp(ct.Name())
```
typhoonzero, line 265: `Decomp` is sort of database functionality, put it to database package should be more reasonable.

PR #2697, opened 2020-07-16
Add test case for PAI TF training

lhw362950217:
Related to #2656


python/runtime/pai/entry/tf.py
```diff
@@ -0,0 +1,74 @@
...
+import types
+from inspect import getargspec
+
+from runtime.pai import train
+from runtime.pai.pai_distributed import define_tf_flags, set_oss_environs
+from runtime.tensorflow import explain, is_tf_estimator, predict
+from runtime.tensorflow.diag import SQLFlowDiagnostic
+
+
+def call_fun(func, params):
```
typhoonzero, line 25: `call_func`?

By the way, I don't like this function, it's hard to read. Can we have generic runtime API like `runtime.pai.train(datasource, select, feature_columns, ...)` and call the API by go side code generation?

lhw362950217: This is an entry point function on PAI, programmer do not call this function directly. They call [submit_pytf_train](https://github.com/sql-machine-learning/sqlflow/blob/30c90f6e47135111c37eb28ca89c569b57fc202f/python/runtime/pai/submitter.py#L324) after feature derivation and other things. `submit_pytf_train` wrap params and resources for the entry function. On PAI ,the entry function unzip the resource and call specific train/predict/explain by this `call_fun`. It is a dispatcher, all param checks have already done in function like `submit_pytf_train`(of course, this `submit_pytf_train` function name may change according to runtime.pai.train). If we don't use this way, we just expand every params, and that is equivalent to this.

typhoonzero (approved):
Approve for merge this first and to do further polishment.

typhoonzero:
For more discussions: https://github.com/sql-machine-learning/sqlflow/issues/2699

PR #2698, opened 2020-07-16
Add selected_columns_and_types method

sneaxiy:
The method `selected_columns_and_types` would be used by the `verifier` module in the Python side after refactory.

### DIFF
--- a/python/runtime/db.py
+++ b/python/runtime/db.py
@@ -192,29 +192,102 @@ def read_feature(raw_val, feature_spec, feature_name):
         return raw_val,
 
 
-def selected_cols(conn, select):
+INT64_TYPE = long if six.PY2 else int
+LIMIT_PATTERN = re.compile("LIMIT\\s+[0-9]+", flags=re.I)
+BLANK_PATTERN = re.compile("\\s+")
+
+
+def limit_select(select, n):
+    """Make the select SQL statement with limited row number to query.
+
+    Args:
+        select (str): the select SQL statement.
+        n (int): the limited row number to query.
+
+    Returns:
+        If n >= 0, return a new SQL statement which would query n row(s) at most.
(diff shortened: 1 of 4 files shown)


python/runtime/db.py
```diff
@@ -192,29 +192,97 @@ def read_feature(raw_val, feature_spec, feature_name):
...
+    """
+    if n < 0:
+        return select
+
+    def replace_limit_num(matched_limit):
+        num = int(matched_limit.group(1))
+        return "LIMIT {}".format(min(num, n))
+
+    if LIMIT_PATTERN.search(select) is None:
+        return select + " LIMIT {}".format(n)
```
typhoonzero, line 217: If the statement got `;` at the end?

python/runtime/db.py
```diff
@@ -192,29 +192,97 @@ def read_feature(raw_val, feature_spec, feature_name):
...
+MYSQL_DATA_TYPE_DICT = {
+    1: "TINYINT",
+    3: "INT",
+    4: "FLOAT",
+    5: "DOUBLE",
+    8: "BIGINT",
+    246: "DECIMAL",
+    252: "TEXT",
+    253: "VARCHAR",
+    254: "CHAR",
```
lhw362950217, line 231: How about use MySQLdb.constants.FIELD_TYPE constants instead of the numbers?

typhoonzero (approved):
Cool!

python/runtime/db.py
sneaxiy: Thanks for your suggestions! Would fix it in the next PR.

python/runtime/db.py
sneaxiy: Thanks for your suggestions! Would fix it in the next PR.

PR #2703, opened 2020-07-17
Unify python datasource codes

sneaxiy:
Add a new file `runtime/testing.py` to unify the datasource settings used in the Python side (like `GetTestingDBSingleton` in the Go side). All Python codes should use the `runtime/testing.py` settings instead of hard-coding the datasources in each test.

### DIFF
--- a/python/runtime/testing.py
+++ b/python/runtime/testing.py
@@ -1,0 +1,108 @@
+# Copyright 2020 The SQLFlow Authors. All rights reserved.
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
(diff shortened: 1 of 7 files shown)


python/runtime/testing.py
```diff
@@ -0,0 +1,108 @@
...
+    if not project:
+        raise ValueError("SQLFLOW_TEST_DB_MAXCOMPUTE_PROJECT must be set")
+
+    endpoint = os.getenv("SQLFLOW_TEST_DB_MAXCOMPUTE_ENDPOINT",
+                         "http://service-maxcompute.com/api")
+
+    scheme = None
+    if endpoint.startswith("http://"):
+        scheme = "http"
+        endpoint = endpoint[len(scheme) + 2]
```
lhw362950217: Do not need this line?

python/runtime/testing.py
```diff
@@ -0,0 +1,108 @@
...
+    elif endpoint.startswith("https://"):
+        scheme = "https"
+
+    if scheme:
+        endpoint = endpoint[len(scheme) + 2:]
```
lhw362950217: +2 -> +3

python/runtime/testing.py
sneaxiy: Thanks. Removed.

python/runtime/testing.py
sneaxiy: Thanks! Fixed.

lhw362950217 (approved):
LGTM

PR #2705, opened 2020-07-17
Add docstrings for several functions to improve readability

shendiaomo:


python/runtime/seeding.py
```diff
@@ -15,5 +15,12 @@
 
 
 def get_tf_random_seed():
+    """
+    get_tf_random_seed returns an integer that can be used as a random seed
```
sneaxiy: How about:

```
get_tf_random_seed returns an integer from the environment variable SQLFLOW_TF_RANDOM_SEED, that can be used as a random seed.
```

sneaxiy (approved):
LGTM

PR #2719, opened 2020-07-21
PAI submitter for XGBoost training

lhw362950217:
Related to #2656

### DIFF
--- a/python/runtime/pai/submitter.py
+++ b/python/runtime/pai/submitter.py
@@ -628,7 +628,7 @@ def create_explain_result_table(datasource, data_table, result_table,
 
     create_stmt = ""
     if model_type == oss.MODEL_TYPE_TF:
-        if estimator.startsWith("BoostedTrees"):
+        if estimator.startswith("BoostedTrees"):
             column_def = ""
             if conn.driver == "mysql":
                 column_def = "(feature VARCHAR(255), dfc FLOAT, gain FLOAT)"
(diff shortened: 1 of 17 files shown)


python/runtime/pai/submitter.py
```diff
@@ -41,6 +41,26 @@
...
+def get_requirement(model_name):
+    """Get required python package according to estimator name
+
+    Args:
+        estimator: name of the model
+
+    Returns:
+        A string with multilines, each line is a requirement
+    """
+    if model_name.lower() == "xgboost":
```
sneaxiy: How about `model_name.lower().startswith("xgboost")`? The `model_name` may be in the form of `xgboost.gbtree`.

python/runtime/pai/submitter.py
```diff
@@ -381,7 +406,11 @@ def submit_pai_tf_train(datasource, estimator_string, select,
     params = dict(locals())
     del params["train_params"]
     params.update(train_params)
-    params["entry_type"] = "train"
+
+    if estimator_string.lower() == "xgboost":
```
sneaxiy: Same above.

python/runtime/pai/xgboost/train.py
```diff
@@ -0,0 +1,113 @@
...
+                   is_pai=True,
+                   pai_train_table=pai_table,
+                   pai_validate_table=pai_val_table,
+                   oss_model_dir=oss_model_dir,
+                   transform_fn=transform_fn,
+                   feature_column_code=feature_columns_code,
+                   model_repo_image=model_repo_image,
+                   original_sql=original_sql)
+    else:
+        local_train(datasource=datasource,
```
sneaxiy, line 92: Where is the definition of `local_train`?

lhw362950217: It is an alias : from runtime.xgboost.train import train as local_train

sneaxiy (approved):
LGTM++

PR #2721, opened 2020-07-22
Remove comments in the SQL statement

sneaxiy:
Remove single line (--) and multiple line (/**/) comments in the SQL statement before running.

### DIFF
--- a/go/parser/sqlflow_parser_test.go
+++ b/go/parser/sqlflow_parser_test.go
@@ -308,3 +308,35 @@ func TestParserErrorMessage(t *testing.T) {
 	a.Equal(-1, idx)
 	a.True(strings.Contains(e.Error(), `near or before "select b f`))
 }
+
+func TestRemoveCommentInSQLStatement(t *testing.T) {
+	testFunc := func(inputSQL, expectedOutputSQL string, noError bool) {
+		actualOutputSQL, err := RemoveCommentInSQLStatement(inputSQL)
+		if noError {
+			assert.NoError(t, err)
+			assert.Equal(t, expectedOutputSQL, actualOutputSQL)
+			return
+		}
+		assert.Error(t, err)
+	}
+
+	testFunc(`SELECT * FROM a;`, `SELECT * FROM a;`, true)
+	testFunc(`SELECT * FROM a -- I am a comment`, `SELECT * FROM a `, true)
+	testFunc("SELECT * FROM a -- I am a comment\n TO TRAIN b WITH -- 2nd comment\n INTO c",
+		"SELECT * FROM a \n TO TRAIN b WITH \n INTO c", true)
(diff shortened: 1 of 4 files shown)


go/parser/sqlflow_parser_test.go
```diff
@@ -308,3 +308,35 @@ func TestParserErrorMessage(t *testing.T) {
...
+
+	testFunc(`SELECT * FROM a;`, `SELECT * FROM a;`, true)
+	testFunc(`SELECT * FROM a -- I am a comment`, `SELECT * FROM a `, true)
+	testFunc("SELECT * FROM a -- I am a comment\n TO TRAIN b WITH -- 2nd comment\n INTO c",
+		"SELECT * FROM a \n TO TRAIN b WITH \n INTO c", true)
+	testFunc("SELECT '--' FROM a -- I am a comment\n TO TRAIN b WITH -- 2nd comment\n INTO c",
+		"SELECT '--' FROM a \n TO TRAIN b WITH \n INTO c", true)
+	testFunc("SELECT \"abc -- FROM a -- I am a comment\n TO TRAIN b WITH -- 2nd comment\n\" INTO c",
+		"SELECT \"abc -- FROM a -- I am a comment\n TO TRAIN b WITH -- 2nd comment\n\" INTO c", true)
+	testFunc("SELECT \"abc\\\" -- comment\" FROM b", "SELECT \"abc\\\" -- comment\" FROM b", true)
```
typhoonzero, line 331: Field names are not allowed to be like `abc\" -- comment`, this test case may be confusing, maybe change to `select * from t where f="...."`

typhoonzero (approved):
LGTM

go/parser/sqlflow_parser_test.go
sneaxiy: The `abc\" -- comment` here is not a field name but a string literal.

I have tried this SQL statement (`SELECT "abc\" --comment"`) in MySQL, and it works well.

```
mysql> SELECT "abc\" --comment";
+----------------+
| abc" --comment |
+----------------+
| abc" --comment |
+----------------+
1 row in set (0.00 sec)
```

typhoonzero: `SELECT "abc\" --comment"` is OK yet `SELECT "abc\" --comment" FROM sometable` may cause error.

sneaxiy: `SELECT "abc\" --comment" FROM iris.train` is also OK in MySQL.

```
mysql> SELECT "abc\" --comment" FROM iris.train LIMIT 2;
+----------------+
| abc" --comment |
+----------------+
| abc" --comment |
| abc" --comment |
+----------------+
2 rows in set (0.00 sec)
```

typhoonzero: OK, cool.

PR #2727, opened 2020-07-22
Move the optimize Go code to Python code

### LINKED ISSUE #2725: Cleanup optimize codes
We have 2 kinds of optimize codes:

- one in the Python side (`optimize.py`), which is only used for the open source cases.
- one in the Go side (`optimize/codegen.go`), which is only used for the OptFlow cases.

The reason why we have these 2 kinds of codes is that we needed to get the whole table data when there was any `GROUP BY` in the `CONSTRAINT` clauses if we used OptFlow in the past. Because we have `goalisa` but not `pyalisa`, we had to get the whole table data in the Go side instead of in the Python side which we use OptFlow.

However, after https://github.com/sql-machine-learning/sqlflow/pull/2670 , we only need the table schema but not the whole table data. We need to unify these 2 kinds of codes into one during the refactory process.

sneaxiy:
Fix https://github.com/sql-machine-learning/sqlflow/issues/2725 .


python/runtime/optimize/model_generation.py
```diff
@@ -0,0 +1,624 @@
+# Copyright 2020 The SQLFlow Authors. All rights reserved.
```
typhoonzero, line 1: How about refactoring to:

1. generate a grammar tree (AST) from the SQLFlow expression tokens
2. transverse the grammar tree and generate corresponding expressions for pyomo or optflow.

Can we make the code simpler by using this way?

typhoonzero (approved):
LGTM with one suggestion

python/runtime/optimize/model_generation.py
sneaxiy: I have thought about this way as well. It may be simpler and more standard. I think that it would be more reasonable if we transform the SQLFlow Go AST to be the final model code directly, instead of transforming the SQLFlow Go AST to tokens and then parsing the tokens to another AST in the Python side. I would think about this idea. The related issue is https://github.com/sql-machine-learning/sqlflow/issues/2531 .

Issue #2568, opened 2020-06-29
Implement feature derivation and verifier in Python


shendiaomo:
As the discussion of #2494 , SQLFlow would be made a pure compiler. This requires we support feature derivation/verifier and columns in Python. There's a reference implementation in https://github.com/sql-machine-learning/sqlflow/pull/2477/files#diff-606462e9e98c6c23cdf010f1ec243c0e .
- [x] Implement feature derivation in Python (Add new module). https://github.com/sql-machine-learning/sqlflow/pull/2735, https://github.com/sql-machine-learning/sqlflow/pull/2739
- [x] Implement verifier in Python (Add new modules). https://github.com/sql-machine-learning/sqlflow/pull/2706, https://github.com/sql-machine-learning/sqlflow/pull/2708
- [ ] Modify the interface of the Python package `sqlflow.runtime` to call the new Python modules above instead of the existing Go feature derivation and verifier code.
- [ ] Remove the Go feature derivation and verifier code.

PR #2735, opened 2020-07-23
Add feature_column.py

sneaxiy:
c.f. https://github.com/sql-machine-learning/sqlflow/issues/2568

### DIFF
--- a/python/runtime/feature_derivation/feature_column.py
+++ b/python/runtime/feature_derivation/feature_column.py
@@ -11,65 +11,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import json
-
 import six
-
-INT = 0
-FLOAT = 1
-STRING = 2
-
-
-class FieldDesc(object):
-    """
-    FieldDesc describes a field used as the input to a feature column.
-
-    Args:
-        name (str): the field name. Default "".
-        dtype (int): the data type of the field. It must be one of INT,
-            FLOAT and STRING. Default INT.
-        delimiter (str): the delimiter of the field data. Default "".
-        format (str): the format of the field data. Default "".
-        shape (list[int]): the shape of the field data. Default None.
-        is_sparse (bool): whether the field data is sparse. Default False.
-        vocabulary (list[str]): the vocabulary used for categorical feature column. Default None.
-        max_id (int): the maximum id number of the field data. Used in CategoryIDColumn. Default 0.
-    """
-    def __init__(self,
-                 name="",
-                 dtype=INT,
-                 delimiter="",
-                 format="",
-                 shape=None,
-                 is_sparse=False,
-                 vocabulary=None,
-                 max_id=0):
-        self.name = name
-        self.dtype = dtype
-        self.delimiter = delimiter
-        self.format = format
-        self.shape = shape
-        self.is_sparse = is_sparse
-        self.vocabulary = vocabulary
-        self.max_id = max_id
-
-    def to_json(self):
-        """
-        Convert the FieldDesc object to a json string.
-
-        Returns:
-            A string which represents the json value of the FieldDesc object.
-        """
-        return json.dumps({
-            "name": self.name,
-            "dtype": self.dtype,
-            "delimiter": self.delimiter,
-            "format": self.format,
-            "shape": self.shape,
-            "is_sparse": self.is_sparse,
-            "vocabulary": self.vocabulary,
-            "max_id": self.max_id,
-        })
+from runtime.feature_derivation.field_desc import DataType, FieldDesc
 
 
 class FeatureColumn(object):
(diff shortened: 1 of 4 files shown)


python/runtime/feature_derivation/feature_column.py
```diff
@@ -0,0 +1,512 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import json
+
+import six
+
+INT = 0
```
brightcoder01: INT, FLOAT, STRING looks like an enumerator. How about wrap them into a class.

```Python
class DataType:
    INT = 0,
    FLOAT = 1,
    STRING = 2
```

python/runtime/feature_derivation/feature_column.py
```diff
@@ -0,0 +1,512 @@
...
+
+        Returns:
+            A list of the FieldDesc objects.
+        """
+        if self.category_column is None:
+            return [FieldDesc()]
+
+        return self.category_column.get_field_desc()
+
+    def apply_to(self, field_desc):
```
brightcoder01: It may be confusing that `apply_to` returns a FeatureColumn object.

python/runtime/feature_derivation/feature_column.py
```diff
@@ -0,0 +1,512 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import json
+
+import six
+
+INT = 0
+FLOAT = 1
```
brightcoder01: I'm wondering why we don't have double type?

python/runtime/feature_derivation/feature_column.py
typhoonzero: Add comment here to describe how these types will be used.

python/runtime/feature_derivation/feature_column.py
```diff
@@ -0,0 +1,512 @@
...
+        shape (list[int]): the shape of the field data. Default None.
+        is_sparse (bool): whether the field data is sparse. Default False.
+        vocabulary (list[str]): the vocabulary used for categorical feature column. Default None.
+        max_id (int): the maximum id number of the field data. Used in CategoryIDColumn. Default 0.
+    """
+    def __init__(self,
+                 name="",
+                 dtype=INT,
+                 delimiter="",
+                 format="",
```
brightcoder01: Can we have an enumerator for format?

python/runtime/feature_derivation/feature_column.py
```diff
@@ -0,0 +1,512 @@
...
+        object holds.
+
+        Returns:
+            A list of the FieldDesc objects, whose length is 1.
+        """
+        return [self.field_desc]
+
+    def apply_to(self, field_desc):
+        """
+        Create a new CategoryIDColumn object that holds the given
```
typhoonzero: Maybe we can only keep the base class doc string?

python/runtime/feature_derivation/feature_column.py
```diff
@@ -0,0 +1,512 @@
...
+import json
+
+import six
+
+INT = 0
+FLOAT = 1
+STRING = 2
+
+
+class FieldDesc(object):
```
typhoonzero: Put `FieldDesc` in a separated file.

python/runtime/feature_derivation/feature_column.py
sneaxiy: Done.

python/runtime/feature_derivation/feature_column.py
sneaxiy: Great suggestions! Change it to be `new_feature_column_from`.

python/runtime/feature_derivation/feature_column.py
sneaxiy: Done.

python/runtime/feature_derivation/feature_column.py
sneaxiy: Great idea! Done.

python/runtime/feature_derivation/feature_column.py
sneaxiy: Done.

python/runtime/feature_derivation/feature_column.py
sneaxiy: Done.

python/runtime/feature_derivation/feature_column.py
sneaxiy: Machine learning usually uses float32 instead of float64 as the major floating-point type. Maybe we can add `DOUBLE` if needed in the future.

typhoonzero (approved):

brightcoder01 (approved):

PR #2738, opened 2020-07-24
Fix feature derivation bug

sneaxiy:
Fix bugs:
- `TestFeatureDerivation` should connect Hive db instead of MySQL db if `SQLFLOW_TEST_DB=hive`.
- The `size` calculation in `fillCSVFieldDesc` is wrong. The `size` was always 0 in the previous implementation.


typhoonzero (approved):
Cool, thanks for the fix

PR #2739, opened 2020-07-25
Move feature derivation code from Go to Python

sneaxiy:
c.f. https://github.com/sql-machine-learning/sqlflow/issues/2568

### DIFF
--- a/python/runtime/feature_derivation/feature_derivation.py
+++ b/python/runtime/feature_derivation/feature_derivation.py
@@ -1,0 +1,630 @@
+# Copyright 2020 The SQLFlow Authors. All rights reserved.
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+__all__ = [
+    'infer_feature_columns',
+]
+
+import re
+
+import numpy as np
+import six
+from runtime.feature_derivation.feature_column import (CategoryIDColumn,
+                                                       EmbeddingColumn,
+                                                       FeatureColumn,
+                                                       IndicatorColumn,
+                                                       NumericColumn)
+from runtime.feature_derivation.field_desc import (DataFormat, DataType,
+                                                   FieldDesc)
+from runtime.verifier import fetch_samples
+
+
+def init_column_map(target_fc_map, fc):
+    """
+    Init the target_fc_map by the feature column fc.
+
+    Args:
+        target_fc_map (dict[str -> FeatureColumn): the feature column map,
+            where the key is the field name.
+        fc (FeatureColumn): the feature column object.
+
+    Returns:
+        None.
+    """
+    if isinstance(fc, (EmbeddingColumn, IndicatorColumn)) and \
+        len(fc.get_field_desc()) == 0:
+        if fc.name not in target_fc_map:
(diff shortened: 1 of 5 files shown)


typhoonzero (approved):

typhoonzero (commented):
Try keep the files short.

python/runtime/feature_derivation/feature_derivation.py
```diff
@@ -0,0 +1,630 @@
+# Copyright 2020 The SQLFlow Authors. All rights reserved.
```
typhoonzero, line 1: maybe keep the package strucure: `runtime.feature.derivation`?

sneaxiy: Great idea! Would rename in the next PR.

python/runtime/feature_derivation/feature_derivation.py
```diff
@@ -0,0 +1,630 @@
...
+        float_value = float(cell)
+    except ValueError:
+        float_value = None
+
+    if float_value is None:
+        field_desc.dtype = DataType.STRING
+        field_desc.shape = [1]
+        if field_desc.vocabulary is None:
+            field_desc.vocabulary = set()
+        field_desc.vocabulary.add(cell)
```
brightcoder01, line 269: Do we build this vocabulary from the sample data?

python/runtime/feature_derivation/feature_derivation.py
```diff
@@ -0,0 +1,630 @@
...
+
+    Returns:
+        A FieldDesc object whose name is the given name,
+        and the data type is INT.
+    """
+    return FieldDesc(name=name, dtype=DataType.INT)
+
+
+# A regular expression to match any real number
+REAL_NUMBER_PATTERN = re.compile(
```
brightcoder01, line 120: Can we put this Regex expressions into a separate file?

python/runtime/feature_derivation/feature_derivation.py
```diff
@@ -0,0 +1,630 @@
...
+                                           len(field_desc.vocabulary))
+        # NOTE(typhoonzero): a default embedding size of 128 is enough for most cases.
+        embedding = EmbeddingColumn(category_column=category_column,
+                                    dimension=128,
+                                    combiner="sum")
+        embedding.name = field_desc.name
+        return embedding
+
+
+def derive_feature_column(targets, fc_map, fd_map, selected_field_names,
```
brightcoder01, line 405: derive_feature_columns?

python/runtime/feature_derivation/feature_derivation.py
```diff
@@ -0,0 +1,630 @@
...
+                new_fc_target_map[selected_field_name] = [new_fc]
+
+        fc_target_map.clear()
+        fc_target_map.update(new_fc_target_map)
+
+
+def transform_derived_feature_column_map_to_ir_feature_column_map(
+    features, fc_map, selected_field_names, label_name):
+    """
+    Transform the derived FeatureColumn map to the FeatureColumn
```
brightcoder01, line 487: Can we add more description about what transform is done in this method?
From the comments, we can see that the types of input and output are the same. It may be reader friendly if we have the description above.

sneaxiy: Added. Please see https://github.com/sql-machine-learning/sqlflow/pull/2744 .

python/runtime/feature_derivation/feature_derivation.py
sneaxiy: Renamed. Please see https://github.com/sql-machine-learning/sqlflow/pull/2744 .

python/runtime/feature_derivation/feature_derivation.py
sneaxiy: It is only used inside this file currently. We can remove it to a separate file when it is needed in other places in the future.

python/runtime/feature_derivation/feature_derivation.py
sneaxiy: Yes.

PR #2748, opened 2020-07-27
Add token validation check to math programming

sneaxiy:
Add error check to the math programming `objective` and `constraint` as early as possible.

### DIFF
--- a/python/runtime/optimize/model_generation.py
+++ b/python/runtime/optimize/model_generation.py
@@ -679,7 +679,7 @@ def generate_objective_and_constraint_expression(columns,
             group_by = c.get("group_by")
 
             assert_are_valid_tokens(columns=columns,
-                                    tokens=objective,
+                                    tokens=tokens,
                                     result_value_name=result_value_name,
                                     group_by=group_by)
             expr, for_range, iter_vars = generate_objective_or_constraint_expression(
--- a/python/runtime/optimize/model_generation.py
+++ b/python/runtime/optimize/model_generation.py
@@ -12,11 +12,59 @@
 # limitations under the License.
 
 import copy
+import re
 
 __all__ = [
     'generate_unique_result_value_name',
     'generate_objective_and_constraint_expression',
 ]
+
+IDENTIFIER = re.compile("[_a-zA-Z]\w*")
+
+
+def is_identifier(token):
+    return IDENTIFIER.fullmatch(token) is not None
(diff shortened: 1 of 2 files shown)


python/runtime/optimize/model_generation.py
```diff
@@ -12,12 +12,56 @@
...
+    Raises:
+        AssertionError if any token is invalid.
+    """
+    valid_columns = [c.lower() for c in columns]
+
+    if group_by and group_by.lower() not in valid_columns:
+        raise AssertionError("GROUP BY column %s not found" % group_by)
+
+    if not tokens:
+        return
```
typhoonzero: No tokens should raise error?

python/runtime/optimize/model_generation_test.py
```diff
@@ -72,12 +86,17 @@ def replace_objective_token(self, objective, old, new):
 
     def replace_constraint_token(self, constraint, old, new):
         def replace_one_constraint(c):
-            c = copy.copy(c)
-            for i, token in enumerate(c["tokens"]):
+            ret = {
```
typhoonzero: use `copy.deepcopy`?

python/runtime/optimize/model_generation_test.py
```diff
@@ -18,8 +18,22 @@
...
+        return IDENTIFIER_REGEX.fullmatch(token) is not None
+
+    def test_main(self):
+        tokens = ['a', '_', 'a123', '__', '_123']
+        for t in tokens:
+            self.assertTrue(self.is_identifier(t))
+
+        tokens = ['1', '123_', '3def']
+        for t in tokens:
+            self.assertFalse(self.is_identifier(t))
```
typhoonzero, line 37: Add test case for `assert_are_valid_tokens`?

python/runtime/optimize/model_generation.py
sneaxiy: Done.

python/runtime/optimize/model_generation_test.py
sneaxiy: Great idea! Done.

python/runtime/optimize/model_generation_test.py
sneaxiy: Done.

typhoonzero (approved):

PR #2770, opened 2020-07-29
Add model zoo to playground

### LINKED ISSUE #2769: Add Model Zoo to playground
We need to add Model Zoo to SQLFlow playground to complete the ecosystem. After that, we can release model/repo on the playground.

**Additional Notes**

lhw362950217:
Fixes #2769

### DIFF
--- a/doc/run/k8s/install-sqlflow.yaml
+++ b/doc/run/k8s/install-sqlflow.yaml
@@ -81,13 +81,13 @@ spec:
       privileged: true
     command: [ "/bin/bash", "-c" ]
     args:
-      - |
-        while ! nc -z -w 1 ${MYSQL_ADDR/:/ }; do
-          sleep 3
-          echo "Waiting for mysql..."
-        done
-        echo "Starting SQLFlow Model Zoo server ..."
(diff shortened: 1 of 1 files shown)


doc/run/k8s/install-sqlflow.yaml
```diff
@@ -64,3 +64,33 @@ spec:
...
+    imagePullPolicy: IfNotPresent
+    ports:
+    - containerPort: 50055
+      protocol: TCP
+    env:
+    - name: MYSQL_ADDR
+      value: 127.0.0.1:3306
+    volumeMounts:
+    # bind to host's docker.sock to build image
+    - mountPath: /var/run/docker.sock
```
typhoonzero, line 78: If the model zoo server runs in k8s, it will use kaniko to build image automatically.

lhw362950217: This config is for single-user playground only. Using local docker is to avoid introducing new image dependency and keep the vm small.

And, thanks, I will use the `kaniko` way in our cloud playground.

typhoonzero: I see.

typhoonzero (approved):

PR #2772, opened 2020-07-29
Add feature column compilation

sneaxiy:
Add `python/runtime/feature/compile.py` to compile the IR feature column map to runtime feature column map (i.e., call the TensorFlow or XGBoost feature column APIs). It is similar to the Go function `GenerateFeatureColumnCode`, but it returns the runtime feature column object instead of the string code.

### DIFF
--- a/python/runtime/feature/compile.py
+++ b/python/runtime/feature/compile.py
@@ -1,0 +1,164 @@
+# Copyright 2020 The SQLFlow Authors. All rights reserved.
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import numpy as np
+import six
+from runtime.feature.column import (BucketColumn, CategoryHashColumn,
+                                    CategoryIDColumn, CrossColumn,
+                                    EmbeddingColumn, IndicatorColumn,
+                                    NumericColumn, SeqCategoryIDColumn)
+from runtime.feature.field_desc import DataType
+from runtime.model import EstimatorType
+
+__all__ = [
+    'compile_ir_feature_columns',
+]
+
+
+def to_package_dtype(dtype, package):
+    """
+    Convert dtype to the data type accepted by package.
+
+    Args:
+        dtype (DataType): one of INT, FLOAT and STRING.
+        package (module): the Python package.
+
+    Returns:
+        The data type accepted by the package.
+    """
+    if dtype == DataType.INT:
+        return package.dtypes.int64
+
+    if dtype == DataType.FLOAT:
+        return package.dtypes.float32
+
+    if dtype == DataType.STRING:
+        return package.dtypes.string
+
+    raise ValueError("unsupported data type {}".format(dtype))
+
+
+def compile_feature_column(fc, model_type, package):
+    """
+    Compile a IR FeatureColumn object to a runtime feature column object.
+
+    Args:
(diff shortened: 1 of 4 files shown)


python/runtime/feature/compile.py
```diff
@@ -0,0 +1,164 @@
...
+                raise ValueError(
+                    "field in CROSS must be of FeatureColumn or string type")
+
+        return fc_package.crossed_column(key_strs,
+                                         hash_bucket_size=fc.hash_bucket_size)
+
+    if isinstance(fc, EmbeddingColumn):
+        assert model_type != EstimatorType.XGBOOST, \
+            "EMBEDDING is not supported in XGBoost models"
+        cc = compile_feature_column(fc.category_column, model_type, package)
```
brightcoder01: What does cc stand for?

python/runtime/feature/compile.py
```diff
@@ -0,0 +1,164 @@
...
+    raise ValueError("unsupported data type {}".format(dtype))
+
+
+def compile_feature_column(fc, model_type, package):
+    """
+    Compile a IR FeatureColumn object to a runtime feature column object.
+
+    Args:
+        fc (FeatureColumn): the IR FeatureColumn object.
+        model_type (EstimatorType): one of TENSORFLOW and XGBOOST.
```
brightcoder01, line 60: model_type -> engine_type？

python/runtime/feature/compile.py
```diff
@@ -0,0 +1,164 @@
...
+from runtime.model import EstimatorType
+
+__all__ = [
+    'compile_ir_feature_columns',
+]
+
+
+def to_package_dtype(dtype, package):
+    """
+    Convert dtype to the data type accepted by package.
```
typhoonzero: by package -> by the feature column implementation packages including `tensorflow` and `xgboost`.

python/runtime/feature/compile.py
```diff
@@ -0,0 +1,164 @@
...
+
+    Args:
+        dtype (DataType): one of INT, FLOAT and STRING.
+        package (module): the Python package.
+
+    Returns:
+        The data type accepted by the package.
+    """
+    if dtype == DataType.INT:
+        return package.dtypes.int64
```
typhoonzero, line 43: Can we have a universal package dtype?

python/runtime/feature/compile.py
```diff
@@ -0,0 +1,164 @@
...
+
+    raise ValueError("unsupported data type {}".format(dtype))
+
+
+def compile_feature_column(fc, model_type, package):
+    """
+    Compile a IR FeatureColumn object to a runtime feature column object.
+
+    Args:
+        fc (FeatureColumn): the IR FeatureColumn object.
```
brightcoder01: How about marking fc specifically to `fc_ir`

python/runtime/feature/compile.py
sneaxiy: Done.

python/runtime/feature/compile.py
sneaxiy: A bit hard. 

- If we have a universal package dtype and it depends on a framework (say, in the form of `tf.dtypes.xxx`), we have to convert it to XGBoost-compatible dtype when training the XGBoost model. 
- If we have a universal package dtype that is independent of any framework (the current implementation), we have to convert it to the framework related dtype when training.

In a word, it seems that the conversation is inevitable no matter whether we have a universal package dtype.

python/runtime/feature/compile.py
sneaxiy: I have replaced `cc` with `category_column` for a better review.

python/runtime/feature/compile.py
sneaxiy: Done. Renamed to `ir_fc`.

python/runtime/feature/compile.py
sneaxiy: To be compatible with the naming rules in https://github.com/sql-machine-learning/sqlflow/blob/develop/python/runtime/pai/submitter.py#L518 , I prefer to use `model_type` here.

typhoonzero (approved):
LGTM++

brightcoder01 (approved):

PR #2778, opened 2020-07-30
Update OptFlow FSL generation code when len(variables) == 2

sneaxiy:
To satisfy the requirements of the OptFlow API.

### DIFF
--- a/go/executor/pai.go
+++ b/go/executor/pai.go
@@ -476,7 +476,7 @@ func executeOptimizeUsingOptFlow(pythonExecutor *pythonExecutor, stmt *ir.Optimi
 		dropTmpTables([]string{table}, pythonExecutor.Session.DbConnStr)
 	}
 
-	if len(stmt.Variables) > 1 {
+	if len(stmt.Variables) > 2 {
 		joinedVarName := strings.Join(stmt.Variables, "__")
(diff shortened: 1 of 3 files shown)


go/executor/pai.go
```diff
@@ -476,7 +476,7 @@ func executeOptimizeUsingOptFlow(pythonExecutor *pythonExecutor, stmt *ir.Optimi
 		dropTmpTables([]string{table}, pythonExecutor.Session.DbConnStr)
 	}
 
-	if len(stmt.Variables) > 1 {
+	if len(stmt.Variables) > 2 {
```
typhoonzero, line 479: Should it be `>=2` ?

sneaxiy: No. It should be `len(stmt.Variables) > 2`. The code inside the `if` statement is to create a table with a concatenated variable column. For example, if `variables=plants,markets`, an additional column named `plants__markets` is created and the cell value of this column is the concatenated cell values of the column `plants` and `market`. OptFlow has supported `len(stmt.Variables) == 2`, and we do not need to concatenate the variable column when `len(stmt.Variables) == 2`.

typhoonzero: I see.

typhoonzero (approved):

PR #2779, opened 2020-08-03
Support download model to local in cli

typhoonzero:


go/cmd/sqlflow/main.go
```diff
@@ -460,11 +463,18 @@ func processOptions(opts *options) {
...
+		filename, err = downloadModelFromDB(opts)
 	default:
 		err = runSQLFlowClient(opts)
 	}
 	if err != nil {
 		log.Printf("Failed due to %v", err)
+	} else {
+		// output downloaded model path if no error.
+		if opts.Get && opts.Model {
+			fmt.Printf("model \"%s\" downloaded successfully at %s\n", opts.ModelName, filename)
```
lhw362950217: How about put this piece of code in `downloadModelFromDB`, so the `processOptions` only do dispatch and don't do anything else.

typhoonzero: Good point.

lhw362950217 (approved):
LGTM

PR #2741, opened 2020-07-26
Add cora dataset

Derek-Wds:
Hi, this is a initial PR for adding the Cora dataset related to [#2714 ](https://github.com/sql-machine-learning/sqlflow/issues/2714).

The node table is created as:
```sql
CREATE TABLE cora.node (
                    id INT,
                    node_name TEXT,
                    features  TEXT,
                    label TEXT);
```
and the edge table is created as:
```sql
CREATE TABLE cora.edge (
                    id INT,
                    from_node_id INT,
                    to_node_id  INT,
                    weight FLOAT);
```

I have tested on the mysql server on my own machine, and I was able to use this [api](https://github.com/sql-machine-learning/sqlflow/blob/develop/python/runtime/db.py#L351) from SQLFlow to query and load the data.
```sql
SELECT from_node_id, to_node_id, weight, cora.node.features 
FROM cora.edge 
LEFT JOIN cora.node ON cora.edge.from_node_id = cora.node.id                                                                                                                                      
```
If there has any problem, please let me know. Thanks!

PR #2780, opened 2020-08-03
Modify cora dataset to adapt csv format

lhw362950217:
Our feature derivation accept csv format, but no space between fields.

Releated to #2741


sneaxiy (commented):
How about making our feature derivation code to receive the csv strings with blanks?

lhw362950217:
> How about making our feature derivation code to receive the csv strings with blanks?

Spaces in `.csv` file can be ambiguous, c.f. https://stackoverflow.com/questions/46228858/spaces-between-separator-and-value-in-csv-file. Maybe we can keep it simple this time, just wipe them out.

sneaxiy (approved):

Issue #2781, opened 2020-08-03
Unify db api


lhw362950217:
**Is your feature request related to a problem? Please describe.**
Current db.py make so many special check and branch to adapt different DB engines. We need to unify the DB api interface according to this [doc](https://www.python.org/dev/peps/pep-0249/). Especially for maxcompute.

**Describe the solution you'd like**
1. Modify maxcompute.py to adapt db api
2. Provide a query interface in the db.py

**Describe alternatives (optional)**

**Additional Notes**

sneaxiy:
@lhw362950217 @typhoonzero @weiguoz I wonder should we make a common base class like this?

```
class Connection:
    def execute(self, sql): # return affected rows 
        raise NotImplementedError()

    def query(self, sql):  # return a generator which yields each row value
        raise NotImplementedError()

class MySQLConnection(Connection):
    def execute(self, sql): ...
    def query(self, sql): ...

class HiveConnection(Connection):
    def execute(self, sql): ...
    def query(self, sql): ...

class MaxComputeConnection(Connection):
    def execute(self, sql): ...
    def query(self, sql): ...

class AlisaConnection(Connection):
    def execute(self, sql): ...
    def query(self, sql): ...
```

In this way, we can remove so many if-else. And it would be easy when we need to support a new DBMS.

lhw362950217:
> @lhw362950217 @typhoonzero @weiguoz I wonder should we make a common base class like this?
> 
> ```
> class Connection:
>     def execute(self, sql): # return affected rows 
>         raise NotImplementedError()
> 
>     def query(self, sql):  # return a generator which yields each row value
>         raise NotImplementedError()
> 
> class MySQLConnection(Connection):
>     def execute(self, sql): ...
>     def query(self, sql): ...
> 
> class HiveConnection(Connection):
>     def execute(self, sql): ...
>     def query(self, sql): ...
> 
> class MaxComputeConnection(Connection):
>     def execute(self, sql): ...
>     def query(self, sql): ...
> 
> class AlisaConnection(Connection):
>     def execute(self, sql): ...
>     def query(self, sql): ...
> ```
> 
> In this way, we can remove so many if-else. And it would be easy when we need to support a new DBMS.

Yes，I intend to do this, as described in point 2. As for hive & mysql, they already support DB-API specification, so they basically can share the same base class. As for `maxcompute`, we need to define a new class and customize some function to do this. Thanks for your advise for the OOP pattern.

PR #2782, opened 2020-08-03
Add query api to db.py

lhw362950217:
Related to #2781, we first add a function named `query`. Then we will change `maxcompute.py` to add adapt the `cursor` interface.

### DIFF
--- a/python/runtime/db_test.py
+++ b/python/runtime/db_test.py
@@ -419,7 +419,7 @@ class TestQuery(TestCase):
         rows = [row for row in gen()]
         self.assertEqual(1, len(rows))
 
-        query(conn, "delete from A")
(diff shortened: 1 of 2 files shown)


python/runtime/db_test.py
```diff
@@ -403,5 +403,32 @@ def test_limit_select(self):
                          limit_select("SELECT * FROM t \t ; ", 4))
 
 
+@unittest.skipIf(testing.get_driver() == "maxcompute", "skip non mysql tests")
```
sneaxiy, line 406: NIT. "skip non mysql and hive tests"?

sneaxiy (approved):
LGTM basically.

weiguoz (approved):
LGTM.

PR #2787, opened 2020-08-04
DB api base class

lhw362950217:
Related to #2781

### DIFF
--- a/python/runtime/dbapi/connection.py
+++ b/python/runtime/dbapi/connection.py
@@ -1,0 +1,120 @@
+# Copyright 2020 The SQLFlow Authors. All rights reserved.
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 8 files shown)


python/runtime/dbapi/connection.py
```diff
@@ -0,0 +1,120 @@
...
+        Args:
+            statement: the statement to execute
+
+        Returns:
+            A ResultSet object which is iteratable, each generated 
+            record in the iterator is a result-row wrapped by list
+        """
+        return self._get_result_set(statement)
+
+    def exec(self, statement):
```
typhoonzero, line 112: Do we also need the `buffered_db_writer` functions?

lhw362950217: I'm afraid not now, it seems the buffered_db_writer keep some internal state and do not follow the DB-API pattern.

typhoonzero (approved):

python/runtime/dbapi/connection.py
lhw362950217: We'd better first implement the DB-API, then let the buffered_db_writer to call the DB-API if possible. We can do something to simplify the buffered_db_writer, for example, let Connection take the information of the URI, so we do not need to supply so much params when calling buffered_db_writer.

PR #2791, opened 2020-08-05
Generate workflow step code using runtime fea der

typhoonzero:
TODO:

- Add end2end test of running step code
- add Couler code gen
- add workflow test


go/codegen/experimental/xgboost.go
```diff
@@ -121,30 +118,46 @@ func XGBoostGenerateTrain(trainStmt *ir.TrainStmt, session *pb.Session) (string,
 var xgbTrainTemplate = `
 def step_entry():
     import json
-    import runtime
+	import tempfile
```
sneaxiy: Correct the indent?

go/codegen/experimental/xgboost.go
```diff
@@ -153,14 +166,55 @@ def step_entry():
...
+		if fd.IsSparse {
+			isSparseStr = "True"
+		}
+		vocabList := []string{}
+		for k := range fd.Vocabulary {
+			vocabList = append(vocabList, k)
+		}
+		dataFormat := "PLAIN"
+		if fd.Format != "" {
+			dataFormat = fd.Format
```
sneaxiy: In the Go side, `Format` is in lower case, but in the Python side, `dataFormat` is in upper case. 

BTW, users cannot specify the data format explicitly. We set data format in the derivation stage. So do we need to parse data format here?

go/codegen/experimental/xgboost.go
```diff
@@ -153,14 +166,55 @@ def step_entry():
...
+            dval = None
         eval_result = runtime.{{.Submitter}}.xgboost.train(dtrain, train_params, model_params, dval)
 `
 
+func generateFeatureColumnCode(fcList []ir.FeatureColumn) (string, error) {
+	fcCodes := make([]string, 0, len(fcList))
+	for _, fc := range fcList {
+		// xgboost have no cross feature column, just get the first field desc
+		fd := fc.GetFieldDesc()[0]
+		tmpl := `fc.%s(fd.FieldDesc(name="%s", dtype=fd.DataType.%s, delimiter="%s", format=fd.DataFormat.%s, shape=%s, is_sparse=%s, vocabulary=%s))`
```
sneaxiy: Can do it in the next PR. XGBoost may support `IndicatorColumn` and `BucketColumn`, whose constructors in Python do not accept `FieldDesc`. Maybe we need a better way to convert the parsed IR feature column map to the Python side.

python/runtime/feature/field_desc.py
```diff
@@ -26,6 +26,7 @@ class DataType(object):
     INT = 0
     FLOAT = 1
     STRING = 2
+    INT64 = 3
```
sneaxiy: We can keep only one of `INT` or `INT64`?

python/runtime/feature/field_desc.py
```diff
@@ -87,6 +90,9 @@ def to_json(self):
         """
         return json.dumps({
             "name": self.name,
+            # this line is used to be compatible to current code, remove it
```
sneaxiy: Add `FIXME`?

go/codegen/experimental/xgboost.go
typhoonzero: Sure.

python/runtime/feature/field_desc.py
typhoonzero: Need to map dtype to TensorFlow data types finally, int64 is needed here.

sneaxiy (approved):
LGTM.

PR #2793, opened 2020-08-05
 Add MySQL db-api implementation

lhw362950217:
Releated to #2781


python/runtime/dbapi/connection.py
```diff
@@ -66,19 +70,38 @@ def success(self):
...
+        self.uristr = conn_uri
+        self.uripts = self._parse_uri()
+        self.params = parse_qs(
+            self.uripts.query,
+            keep_blank_values=True,
+        )
+        self.params["database"] = self.uripts.path.strip("/")
+        for k, l in self.params.items():
+            if len(l) == 1:
+                self.params[k] = self.params[k][0]
```
sneaxiy, line 96: Not quite sure. Is it safe in Python if we change `self.params` when iterating?

sneaxiy (approved):
LGTM with tiny comment.

python/runtime/dbapi/connection.py
lhw362950217: I think it's ok to modify the value, but we should not del or add keys. I tried some case, it just works.

PR #2794, opened 2020-08-05
Add json dump and load support for FeatureColumn

sneaxiy:
Add `json.dumps` and `json.loads` support for `FeatureColumn`, so that the IR feature column map can be saved as JSON string in the future.


typhoonzero (approved):
Can merge and do more discuss later.

PR #2799, opened 2020-08-06
Push images on self hosted machine

typhoonzero:
Use shared workers to build and push docker images to aliyun registry often takes to much time, so use our self-hosted worker to publish docker images when merging PRs to develop.


lhw362950217 (approved):
LGTM

PR #2807, opened 2020-08-10
Refine metadata collect and save/load

sneaxiy:
Make `collect_metadata` save `features` and `label` FeatureColumn possible. And move `model_metadata.py` to `runtime/model` directory.


python/runtime/model/metadata_test.py
```diff
@@ -0,0 +1,102 @@
...
+
+import os
+import unittest
+
+from runtime.feature.column import NumericColumn
+from runtime.feature.field_desc import FieldDesc
+from runtime.model.metadata import (collect_metadata, load_metadata,
+                                    save_metadata)
+
+FILENAME = 'meta.json'
```
lhw362950217: Shall we use FILE_NAME instead of FILENAME?

sneaxiy: Done. Move to `self.file_name`.

lhw362950217 (approved):
LGTM

PR #2809, opened 2020-08-10
Adapt paiio with DB-API

lhw362950217:
Related to #2781


python/runtime/dbapi/__init__.py
```diff
@@ -14,11 +14,13 @@
 from runtime.dbapi.hive import HiveConnection
 from runtime.dbapi.maxcompute import MaxComputeConnection
 from runtime.dbapi.mysql import MySQLConnection
+from runtime.dbapi.paiio import PaiIOConnection
 
 DRIVRE_MAP = {
```
sneaxiy: `DRIVER_MAP`?

sneaxiy (approved):

python/runtime/dbapi/paiio.py
```diff
@@ -0,0 +1,146 @@
...
+        self.params["table"] = match[0]
+        self.params["slice_id"] = self.params.get("slice_id", 0)
+        self.params["slice_count"] = self.params.get("slice_count", 1)
+
+    def _get_result_set(self, statement):
+        if statement is not None:
+            raise ValueError("paiio only support full table read,"
+                             "so you need to pass statement with None.")
+        try:
+            reader = paiio.TableReader(self.params["table"],
```
typhoonzero, line 99: It is said that `paiio.TableRecordDataset` is faster than `TableReader`, and `TableReader` is actually not used currently.

lhw362950217: I will refer to the doc and change the function if possible.

PR #2815, opened 2020-08-12
Polish DB-API to support Python2 so can run on PAI

lhw362950217:
Related to #2781 

1. Enable absolute import in Python2 
1. Add param to super() function
1. Change exec to execute because exec in Python2 is a keyword
1. Add some driver specific API


python/runtime/dbapi/connection.py
```diff
@@ -102,6 +102,11 @@ def __init__(self, conn_uri):
             if len(l) == 1:
                 self.params[k] = l[0]
 
+    def param(self, param_name, default_value=""):
```
sneaxiy, line 105: Seems that this method is not used?

python/runtime/dbapi/maxcompute.py
```diff
@@ -81,18 +83,33 @@ class MaxComputeConnection(Connection):
...
+        self.params["database"] = proj
+        self.endpoint = endpoint
+        self._conn = ODPS(user, pwd, project=proj, endpoint=endpoint)
+
+    @staticmethod
+    def get_uri_parts(uri):
+        """Get username, password, endpoint, projectfrom given uri
+
+        Args:
+            uri: a vliad maxcompute connection uri
```
sneaxiy: Typo `vliad`.

python/runtime/dbapi/connection.py
lhw362950217: This is a tool api, we will need it in our submitter code.

sneaxiy (dismissed):

python/runtime/dbapi/hive_test.py
```diff
@@ -51,15 +51,15 @@ def test_query(self):
 
     def test_exec(self):
         conn = HiveConnection(testing.get_datasource())
-        rs = conn.exec("create table test_exec(a int)")
+        rs = conn.execute("create table test_exec(a int)")
```
typhoonzero, line 54: Is the test running on the CI currently?

lhw362950217: Not really, I will add it.

typhoonzero (approved):

PR #2817, opened 2020-08-12
Add  Model.save_to_oss and Model.load_from_oss

sneaxiy:


python/runtime/model/model.py
```diff
@@ -126,8 +204,8 @@ def _dump_pkl(obj, to_file):
         pickle.dump(obj, f, pickle.HIGHEST_PROTOCOL)
```
lhw362950217: I'm curious about if `pickle.HIGHEST_PROTOCOL` can work on lower PROTOCOL, like we upload a model created by Py3, and load it by Py2.

python/runtime/model/model.py
```diff
@@ -67,56 +69,132 @@ def __init__(self, typ, meta):
...
+            self._zip(local_dir, tarball)
+            oss.save_file(oss_model_dir, tarball, TARBALL_NAME)
+
+    @staticmethod
+    def load_from_oss(oss_model_dir, local_dir=os.getcwd()):
+        """
+        Load the saved model from OSS and unzip it on local_dir.
+
+        Args:
+            oss_model_dir (str): the OSS model directory to load.
```
typhoonzero, line 188: Maybe document the format of `oss_model_dir` should be like `oss://bucket/path/to/dir`?

python/runtime/model/model.py
sneaxiy: Great suggestions! Changed to `protocol=2` to be compatible with Py2 and Py3.

python/runtime/model/model.py
sneaxiy: Done.

lhw362950217 (approved):
LGTM

PR #2821, opened 2020-08-12
Use unified DB-API in codebase

### LINKED ISSUE #2781: Unify db api

lhw362950217:
Fixes #2781


python/runtime/xgboost/predict.py
```diff
@@ -124,19 +121,8 @@ def predict_and_store_result(bst, dpred, feature_file_id, model_params,
...
-                               conn,
-                               result_table,
-                               result_column_names,
-                               100,
-                               hdfs_namenode_addr=hdfs_namenode_addr,
-                               hive_location=hive_location,
-                               hdfs_user=hdfs_user,
-                               hdfs_pass=hdfs_pass) as w:
+    with db.buffered_db_writer(conn, result_table, result_column_names,
+                               100) as w:
```
typhoonzero, line 125: Why remove the other parameters?

python/runtime/pai/submitter.py
```diff
@@ -130,15 +132,18 @@ def get_oss_model_url(model_full_path):
     return "oss://%s/%s" % (oss.SQLFLOW_MODELS_BUCKET, model_full_path)
 
 
+def parse_maxcompute_dsn(datasource):
```
sneaxiy, line 135: Do we need this method or just write `MaxComputeConnection.get_uri_parts(datasource)` when calling?

sneaxiy (approved):
LGTM basically.

python/runtime/xgboost/predict.py
lhw362950217: Parameters like `hdfs_pass` can be passed by connection uri, they are driver specific. We have a unified way to save these params (in `conn.params dict`), we can get them  by `conn.param(param_name)` when needed.

python/runtime/pai/submitter.py
lhw362950217: Good, idea. I will figure this out and do some modify if possible. In next PR :)

python/runtime/xgboost/predict.py
typhoonzero: I see, the `db.py` file was folded and I missed the file.

PR #2822, opened 2020-08-13
Save the trained xgboost model

sneaxiy:
Refactorization: save the XGBoost model and metadata after training.


typhoonzero (approved):
LGTM++

PR #2824, opened 2020-08-13
Generate workflow step for normal statement run

typhoonzero:


python/runtime/dbapi/connection.py
```diff
@@ -144,6 +144,20 @@ def query(self, statement):
...
         return self._get_result_set(statement)
 
+    def is_query(self, statement):
+        """Return true if the statement is a query SQL statement."""
+        s = statement.strip()
+        s = s.upper()
+
+        if s.startswith("SELECT") and s.find("INTO") == -1:
+            return True
+        if s.startswith("SHOW") and s.find("CREATE") >= 0 and s.find(
```
sneaxiy: Seems not the same with the Go code:

https://github.com/sql-machine-learning/sqlflow/blob/a96cb7967a4b9cc82620b0286804b17e6d879f2c/go/executor/sql_stmt.go#L52

python/runtime/dbapi/connection.py
```diff
@@ -144,6 +144,20 @@ def query(self, statement):
...
+                "DATABASES") >= 0 and s.find("TABLES"):
```
sneaxiy: `s.find("TABLES") >= 0`

sneaxiy (approved):

PR #2828, opened 2020-08-14
Fix pai training with optimizer config

typhoonzero:
When training on PAI, we generate the train code calls `runtime.pai.tensorflow.train` which will call `oss.save_oss_model` to save the model and model meta-data to OSS, yet we generate the code using `tfSaveModelTmplText` to save the model meta-data again:

https://github.com/sql-machine-learning/sqlflow/blob/6974de27a052fc7b4b18f02be93fa6f04ffc4e2f/go/codegen/pai/template_tf.go#L95-L127

We do not need to save the meta-data twice.

And, we need to save the model meta-data as some Python code to save feature column, optimizer info. Yet `runtime.pai.tensorflow.train` tries to save the `model_params` as objects which will cause an error, we need to save the model_params before it is converted to objects.

```
File "/tensorflow_jobs/runtime/pai/tensorflow/train.py", line 123, in train
feature_columns_code, num_workers)
File "/tensorflow_jobs/runtime/model/oss.py", line 292, in save_oss_model
label_meta, model_params, feature_columns_code)
File "/tensorflow_jobs/runtime/model/oss.py", line 232, in save_metas
serialized = pickle.dumps(list(meta))
...
raise TypeError, "can't pickle %s objects" % base.__name__
TypeError: can't pickle SwigPyObject objects
```

We should polish more after this fix was merged.

### DIFF
--- a/go/codegen/pai/tensorflow.go
+++ b/go/codegen/pai/tensorflow.go
@@ -70,20 +70,7 @@ func TFTrainWithLoadAndSave(ir *ir.TrainStmt, session *pb.Session, modelPathToSa
 		return "", err
 	}
 
-	// append code snippet to save model
-	checkpointDir := OSSModelURL(modelPathToSave)
-	var tpl = template.Must(template.New("SaveModel").Parse(tfSaveModelTmplText))
-	filler := saveModelFiller{
-		OSSModelDir: checkpointDir,
-		Estimator:   ir.Estimator,
-		NumWorkers:  cc.Worker.Count,
-	}
-	var saveCode bytes.Buffer
-	if err = tpl.Execute(&saveCode, filler); err != nil {
-		return "", err
-	}
-
-	fullCode := fmt.Sprintf("%s\n%s\n%s", loadCode, trainCode, saveCode.String())
+	fullCode := fmt.Sprintf("%s\n%s", loadCode, trainCode)
 	return fullCode, nil
 }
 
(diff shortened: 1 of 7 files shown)


sneaxiy (commented):
Not quite understand what this PR does. Would you like to explain why this PR?

go/codegen/pai/tensorflow.go
```diff
@@ -70,20 +70,7 @@ func TFTrainWithLoadAndSave(ir *ir.TrainStmt, session *pb.Session, modelPathToSa
...
-		OSSModelDir: checkpointDir,
-		Estimator:   ir.Estimator,
-		NumWorkers:  cc.Worker.Count,
-	}
-	var saveCode bytes.Buffer
-	if err = tpl.Execute(&saveCode, filler); err != nil {
-		return "", err
-	}
-
-	fullCode := fmt.Sprintf("%s\n%s\n%s", loadCode, trainCode, saveCode.String())
```
sneaxiy, line 86: Why remove the save code?

python/runtime/pai/tensorflow/train.py
```diff
@@ -52,6 +52,7 @@ def train(datasource,
           pai_table="",
           pai_val_table="",
           feature_columns_code="",
+          model_params_code_map={},
```
sneaxiy, line 55: Why add this parameter, and what does this parameter mean?

sneaxiy (approved):
LGTM.

Issue #2831, opened 2020-08-14
Move hdfs config from specific params to uri


lhw362950217:
**Is your feature request related to a problem? Please describe.**
We pass hdfs params from go to python, and the params are then passed down for times, even if we do not really need them in non-hive environment. Maybe we can remove these params, and assemble them in the hive uri (which is a conventional practice) and make the interface simpler.

**Describe the solution you'd like**
Remove hdfs related params, instead, we encode the params in connection uri like:
hive://user:pwd@localhost:1000/iris?hdfs_namenode_addr=xxx&hive_location=xxx

**Describe alternatives (optional)**

**Additional Notes**

PR #2833, opened 2020-08-14
Remove hdfs params

lhw362950217:
Related to #2831 

1. Remove most hdfs params
1. Did not remove protobuf fields for backward compatibility
1. As hdfs params are all optional in Python, we can remove the Go part first, then the Python part


wangkuiyi (approved):
LGTM!

PR #2838, opened 2020-08-17
Add temp_file.py

sneaxiy:
`tempfile.TemporaryDirectory` does not exist in Python 2. Add `temp_file.py` to unify the API.


typhoonzero (approved):
Cool

PR #2844, opened 2020-08-18
Add xgboost evaluation local code

sneaxiy:


python/runtime/local/xgboost_submitter/predict.py
```diff
@@ -122,8 +114,27 @@ def _predict_and_store_result(bst, dpred, model_params, result_table,
...
 
-    with db.buffered_db_writer(conn, result_table, result_column_names,
-                               100) as w:
+    return preds
+
+
+def _store_predict_result(preds, result_table, result_column_names,
+                          train_label_idx, feature_file_name, conn):
+    """
+    Save the prediction result in the table.
```
typhoonzero, line 129: Maybe documents for private functions can be omitted.

sneaxiy: These documents are written for our developers to better code review. It may be necessary because the Python code is not easy to read than the Go code.

typhoonzero (approved):

Issue #2850, opened 2020-08-19
Fix quick start doc


lhw362950217:
**Is your feature request related to a problem? Please describe.**
Our quick start doc has been expired for a while, we should make a quick fix and let user get a first experience 

**Describe the solution you'd like**
1. As a quick fix, first point the quick start entrance to our cloud playground
1. Hide the old docker approach pages
2. Add doc to replace the old docker way, (with hive, maxcompte, etc.)

**Describe alternatives (optional)**
A clear description of any alternative solutions or features

**Additional Notes**

PR #2851, opened 2020-08-19
Point quick start to our cloud playground

lhw362950217:
This is the first step in  #2850


typhoonzero:
Should merge after we polished the playground.

PR #2859, opened 2020-08-21
Polish pai submitter structure

typhoonzero:


lhw362950217 (approved):
LGTM

Let's do some e2e test on PAI after merge.

PR #2864, opened 2020-08-21
Fix PAI XGBoost model evaluation driver bug

sneaxiy:


lhw362950217:
I think it's better to change https://github.com/sql-machine-learning/sqlflow/blob/fed16dfcd5e1e6e019ab9e66de7be77d3b3c5066/python/runtime/xgboost/evaluate.py#L61
to：
```python
from runtime.dbapi import PaiioConnection
conn = PaiioConnection.from_table(result_table)
```
and keep other place as they are

sneaxiy:
@lhw362950217 Great idea! I have changed the implementation as you suggested. Please take a look.

typhoonzero (approved):

PR #2865, opened 2020-08-24
Runtime.pai use python feature column implementation

typhoonzero:
Can merge this then I'll update prediction/explain/evaluate.

### DIFF
--- a/python/runtime/pai/tensorflow_submitter/train.py
+++ b/python/runtime/pai/tensorflow_submitter/train.py
@@ -18,7 +18,7 @@ from runtime import db
 from runtime.feature.compile import compile_ir_feature_columns
 from runtime.feature.derivation import (get_ordered_field_descs,
                                         infer_feature_columns)
-from runtime.model import collect_metadata, oss
+from runtime.model import EstimatorType, Model, collect_metadata, oss
 from runtime.pai.pai_distributed import define_tf_flags, set_oss_environs
 from runtime.pai.tensorflow_submitter.train_estimator import \
     estimator_train_and_save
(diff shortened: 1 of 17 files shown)


python/runtime/pai/tensorflow_submitter/train.py
```diff
@@ -125,3 +130,130 @@ def train(datasource,
...
+                                                   feature_column_map,
+                                                   label_column,
+                                                   n=1000)
+    fc_map = compile_ir_feature_columns(fc_map_ir, EstimatorType.TENSORFLOW)
+    field_descs = get_ordered_field_descs(fc_map_ir)
+    feature_column_names = [fd.name for fd in field_descs]
+    feature_metas = dict([(fd.name, fd.to_dict()) for fd in field_descs])
+    label_meta = label_column.get_field_desc()[0].to_dict()
+
+    feature_column_names_map = dict()
```
sneaxiy, line 161: Seems that `feature_column_names_map` is not used?

python/runtime/temp_file.py
```diff
@@ -19,7 +19,7 @@
 # NOTE: Python 2 does not have tempfile.TemporaryDirectory. To unify the code
 # of Python 2 and 3, we make the following class.
 class TemporaryDirectory(object):
-    def __init__(self, as_cwd=False, suffix=None, prefix=None, dir=None):
+    def __init__(self, as_cwd=False, suffix="", prefix="", dir=""):
```
sneaxiy, line 22: Why change this?

python/runtime/pai/tensorflow_submitter/train.py
```diff
@@ -125,3 +130,130 @@ def train(datasource,
...
+                                        batch_size)
+
+    model_meta = collect_metadata(original_sql=original_sql,
+                                  select=select,
+                                  validation_select=validation_select,
+                                  model_repo_image=model_image,
+                                  class_name=estimator_string,
+                                  attributes=model_params,
+                                  features=fc_map_ir,
+                                  label=label_meta)
```
sneaxiy: How about `label= fc_label_ir` instead of `label=label_meta`?

python/runtime/temp_file.py
typhoonzero: When running on PAI using Python 2.7.x, below `suffix + some_str` may cause error since `suffix` is None type by default.

python/runtime/pai/tensorflow_submitter/train.py
typhoonzero: Added back saving model meta to OSS, it will use `feature_column_names_map`.

sneaxiy (approved):

PR #2868, opened 2020-08-27
Retain build-essential and python3-dev in the step image.

brightcoder01:
It's common that we will build the runnable images based on the base image - sqlflow/sqlflow:step and then install additional python packages. Some python packages such as `psutil` need compilation which requires gcc and `Python.h`. In order the prepare the compilation environment in the base image, we will retain build-essential and python3-dev packages in sqlflow/sqlflow:step.

### DIFF
--- a/docker/step/Dockerfile
+++ b/docker/step/Dockerfile
@@ -13,7 +13,7 @@ RUN /bin/bash -c 'if [ "$FIND_FASTED_MIRROR" == "true" ]; then source find_faste
   && echo "Choose the fastest PIP source ..." \
   && choose_fastest_pip_source; fi' && \
   apt-get update && \
-  apt-get -qq install -y --no-install-recommends openjdk-8-jre-headless python3 libmysqlclient20 python3-idna libgomp1 python3-setuptools python3-pip && \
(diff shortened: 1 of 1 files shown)


docker/step/Dockerfile
```diff
@@ -13,7 +13,7 @@ RUN /bin/bash -c 'if [ "$FIND_FASTED_MIRROR" == "true" ]; then source find_faste
   && echo "Choose the fastest PIP source ..." \
   && choose_fastest_pip_source; fi' && \
   apt-get update && \
-  apt-get -qq install -y --no-install-recommends openjdk-8-jre-headless python3 libmysqlclient20 python3-idna libgomp1 python3-setuptools python3-pip && \
+  apt-get -qq install -y --no-install-recommends openjdk-8-jre-headless python3 libmysqlclient20 python3-idna libgomp1 python3-setuptools python3-pip build-essential python3-dev && \
```
weiguoz, line 16: Do we need both `python3` and `python3-dev`?

brightcoder01: `python3-dev` has dependency on `python3`. Installing python3-dev will trigger the python3 installation automatically. Keeping both is also fine because it won't bring additional overhead.

weiguoz (approved):
LGTM

PR #2869, opened 2020-08-27
Fix xgb string data contains slash

typhoonzero:
When the string feature contains "/", then after training a xgboost model using `COLUMN CATEGORY_HASH(col)` and predict some data using this model, the original data is processed with error.


weiguoz (dismissed):
LGTM

brightcoder01 (approved):

PR #2876, opened 2020-09-01
Fix xgb predict error

typhoonzero:


go/cmd/sqlflowserver/testing.go
```diff
@@ -101,26 +102,50 @@ func sqlRequest(sql string) *pb.Request {
...
-// AssertEqualAny checks any type of returned protobuf message to an interface
-func AssertEqualAny(a *assert.Assertions, expected interface{}, actual *any.Any) {
+// EqualAny checks any type of returned protobuf message to an interface
+func EqualAny(expected interface{}, actual *any.Any) bool {
 	switch actual.TypeUrl {
 	case "type.googleapis.com/google.protobuf.StringValue":
 		b := wrappers.StringValue{}
 		ptypes.UnmarshalAny(actual, &b)
-		a.Equal(expected, b.Value)
+		if expected == b.Value {
```
sneaxiy: NIT. Just write `return expected == b.Value`? Same below.

go/cmd/sqlflowserver/e2e_mysql_test.go
```diff
@@ -54,6 +54,71 @@ INTO sqlflow_models.custom_loop_model_eval_result;`, caseTrainTable)
 	}
 }
 
+func caseTrainXGBoostMultiClassi(t *testing.T) {
```
sneaxiy: `caseTrainXGBoostMultiClassi` -> `caseTrainXGBoostMultiClass`?

python/runtime/xgboost/predict.py
```diff
@@ -78,23 +78,11 @@ def predict_and_store_result(bst, dpred, feature_file_id, model_params,
...
-        # prediction output with multi-class job has two dimensions, this
-        # is a temporary way, can remove this else branch when we can load
-        # the model meta not only on PAI submitter.
-        if len(preds.shape) == 2:
-            preds = np.argmax(np.array(preds), axis=1)
+    # prediction output with multi-class job has two dimensions, this
+    # is a temporary way, can remove this else branch when we can load
+    # the model meta not only on PAI submitter.
+    if len(preds.shape) == 2:
+        preds = np.argmax(np.array(preds), axis=1)
```
sneaxiy, line 85: Why change this? I think that `model_params` has been got [here](https://github.com/sql-machine-learning/sqlflow/pull/2876/files#diff-009614f1b6ee464541b3aec9b86ad5d0L60). And we can get `objective` here so that we can distinguish binary classification, multi-class classification, and regression models.

go/ir/derivation_test.go
```diff
@@ -223,6 +224,15 @@ func TestFeatureDerivation(t *testing.T) {
 
 	a := assert.New(t)
 	db := database.GetTestingDBSingleton()
+	var testDataSQL string
+	if testDB == "mysql" {
+		testDataSQL = testdata.FeatureDerivationCaseSQL
+	} else if testDB == "hive" {
+		testDataSQL = testdata.FeatureDerivationCaseSQLHive
+	}
+	if err := testdata.Popularize(db.DB, testDataSQL); err != nil {
```
sneaxiy, line 233: How about add the dataset inside https://github.com/sql-machine-learning/sqlflow/blob/ab1592cf3f998630540fde4ec7d23befb90b6812/go/database/testing.go#L86 and https://github.com/sql-machine-learning/sqlflow/blob/ab1592cf3f998630540fde4ec7d23befb90b6812/go/database/testing.go#L111?

python/runtime/xgboost/predict.py
typhoonzero: `preds` here is already the classification label here, no need to do `argmax` anymore.

go/ir/derivation_test.go
typhoonzero: There are many places that create test datasets, will try refine this in later PRs.

python/runtime/xgboost/predict.py
sneaxiy: I see.

sneaxiy (approved):

PR #2891, opened 2020-09-07
Add refactored workflow codegen e2e pai and local

typhoonzero:


python/runtime/pai/get_pai_tf_cmd.py
```diff
@@ -66,8 +66,9 @@ def get_pai_tf_cmd(cluster_config, tarball, params_file, entry_file,
            ) % (job_name, tarball, entry_file, submit_tables, output_tables,
                 params_file)
 
-    # format the oss checkpoint path with ARN authorization.
-    oss_checkpoint_configs = os.getenv("SQLFLOW_OSS_CHECKPOINT_CONFIG")
+    # format the oss checkpoint path with ARN authorization, should use eval
+    # because we use '''json''' in the workflow yaml file.
+    oss_checkpoint_configs = eval(os.getenv("SQLFLOW_OSS_CHECKPOINT_CONFIG"))
```
sneaxiy, line 71: How about use `json.loads`?

python/runtime/pai/submitter_test.py
```diff
@@ -33,8 +33,8 @@ def test_get_oss_model_url(self):
 
     def test_get_pai_tf_cmd(self):
         conf = get_cluster_config({})
-        os.environ[
-            "SQLFLOW_OSS_CHECKPOINT_CONFIG"] = '{"arn":"arn", "host":"host"}'
+        os.environ["SQLFLOW_OSS_CHECKPOINT_CONFIG"] = \
```
sneaxiy, line 36: Why use this format instead of standard JSON format `'''{"arn":"arn", "host":"host"}'''`?

python/runtime/pai/xgboost_submitter/train.py
```diff
@@ -155,11 +155,12 @@ def train_step(original_sql,
         feature_column_names, *feature_columns["feature_columns"])
 
     batch_size = train_params.get("batch_size", None)
+    train_params.pop("batch_size")
```
sneaxiy: Just write `batch_size = train_params.pop("batch_size", None)`. Same below.

python/runtime/pai/get_pai_tf_cmd.py
typhoonzero: The string is actually `'''{json object}'''` we need to eval the string then call `json.loads`.

python/runtime/pai/submitter_test.py
typhoonzero: Couler generates the env like `'''{json object}'''`.

sneaxiy (approved):
LGTM.

PR #2894, opened 2020-09-07
Support null values for xgboost

### LINKED ISSUE #2888: Support None value in wide table for XGBoost train and explain.
In the Jupyter Notebook of the [Boston house tutorial](https://github.com/sql-machine-learning/sqlflow/blob/develop/doc/tutorial/housing-explain.md), input the following SQL statements:

```SQL
CREATE TABLE boston.train_ext AS SELECT * FROM boston.train;

UPDATE boston.train_ext
SET rad = NULL
WHERE zn < 18.1 AND zn > 17.0;

SELECT * FROM boston.train_ext
TO TRAIN xgboost.gbtree
WITH
    objective="reg:squarederror",
    train.num_boost_round = 30
COLUMN crim, zn, indus, chas, nox, rm, age, dis, rad, tax, ptratio, b, lstat
LABEL medv
INTO sqlflow_models.my_xgb_regression_model;
```

We will get the following error messages:

```TXT

  File "<stdin>", line 108, in <module>

  File "/opt/sqlflow/python/runtime/xgboost/train.py", line 170, in train

    for per_batch_dmatrix in dtrain:

  File "/opt/sqlflow/python/runtime/xgboost/dataset.py", line 89, in xgb_dataset

    raw_data_dir=raw_data_dir)

  File "/opt/sqlflow/python/runtime/xgboost/dataset.py", line 130, in dump_dmatrix

    feature_metas)

  File "/opt/sqlflow/python/runtime/db.py", line 350, in read_features_from_row

    feature_metas[name], name)

  File "/opt/sqlflow/python/runtime/db.py", line 188, in read_feature

    int_raw_val = INT64_TYPE(raw_val)

TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType'
```

Since XGBoost model can handle the missing values, we can pass the `none` value into XGBoost directly instead of reporting error while dumping the dmatrix file.

typhoonzero:
Fix https://github.com/sql-machine-learning/sqlflow/issues/2888

### DIFF
--- a/python/runtime/db.py
+++ b/python/runtime/db.py
@@ -30,7 +30,7 @@ INT64_TYPE = long if six.PY2 else int  # noqa: F821
 XGBOOST_NULL_MAGIC = 9999.0
 
 
-def read_feature(raw_val, feature_spec, feature_name):
+def read_feature(raw_val, feature_spec, feature_name, is_xgboost):
     # FIXME(typhoonzero): Should use correct dtype here.
     if feature_spec["is_sparse"]:
         if feature_spec["format"] == "kv":
@@ -61,14 +61,18 @@ def read_feature(raw_val, feature_spec, feature_name):
         # Dense string vector
         if feature_spec["dtype"] == "float32":
             if raw_val is None:
-                vec = np.array([XGBOOST_NULL_MAGIC], dtype=np.float32)
+                raise ValueError(
+                    "column %s value is NULL, expected dense vector with delimiter %s"
+                    % (feature_name, feature_spec["delimiter"]))
             else:
                 vec = np.fromstring(raw_val,
                                     dtype=np.float32,
                                     sep=feature_spec["delimiter"])
         elif feature_spec["dtype"] == "int64":
             if raw_val is None:
-                vec = np.array([int(XGBOOST_NULL_MAGIC)], dtype=np.int64)
+                raise ValueError(
+                    "column %s value is NULL, expected dense vector with delimiter %s"
+                    % (feature_name, feature_spec["delimiter"]))
             else:
                 vec = np.fromstring(raw_val,
                                     dtype=np.int64,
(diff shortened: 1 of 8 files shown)


python/runtime/db.py
```diff
@@ -25,9 +25,13 @@ def connect_with_data_source(driver_dsn):
 
 
 INT64_TYPE = long if six.PY2 else int  # noqa: F821
+XGBOOST_NULL_MAGIC = 9999.0
 
 
 def read_feature(raw_val, feature_spec, feature_name):
+    # FIXME(typhoonzero): return None if not XGBoost
+    if raw_val is None:
+        return XGBOOST_NULL_MAGIC,
```
brightcoder01: What about the feature value of int type instead of double?
And how to handle np.array?

python/runtime/db.py
```diff
@@ -25,50 +25,76 @@ def connect_with_data_source(driver_dsn):
...
+            else:
+                vec = np.fromstring(raw_val,
+                                    dtype=np.float32,
+                                    sep=feature_spec["delimiter"])
         elif feature_spec["dtype"] == "int64":
-            vec = np.fromstring(raw_val,
-                                dtype=np.int64,
-                                sep=feature_spec["delimiter"])
+            if raw_val is None:
+                vec = np.array([int(XGBOOST_NULL_MAGIC)], dtype=np.int64)
```
brightcoder01: The shape is not matched with the not-none case.
Will there be any offset inconsistency between none/not-none case inside [dump_matrix](https://github.com/sql-machine-learning/sqlflow/blob/c67113ecd3cc3115aac8bf56f0f7439054969de6/python/runtime/xgboost/dataset.py#L149-L155).

python/runtime/db.py
```diff
@@ -25,50 +25,85 @@ def connect_with_data_source(driver_dsn):
...
+                                    dtype=np.float32,
+                                    sep=feature_spec["delimiter"])
         elif feature_spec["dtype"] == "int64":
-            vec = np.fromstring(raw_val,
-                                dtype=np.int64,
-                                sep=feature_spec["delimiter"])
+            if raw_val is None:
+                raise null_feature_error
+            else:
+                vec = np.fromstring(raw_val,
```
brightcoder01, line 76: How can we ensure that each item of a specific dense feature has the same shape?
np.fromstring may returns various shape for different input strings.

typhoonzero: This is done by feature derivation before training.

brightcoder01 (approved):

PR #2906, opened 2020-09-09
Add step init code for step container

lhw362950217:
Related to #2905 

In some circumstances, we need to do some container initialization (which can't be done on image build, say, get current route info) before start the step code. We add a place holder file `/opt/sqlflow/init_step_container.sh`, if the image builder generate the file, we will call it before execute the step logic.


Yancey0623 (approved):
LGTM

Issue #2913, opened 2020-09-10
Export TensorFlow SavedModel for online serving


sneaxiy:
TensorFlow supports using `SavedModel` for online serving.

Currently, we have several ways to save the trained TensorFlow model.

- For Keras model:
  - For distributed training, we use `Estimator.export_saved_model` to save the trained model. This saved model includes model definition and weights. This case can be used for online serving.
https://github.com/sql-machine-learning/sqlflow/blob/30bba2983260c4bd615d966efb31657a65958f15/python/runtime/pai/tensorflow_submitter/train_keras.py#L105
  - For single-machine training, we use `keras.Model.save_weights` to save an HDF5 file, which only includes model weights. It cannot be used for online serving directly, because (1) it lacks model definition; (2) HDF5 file must be converted to `SavedModel` so that it can be loaded by TensorFlow serving components.
https://github.com/sql-machine-learning/sqlflow/blob/30bba2983260c4bd615d966efb31657a65958f15/python/runtime/tensorflow/train_keras.py#L185

- For Estimator model, we use `Estimator.export_saved_model` to save the trained model too. This save model includes model definition and weights. This case can be used for online serving.

https://github.com/sql-machine-learning/sqlflow/blob/30bba2983260c4bd615d966efb31657a65958f15/python/runtime/tensorflow/train_estimator.py#L61

We should unify the model saving codes and make the saved model can be used for online serving directly.

sneaxiy:
Close by https://github.com/sql-machine-learning/sqlflow/pull/2914 and https://github.com/sql-machine-learning/sqlflow/pull/2917 .

PR #2914, opened 2020-09-10
Try to save model definition and weights together after training Keras model

sneaxiy:
Address https://github.com/sql-machine-learning/sqlflow/issues/2913 .


typhoonzero (approved):
LGTM++

PR #2917, opened 2020-09-11
Add download model doc

typhoonzero:


doc/download_model.md
```diff
@@ -0,0 +1,33 @@
...
+[command-line tool](run/cli.md) first. Then run the below command to download
+your model:
+
+```shell
+# Configure database connection string
+export SQLFLOW_DATASOURCE="mysql://root:root@tcp(127.0.0.1:3306)/?"
+# Configure sqlflow server address
+export SQLFLOW_SERVER="localhost:50051"
+# Assume you have trained a model named sqlflow_models.my_dnn_model
+./sqlflow get model "sqlflow_models.my_dnn_model"
```
wangkuiyi: It would be easier for readers to understand the correlation if we include an SQLFlow TO TRAIN statement with the model name before this command-line.

doc/download_model.md
```diff
@@ -0,0 +1,33 @@
+# Download Your Trained Model
+
+Once you have trained your model with SQLFlow you may need to download your
```
wangkuiyi: Once we train a model using SQLFlow via a statement like the following,

```sql
SELECT ... TO TRAIN # Please put a realistic statement here.
```

the trained model goes to the model zoo.  Anyone with the read access can write SQLFlow statement to visually explain the model or to use the model for prediction.

In some cases, you might want to export and download a trained model from the model zoo, so you can use it out of SQLFlow, for example, load it to an online prediction service of an online advertising system.  To export a model, you can use the command-line tool `sqlflow`.
we can download the trained model, if necessary, using the

doc/download_model.md
```diff
@@ -0,0 +1,33 @@
+# Download Your Trained Model
```
wangkuiyi: Export Trained Models

wangkuiyi (commented):
Glad to see the command-line tool can now support model export.  A few suggestions about rephrasing.

doc/export_model.md
```diff
@@ -0,0 +1,46 @@
...
+Once we train a model using SQLFlow via a statement like the following,
+
+```sql
+SELECT * FROM iris.train
+TO TRAIN DNNClassifier
+WITH model.n_classes=3,model.hidden_units=[10,20]
+INTO sqlflow_models.my_dnn_model;
+```
+
+the trained model `sqlflow_models.my_dnn_model` is saved in the model zoo.
```
lhw362950217: How about link model zoo to our model zoo doc?

typhoonzero: Changed model zoo -> database. The model is uploaded to the model zoo only if the user does so.

lhw362950217 (approved):
LGTM

PR #2899, opened 2020-09-08
[WIP]Fix feature column type derivation

lhw362950217:
Fixes #2898

PR #1964, opened 2020-03-09
Fix scarlar int field used as category id column when predicting

typhoonzero:
Fix https://github.com/sql-machine-learning/sqlflow/issues/1957

TODO: need add test cases for `DNNLinearCombinedClassifier`

PR #2920, opened 2020-09-14
Fix feature column type derivation

### LINKED ISSUE #2898: IMDB tutorial report feature column type error
**Description**
Our imdb dataset has a csv format, all features are integer. However, our generated feature column did not specify data type, so it use the default float type, which cause error when prediction.

**Reproduction Steps**
```sql
    SELECT content, class FROM imdb.train
    TO TRAIN DNNClassifier
    WITH
    model.n_classes = 2,
    model.hidden_units = [128, 64]
    LABEL class
    INTO sqlflow_models.my_text_model_en;

    SELECT *
    FROM imdb.test
    TO PREDICT imdb.predict.class
    USING sqlflow_models.my_text_model_en;
```

Error message is:
```text
tensorflow.python.framework.errors_impl.InvalidArgumentError:  Name: <unknown>, Key: content, Index: 0.  Data types don't match. Data type: int64 but expected type: float
```

**Expected Behavior**

**Screenshots**

**Environment (Please complete the following information):**
 - OS:
 - Browser:
 - Version:

**Additional Notes**

lhw362950217:
Moved from #2899, Fixes #2898

Let's first fix tensorflow part.

### DIFF
--- a/python/runtime/tensorflow/predict.py
+++ b/python/runtime/tensorflow/predict.py
@@ -175,27 +175,6 @@ def estimator_predict(estimator, model_params, save, result_table,
                 example.features.feature[feature_name].int64_list.value.extend(
                     list(values))
         else:
-            if "feature_columns" in feature_columns:
-                idx = feature_column_names.index(feature_name)
-                fc = feature_columns["feature_columns"][idx]
-            else:
-                # DNNLinearCombinedXXX have dnn_feature_columns and
-                # linear_feature_columns param.
-                idx = -1
-                try:
-                    idx = feature_column_names_map[
-                        "dnn_feature_columns"].index(feature_name)
-                    fc = feature_columns["dnn_feature_columns"][idx]
-                except:  # noqa: E722
-                    try:
-                        idx = feature_column_names_map[
-                            "linear_feature_columns"].index(feature_name)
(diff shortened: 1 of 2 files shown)


python/runtime/tensorflow/predict.py
```diff
@@ -175,43 +175,15 @@ def add_to_example(example, x, i):
                 example.features.feature[feature_name].int64_list.value.extend(
                     list(values))
         else:
-            if "feature_columns" in feature_columns:
```
typhoonzero, line 178: Why we should remove this block?

lhw362950217: The removed block is to calculate `fc`. And `fc` is used to do special check when field's meta is int but should be convert to float. I think this convert is not necessary any more. The real reason we convert it is we did not pass the type info in go code, so tf will treat the data as float. And now we have fixed it. This may be relative to https://github.com/sql-machine-learning/sqlflow/pull/1964;

sneaxiy (dismissed):

python/runtime/tensorflow/predict.py
typhoonzero: Seems `feature_column_names_map` is used only at here, we can remove it later.

sneaxiy (approved):

PR #2923, opened 2020-09-15
Install bash in model zoo docker image.

brightcoder01:
Bash is not installed in modelzoo Docker image. The pod status will be CrashLoopBackOff with the error message `OCI runtime create failed: container_linux.go:349: starting container process caused "exec: \"/bin/bash\": stat /bin/bash: no such file or directory": unknown` in Playground VM.

If FIND_FASTED_MIRROR is set to "false", the required packages are not installed in the Docker image. The case will happen in the daily CI build.

### DIFF
--- a/docker/modelzooserver/Dockerfile
+++ b/docker/modelzooserver/Dockerfile
@@ -11,15 +11,14 @@ RUN /bin/sh -c 'if [ "$FIND_FASTED_MIRROR" == "true" ]; then source find_fastest
     echo "Choose the fastest Alpine source ..." && \
     choose_fastest_alpine_source && \
     echo "Choose the fastest PIP source ..." && \
-    choose_fastest_pip_source && \
+    choose_fastest_pip_source; fi && \
     apk add --no-cache python3 py3-pip sudo bash docker-cli && \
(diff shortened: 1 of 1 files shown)


docker/modelzooserver/Dockerfile
```diff
@@ -11,14 +11,14 @@ RUN /bin/sh -c 'if [ "$FIND_FASTED_MIRROR" == "true" ]; then source find_fastest
     echo "Choose the fastest Alpine source ..." && \
     choose_fastest_alpine_source && \
     echo "Choose the fastest PIP source ..." && \
-    choose_fastest_pip_source && \
+    choose_fastest_pip_source; fi && \
```
lhw362950217, line 14: I believe this code fixes a bug, but not installing the bash. Maybe, we should change the PR title.

lhw362950217 (approved):
LGTM

docker/modelzooserver/Dockerfile
brightcoder01: We have `apk add --no-cache python3 py3-pip sudo bash` after `fi` here. This statement will install bash.

PR #2939, opened 2020-09-17
Add weighted_category_column support for key-value list column data

typhoonzero:


sneaxiy (dismissed):
Almost LGTM, and can merge if urgent. Should we rename `delimiter2` to another more reasonable name like `KeyValueDelimiter`, `IdWeightDelimiter`, etc?

python/runtime/tensorflow/input_fn.py
```diff
@@ -143,12 +172,30 @@ def read_feature_as_tensor(raw_val, feature_spec, feature_name):
...
+            kvsplited = tf.strings.split(
+                kvlist, feature_spec["delimiter2"],
+                result_type='RaggedTensor').to_tensor()
+            # slice key tensor and value tensor
+            indices = tf.reshape(tf.slice(kvsplited, [0, 0], [-1, 1]), [-1])
+            indices = tf.expand_dims(
+                tf.strings.to_number(indices, feature_spec["dtype"]), 1)
+            values = tf.cond(
+                tf.equal(tf.shape(kvsplited)[1],
+                         2), lambda: tf.strings.to_number(
```
sneaxiy, line 191: Can you explain when there would be `tf.shape(kvsplitted)[1] != 2` if `feature_spec["delimiter2"] != ""`?

sneaxiy (approved):
Excellent!

PR #2943, opened 2020-09-18
Collect step log

lhw362950217:
It seems if we do not add a `endTime` param, the logview will use current time as `endTime`, so the user will not see any log if he do not adjust the time param manually and refresh the page. So, we add a default time span.


weiguoz (approved):
LGTM

PR #2961, opened 2020-09-25
clean docker tags after push

typhoonzero:
Clean docker tags after push in the develop branch docker image uploading script, then after running CI, the images can be cleaned as "dangling" image.


Yancey0623 (approved):
LGTM

PR #2964, opened 2020-09-27
Infer weighted categorical columns

typhoonzero:


python/runtime/pai/xgboost_submitter/train.py
```diff
@@ -38,7 +38,7 @@ def train(datasource,
           batch_size=None,
           epoch=1,
           validation_steps=1,
-          verbose=0,
+          verbose=1,
```
sneaxiy: Why change this?

python/runtime/tensorflow/train.py
```diff
@@ -24,7 +24,7 @@
 from runtime.tensorflow.train_keras import keras_train_and_save
 
 # Disable TensorFlow INFO and WARNING logs
-os.environ["TF_CPP_MIN_LOG_LEVEL"] = "3"
+# os.environ["TF_CPP_MIN_LOG_LEVEL"] = "3"
```
sneaxiy: Why change this?

sneaxiy (approved):

PR #2966, opened 2020-09-27
Correct estimator_predict args

weiguoz:
This causes the unit test to fail on PAI.
Let's follow the [function signature](https://github.com/sneaxiy/sqlflow/blob/c6352174a34c7512e5ca3557e0a6677855f59a8a/python/runtime/tensorflow/predict.py#L141-L143).


sneaxiy (approved):
LGTM++

PR #2967, opened 2020-09-28
Replace the normal statement query with the refactorized codegen

sneaxiy:


go/executor/executor.go
```diff
@@ -225,6 +272,9 @@ func (s *pythonExecutor) runCommand(cmd *exec.Cmd, context map[string]string, lo
 }
 
 func (s *pythonExecutor) ExecuteQuery(stmt *ir.NormalStmt) error {
+	if ok, err := s.tryExperimentalExecute(stmt, false); ok {
```
typhoonzero, line 279: Can we have some FLAG to switch between modes?

sneaxiy: Added `SQLFLOW_USE_EXPERIMENTAL_CODEGEN` env.

typhoonzero (approved):

PR #2973, opened 2020-09-29
Speed up imports and fix the bug when selecting None value

sneaxiy:
- When using the refactorized codes to run a simple select statement like `SELECT * FROM iris.train LIMIT 5;`, it consumes about 2s, which is not user-friendly. It is because importing `MySQL/Hive/...Connection` is quite slow. This PR refines it, and the time to run `SELECT * FROM iris.train LIMIT 5;` decreases from 2s to 0.24s.
- Some tables contain None data. It was not supported in the previous refactorized version.


typhoonzero (approved):

lhw362950217 (approved):
LGTM++

PR #2974, opened 2020-09-29
Add self.commit after conn.execute and make dbapi raise Exception

sneaxiy:
- Fix the bug that the statement `INSERT INTO ...` does not take effect. We must commit the cursor after each SQL statement.
- Make the `conn.query` and `conn.execute` raise Exception if there is any error. This is designed to expose the errors asap instead of hiding the error in `ResultSet.success()` or `ResultSet.error()`.


weiguoz (dismissed):
LGTM.

weiguoz (approved):
LGTM++

PR #2998, opened 2020-10-09
Support special char in username/passwd for mysql

typhoonzero:
When the MySQL username or password have char like `@` `#` etc, make it work.


lhw362950217:
Should we exclude the ':' character? It's ambitious in this regexp.

lhw362950217 (approved):
LGTM

typhoonzero:
> Should we exclude the ':' character? It's ambitious in this regexp.

Maybe these chars should be escaped in this string. Will try in future PRs.

PR #3006, opened 2020-10-10
Unify dump model codes

sneaxiy:
Unify the dump model codes of the old codegen and new codegen.

### DIFF
--- a/go/model/model.go
+++ b/go/model/model.go
@@ -261,19 +261,45 @@ func loadModelFromDB(db *database.DB, table, cwd string) (*Model, error) {
 		}
 		defer os.RemoveAll(cwd)
 	}
-	tarFile, err := DumpDBModel(db, table, cwd)
+	tarFile, err := dumpDBModel(db, table, cwd)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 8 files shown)


go/model/model.go
```diff
@@ -261,19 +261,45 @@ func loadModelFromDB(db *database.DB, table, cwd string) (*Model, error) {
...
 	if !unzipModel {
-		return ExtractMetaFromTarball(tarFile, cwd)
+		return extractMetaFromTarball(tarFile, cwd)
 	}
 	return loadTar(path.Dir(tarFile), path.Base(tarFile), cwd)
 }
 
-// DumpDBModel dumps a model tarball from database to local
+// DumpDBModelAndExtractMeta dumps db model into a tar file and extracts the metadata
+func DumpDBModelAndExtractMeta(db *database.DB, table, cwd string, getMeta bool) (string, *Model, error) {
```
typhoonzero, line 275: Can we have two functions like :

```go
// load model and meta.
model := NewModelFromDB(db *database.DB, table)
// dump model and meta to a tarball.
model.DumpToTar()
```

This is more straightforward forward maybe.

PR #3015, opened 2020-10-12
Refactor TO RUN to compiler arch: Go side.

brightcoder01:
The Go side mainly contains the code_gen component.
Add a placeholder implementation in the runtime layer for local run.


python/runtime/local/submitter.py
```diff
@@ -160,6 +160,16 @@ def submit_local_explain(datasource,
                  model=model)
 
 
+def submit_local_run(
+    datasource,
+    select,
+    image_name,
+    params,
+    into):
+    print("Execute local run. datasource:{}, select: {}, image_name: {}, params: {}, into: {}".format(
```
weiguoz: Add a space behind `datasource:` to unify the format.

go/codegen/experimental/codegen_step_run.go
```diff
@@ -0,0 +1,57 @@
...
+	err = runTemplate.Execute(&program, filter)
+	if err != nil {
+		return "", err
+	}
+
+	return program.String(), nil
+}
+
+const runStepTemplate = `
+def step_entry_{{.StepIndex}}():
```
sneaxiy, line 62: What if `TO RUN` is not running a Python program? For example, a C/C++ executable? Of course, we can call the C/C++ executable in `step_entry_0` on the Python side. But it requires that the docker image must install Python.

brightcoder01: Good question. Currently we have the requirement that the runnable docker image is built based on sqlflow/sqlflow:step. It has runtime library inside this image which already contains python.

python/runtime/local/submitter.py
brightcoder01: Good catch, updated.

sneaxiy (approved):

PR #3029, opened 2020-10-15
Fix go maxcompute select error

sneaxiy:
Fix the error:
```
--- FAIL: TestEnd2EndMaxComputePAI/group/CasePAIMaxComputeDNNTrainPredictExplain (250.02s)
            testing.go:165: 
                  Error Trace:  testing.go:165
                                      e2e_pai_maxcompute_test.go:323
                  Error:        unsupported type comparison type.googleapis.com/google.protobuf.StringValue int64
                  Test:         TestEnd2EndMaxComputePAI/group/CasePAIMaxComputeDNNTrainPredictExplain
```

The gomaxcompute would return `string` as the ScanType for `BIGINT/DOUBLE/...` column. This would cause errors when we want to get the integral value of a `BIGINT` column.

### DIFF
--- a/go/executor/fields.go
+++ b/go/executor/fields.go
@@ -16,16 +16,11 @@ package executor
 import (
 	"database/sql"
 	"fmt"
-	"reflect"
 	"strings"
 	"time"
 
 	"github.com/go-sql-driver/mysql"
 )
-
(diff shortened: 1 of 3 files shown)


weiguoz (approved):
LGTM++

PR #3034, opened 2020-10-16
Make pai_local executor can run using experimental codegen

sneaxiy:


python/runtime/model/model.py
```diff
@@ -110,20 +110,23 @@ def _zip(self, local_dir, tarball, save_to_db=False):
...
-            model_obj_file = os.path.join(local_dir, MODEL_OBJ_FILE_NAME)
-            with open(model_obj_file, "w") as f:
-                d = self._to_dict()
-                f.write(json.dumps(d, cls=JSONEncoderWithFeatureColumn))
-        else:
-            model_obj_file = None
+        # NOTE: the unzip files of the job tarball should be skipped
+        from runtime.pai.prepare_archive import ALL_TAR_FILES
+
+        def filter(tarinfo):
```
typhoonzero, line 116: Why need a filter

typhoonzero (approved):

python/runtime/model/model.py
sneaxiy: Because when we submit a job to PAI, there may be Python packages(`sqlflow_models`, `xgboost`, etc.), parameter file (`train.pkl`), etc. These files should be skipped when saving the model.

typhoonzero: We can save all of the files actually, the file list to save is not static, e.g. a custom model may save some files for prediction.

PR #3047, opened 2020-10-20
Add buffer size for SQLFS writer in Python side

sneaxiy:
MySQL `TEXT` can contain 65536 characters at most (r.f. https://dev.mysql.com/doc/refman/8.0/en/storage-requirements.html#data-types-storage-reqs-strings). Add `buf_size` for `SQLFSWriter` to control the maximum character number on one line.

### DIFF
--- a/python/runtime/model/db.py
+++ b/python/runtime/model/db.py
@@ -73,10 +73,10 @@ class SQLFSWriter(object):
         self.writer.write([self.row_idx, block])
         self.row_idx += 1
 
-    def close(self, *args, **kwargs):
+    def close(self):
         self.flush()
         self.writer.close()
-        self.context_manager.__exit__(*args, **kwargs)
+        self.context_manager.__exit__(None, None, None)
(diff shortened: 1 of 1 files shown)


python/runtime/model/db.py
```diff
@@ -44,27 +44,53 @@ def _drop_table_if_exists(conn, table):
     conn.execute(sql)
 
 
+# NOTE: MySQL TEXT type can contain 65536 characters at most.
+# We need to limit the max string length of each row.
+MAX_LENGTH_TO_WRITE_PER_ROW = 4096
```
typhoonzero, line 49: How about increase this to 32k to avoid so that large models will not have too many rows.

typhoonzero (approved):

python/runtime/model/db.py
sneaxiy: Agree. I would do it in the later PR.

PR #3048, opened 2020-10-20
support pai predict using experimental codegen

sneaxiy:


python/runtime/pai/submitter_train.py
```diff
@@ -45,18 +47,25 @@ def get_pai_train_cmd(datasource, estimator_string, model_name, train_table,
...
         The command to submit a PAI train task
     """
+    if label_column is not None:
+        label_name = label_column.get_field_desc()[0].name
+    else:
+        label_name = None
+
+    with db.connect_with_datasource(datasource) as conn:
+        schema = db.get_table_schema(conn, train_table)
+        feature_column_names = [s[0] for s in schema if s[0] != label_name]
```
typhoonzero, line 57: `feature_column_names` should be ordered, the order is determined by SELECT statement, not the table schema, we generate features using this list. Should test `SELECT b,a ... TO TRAIN` using table got column order: `a,b`, don't know whether it will work.

sneaxiy: The `train_table` is created by the `SELECT` statement using `CREATE TABLE train_table AS SELECT ...`.  So, the schema order of the `train_table` is the same as the `SELECT` statement.

typhoonzero: I see.

typhoonzero (approved):

PR #3074, opened 2020-10-28
Add sqlfs read one by one mode to speed up metadata reading

sneaxiy:
Model metadata usually contains little bytes. Add read one by one mode to speed up model metadata reading instead of reading all rows in the model table.


lhw362950217 (approved):
LGTM

PR #3079, opened 2020-10-28
Implement local submitter for TO RUN in SQLFlow runtime.

brightcoder01:


python/runtime/local/submitter.py
```diff
@@ -165,13 +169,50 @@ def submit_local_explain(datasource,
...
+        args.extend(params[1:])
+        sub_process = subprocess.run(args=args,
+                                     env=subprocess_env,
+                                     stdout=subprocess.PIPE,
+                                     stderr=subprocess.PIPE)
+    else:
+        print(
+            "The other executable except Python program is not supported yet")
+
+    if sub_process:
```
sneaxiy, line 214: How about check `sub_process.exitcode == 0` to find whether the subprocess runs well? If `sub_process.exitcode != 0`, try to print `sub_process.stderr` to the stderr of the parent process.

sneaxiy:
Try to enable the `e2e_workflow_test.go` for `TO RUN` statements as well.

https://github.com/sql-machine-learning/sqlflow/blob/9bcbe31872569c87fcb666f9b628988084287e04/go/cmd/sqlflowserver/e2e_workflow_test.go#L84-L102

python/runtime/local/submitter.py
brightcoder01: Good idea, updated.

python/runtime/local/submitter.py
```diff
@@ -165,13 +170,51 @@ def submit_local_explain(datasource,
...
+                                     stderr=subprocess.PIPE)
+    elif file_ext.lower() == ".py":
+        args = ["python", "-m", program_file_path.stem]
+        args.extend(params[1:])
+        sub_process = subprocess.run(args=args,
+                                     env=subprocess_env,
+                                     stdout=subprocess.PIPE,
+                                     stderr=subprocess.PIPE)
+    else:
+        print(
```
sneaxiy, line 211: Should we raise an error here instead of printing a message?

brightcoder01: Good idea, updated.

sneaxiy (approved):

PR #3081, opened 2020-10-29
Remove build directory anyway

sneaxiy:


scripts/travis/deploy_docker.sh
```diff
@@ -91,3 +91,6 @@ push_image jupyter jupyter
 push_image server server
 push_image step step
 push_image modelzooserver modelzooserver
+
+echo "Clean up root permission $TRAVIS_BUILD_DIR/build ..."
+docker run --rm  -v "$TRAVIS_BUILD_DIR":/work sqlflow:ci rm -rf /work/build /work/java /work/python
```
typhoonzero, line 96: There's `set -e` at the beginning of this script, if the above script went wrong, this line will not be executed. We can merge this for a quick fix and fix it later.

typhoonzero (approved):

PR #3082, opened 2020-10-29
Fix PAI label error

sneaxiy:
Some models like `OneClassSVM` may not have `LABEL`. Fix the bug when there is no `LABEL` using PAI.


lhw362950217 (approved):
LGTM

PR #3091, opened 2020-11-04
[Refactory] Remove oss model save/load on PAI 

sneaxiy:


go/codegen/experimental/codegen_normal_stmt.go
```diff
@@ -36,6 +36,7 @@ def step_entry_{{.StepIndex}}():
             print(l)
     else:
         conn.execute(stmt)  # Exception would raise if error
+        print("OK")
```
typhoonzero, line 39: remove this line?

typhoonzero (approved):
LGTM

go/codegen/experimental/codegen_normal_stmt.go
sneaxiy: OK. Would do it in the next PR.

PR #3100, opened 2020-11-10
Refine SQLFS reader

sneaxiy:
- Remove ut `TestSQLFSHiveWriterWriteAndRead` and `TestSQLFSSQLWriterWriteAndRead`, since they are the same as `TestSQLFSWriteAndRead`.
- Make `sqlfs.reader` return `0, io.EOF` at the end of the table, which behaves the same as `os.Open(...)` in Go.


weiguoz (approved):
LGTM

PR #3104, opened 2020-11-20
Update actions env setting

typhoonzero:
Update actions env setting according to https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/


weiguoz (approved):
LGTM

